PR #67, opened 2021-03-26
init: default path if no init called

### LINKED ISSUE #51: Introduce default directory
```
dvclive.error.InitializationError: Initialization error - call `dvclive.init()` before `dvclive.log()`
```

Why don't we use a default dir if it is not set up explicitly?

pared:
Fixes #51

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -143,10 +143,10 @@ def test_custom_steps(tmp_dir):
 
 @pytest.mark.parametrize("html", [True, False])
 @pytest.mark.parametrize("summary", [True, False])
-def test_init_from_env(tmp_dir, summary, html):
-    os.environ[env.DVCLIVE_PATH] = "logs"
-    os.environ[env.DVCLIVE_SUMMARY] = str(int(summary))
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -143,13 +143,19 @@ def test_custom_steps(tmp_dir):
 
 @pytest.mark.parametrize("html", [True, False])
 @pytest.mark.parametrize("summary", [True, False])
-def test_init_from_env(tmp_dir, summary, html):
-    os.environ[env.DVCLIVE_PATH] = "logs"
-    os.environ[env.DVCLIVE_SUMMARY] = str(int(summary))
-    os.environ[env.DVCLIVE_HTML] = str(int(html))
```
pared, line 149: It modified env for all following tests

jorgeorpinel:
Hi! Should we document this in https://dvc.org/doc/dvclive/usage @pared ? It currently implies the call is required and says the `path` argument is required.

pared:
@jorgeorpinel created issue for that.

PR #69, opened 2021-03-31
Save checkpoint for Keras

dmpetrov:


pared:
@dmpetrov Shouldn't user make use of `tf.keras.callbacks.ModelCheckpoint`?

dmpetrov:
@pared good point and it is still possible with the current implementation. But it is better to have all DVCLive functionality in a single call.

Also, you can mix up the order if you separate them that will create checkpoints with previous models - another place when user can make a mistake.

dberenbaum:
What about xgboost? If we start adding functionality to one callback, it should probably apply across all of them.

In that case, it might make more sense to add a `save_model` method to the `MetricLogger` class to wrap this call and ensure a consistent interface across integrations. This would also create more consistency between the built-in integrations like Keras and cases like PyTorch where no integration is available, since a `save_model` method would enable users to save models through `dvclive` calls in either scenario. Thoughts?

pared:
@dberenbaum @dmpetrov Sorry for getting to this so late, I forgot about it during my brief absence. 
@dberenbaum How would user define model path? Would it be via `init` call? This approach makes sense to me because that would also force us to keep consistent behaviour between integrations, no matter how they are written.

dberenbaum:
> How would user define model path? Would it be via `init` call?

I suppose it would need to be via `dvclive.init()`, right?

pared:
@dberenbaum yeah, I think so. It just makes me think whether we shouldn't change the `init` args a little so that first time user does not see 10 different args upon inspecting, but that's minor thing to resolve during review, I think I know how to handle that.

daavoo:
I agree with https://github.com/iterative/dvclive/issues/105#issuecomment-876739618 about not needing to add additional arguments to `dvclive.init`. It feels natural and very small overhead the fact of just passing an additional argument to the callback instantiation:

```python
model.fit(X_train, y_train,  validation_data=(X_test, y_test),
    callbacks=[DvcLiveCallback(model_file="output_model.h5")])
```

One thing we might need to consider is to enable the option of saving `weights_only`, as this seems to be a common feature across different keras integration and also for consistency with `ModelCheckpoint`

daavoo:
I have a some concerns about this. 

When someone uses `keras`, in conjunction with other *ML Loggers* and the `ModelCheckpoint` callback, the behavior will be different to the one we provide would with `DVCLiveCallback`. 

The following features will be missing:

- `ModelCheckpoint` has the option to save individual model files for **each** epoch by [formatting the given `filepath`](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/ModelCheckpoint)
- `ModelCheckpoint` has the option to `monitor` a metric in order to only save the **best** model file (`save_best_only`).

dberenbaum:
Good point, @daavoo! You raise good points, but I want to push back a little to make sure we think through both sides (I don't have a strong preference now as to whether we implement this).

>     * `ModelCheckpoint` has the option to save individual model files for **each** epoch by [formatting the given `filepath`](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/ModelCheckpoint)

If we are assuming a dvc-like workflow, this PR will save models each epoch to the same file path, and each one can be checked out as needed. Even though dvclive is independent of dvc, do we want to start supporting workflows where the filepath designates the epoch number? This could impact how we save a lot more than the model files.

On the other hand, keras also supports saving less frequently than every epoch, similar to #103, and it supports saving only the model weights, both of which could be helpful.

> What would be the benefits for the user if they replace ModelCheckpoint with DVCLiveCallback's model_file? 

@dmpetrov answered a similar question above. What do you think @daavoo?

> @pared good point and it is still possible with the current implementation. But it is better to have all DVCLive functionality in a single call.
>
> Also, you can mix up the order if you separate them that will create checkpoints with previous models - another place when user can make a mistake.

daavoo:
> If we are assuming a dvc-like workflow, this PR will save models each epoch to the same file path, and each one can be checked out as needed.

I understand.

> Even though dvclive is independent of dvc, do we want to start supporting workflows where the filepath designates the epoch number? This could impact how we save a lot more than the model files.

We should probably stick with `dvc-like` workflows.

> On the other hand, keras also supports saving less frequently than every epoch, similar to #103, and it supports saving only the model weights, both of which could be helpful.

Are you saying that we should add support to those features in our callback??


> @dmpetrov answered a similar question above. What do you think @daavoo?
> 
> > @pared good point and it is still possible with the current implementation. But it is better to have all DVCLive functionality in a single call.
> > Also, you can mix up the order if you separate them that will create checkpoints with previous models - another place when user can make a mistake.

I understand both points. Probably it would be worth mentioning in the (not currently existing) integration docs

pared (approved):

pared:
> Are you saying that we should add support to those features in our callback??

This makes providing our own saving capabilities problematic: where do we stop with repeating existing functionality. 
Though in our particular case I would just stop with providing the model name in the constructor - so that it goes nicely with `checkpoints`. If the user wants a more advanced use case, well, that will require additional legwork.

PR #72, opened 2021-04-15
codecov check

pared:


.github/workflows/tests.yaml
```diff
@@ -37,6 +37,8 @@ jobs:
       run: |
         pip install wheel
         pip install -e ".[tests]"
+    - name: check env
+      run: env
```
shcheklein, line 41: since it's a public project does it mean that we exposed all variables by doing this check (in CI logs)?

pared: Yes, probably not wisest on my part, all logs have been deleted

PR #74, opened 2021-05-13
init: throw on dvc/dvclive config mismatch

### LINKED ISSUE #60: Should the `dvc stage add --live <path>` arg override `init(path)`?
It's more of an idea for `dvc stage add` but I thought I'd ask here first (we can transfer the issue if needed).

If you call `dvc.init()` manually, `path` is always required. So in those cases, combined with using DVC, you have to match `path` with the arg. of `dvc stage add --live`. Seems error-prone.

The arg to `--live` could be optional. If there's no `init()` call in code, ~~then DVC would need to give an error~~

UPDATE: Calling `init()` is no longer required though, as DVCLive has a default `path` for its outputs now.

Thoughts?

pared:
Fixes #60 

TODO
~~- [x] - allow overriding if proper env variable is present~~
~~- [x] - If same dir is provided via env, make sure to update config~~

EDIT:
we agreed in #74 to throw on configuration mismatch

### DIFF
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -61,7 +61,7 @@ class MetricLogger:
             os.remove(self.html_path)
 
     @staticmethod
-    def from_env(existing_logger: "MetricLogger" = None):
+    def from_env():
         from . import env
 
         if env.DVCLIVE_PATH in os.environ:
@@ -74,31 +74,8 @@ class MetricLogger:
                 ),
                 "resume": bool(int(os.environ.get(env.DVCLIVE_RESUME, "0"))),
             }
-
-            if existing_logger:
-
-                def config_matches(metric_logger: MetricLogger):
-                    return (
-                        env_config["summary"] == metric_logger._summary
-                        and env_config["html"] == metric_logger._html
-                        and env_config["checkpoint"]
-                        == metric_logger._checkpoint
-                    )
-
-                if existing_logger.dir == directory and config_matches(
-                    existing_logger
-                ):
-                    return existing_logger
-                else:
-                    logger.info(
-                        "Dvclive logger config ('%s') has changed. "
-                        "New logger will write to '%s'.",
-                        existing_logger.dir,
-                        directory,
-                    )
-
             return MetricLogger(directory, **env_config)
-        return existing_logger
+        return None
 
     def matches_env_setup(self):
         from . import env
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -61,23 +61,53 @@ class MetricLogger:
             os.remove(self.html_path)
 
     @staticmethod
-    def from_env():
+    def from_env(existing_logger: "MetricLogger" = None):
         from . import env
 
         if env.DVCLIVE_PATH in os.environ:
             directory = os.environ[env.DVCLIVE_PATH]
-            dump_latest = bool(int(os.environ.get(env.DVCLIVE_SUMMARY, "0")))
-            html = bool(int(os.environ.get(env.DVCLIVE_HTML, "0")))
-            checkpoint = bool(int(os.environ.get(env.DVC_CHECKPOINT, "0")))
-            resume = bool(int(os.environ.get(env.DVCLIVE_RESUME, "0")))
-            return MetricLogger(
-                directory,
-                summary=dump_latest,
-                html=html,
-                checkpoint=checkpoint,
-                resume=resume,
-            )
-        return None
+            env_config = {
+                "summary": bool(int(os.environ.get(env.DVCLIVE_SUMMARY, "0"))),
+                "html": bool(int(os.environ.get(env.DVCLIVE_HTML, "0"))),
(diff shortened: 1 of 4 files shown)


dvclive/metrics.py
```diff
@@ -49,23 +49,44 @@ def __init__(
                 ) from exception
 
     @staticmethod
-    def from_env():
+    def from_env(existing_logger: "MetricLogger" = None):
         from . import env
 
         if env.DVCLIVE_PATH in os.environ:
             directory = os.environ[env.DVCLIVE_PATH]
```
jorgeorpinel, line 56: So we're talking about actual environment variables? Specifically `DVCLIVE_PATH`. This comes a bit out of the left field since there's no info on this in https://dvc.org/doc/dvclive/dvclive-with-dvc

dvclive/metrics.py
```diff
@@ -49,23 +49,44 @@ def __init__(
...
+                    return existing_logger
+                else:
+                    logger.info(
+                        "Dvclive logger config ('%s') has changed. "
+                        "New logger will write to '%s'.",
+                        existing_logger.dir,
+                        directory,
+                    )
+
+            return MetricLogger(directory, **env_config)
```
jorgeorpinel, line 88: OK so if the env vars don't match the existing logger, a new logger is created... Seems correct i.e. this does what the PR states. But I'm not sure I get how this addresses #60. Need to think... ⌛

jorgeorpinel (commented):
Do you need my approval code-wise though? May be best to ask another core member anyway (additionally) once the behavior is clear at least. I'll continue discussing in #60 for now.

dvclive/metrics.py
```diff
@@ -49,23 +49,44 @@ def __init__(
...
+                    )
+
+                if existing_logger.dir == directory and config_matches(
+                    existing_logger
+                ):
+                    return existing_logger
+                else:
+                    logger.info(
+                        "Dvclive logger config ('%s') has changed. "
+                        "New logger will write to '%s'.",
```
jorgeorpinel, line 83: p.s. probably no need for the old path + may make the message too long/unreadable?
```suggestion
                        "Dvclive logger config has changed. "
                        "The new logger will write to '%s'.",
```

jorgeorpinel: Also should it be a `logger.warn` ?

dvclive/metrics.py
jorgeorpinel: p.p.s. https://github.com/iterative/dvclive/issues/79 😬

dvclive/metrics.py
jorgeorpinel: My q was answered in https://github.com/iterative/dvclive/issues/60#issuecomment-843941753:

> when dvc provides --live option for stage/run, it is equivalent to setting the environmental variables that dvclive can understand ... existence of those env vars means that we are running in the context of dvc

It's how DVC and Dvclive communicate. Got it 👍

dvclive/metrics.py
jorgeorpinel: Answered in https://github.com/iterative/dvclive/issues/79#issuecomment-844818063, thanks.

jorgeorpinel (approved):
Happy to approve this. TBH I didn't test it, but the purported change should indeed fix #60.

daavoo:
@pared Is this still [WIP] ?

pared:
@daavoo sorry for the late response, yes, I need to finish this change according to the discussion under #60 I will get to it after later this week.

dvclive/metrics.py
```diff
@@ -61,23 +61,53 @@ def _cleanup(self):
             os.remove(self.html_path)
 
     @staticmethod
-    def from_env():
+    def from_env(existing_logger: "MetricLogger" = None):
```
daavoo, line 64: @pared I'm not sure what the changes to `from_env` are intended to do. `from_env` is never being called with the new arg `existing_logger`. Could these changes be from previous iterations and are now outdated?

tests/test_main.py
```diff
@@ -209,6 +216,28 @@ def test_no_init(tmp_dir):
...
 
+def test_fail_on_conflict(tmp_dir, monkeypatch):
+    dvclive.init("some_dir")
+    monkeypatch.setenv(env.DVCLIVE_PATH, "logs")
+
+    with pytest.raises(ConfigMismatchError):
+        dvclive.log("m", 0.1)
+
+
+def test_dont_override_from_env(tmp_dir, monkeypatch):
```
daavoo, line 227: Not sure what are we testing here, might be related with my comment above about `from_env(existing_logger)` being outdated changes?

dvclive/metrics.py
pared: Yep, wrongly conducted rebase, sorry for that

tests/test_main.py
pared: Wrongly conducted rebase, removing.

daavoo (approved):

Issue #75, opened 2021-05-15
Feature Request: Support different formats of summaries


Benjamin-Etheredge:
Summaries are currently printed out as a single JSON line. It would be handy to be able to change this format to a more readable format (such as YAML or multiline JSON).
These summaries are more likely to be checked into git. As such, having them in a more human-readable format would be beneficial.

pared:
While the issue itself seems easy to implement, it unravels new problem that we had to address at some point: dvclive configuration.

dberenbaum:
Can we start with the easy implementation 😆 ?

pared:
@dberenbaum yes we can, we can probably pass it silently by replacing `summary` flag with string arg `summary_path` that will also tell dvc what format we want the summary in. Though in that case there would be some reworking needed also on DVC side.

dberenbaum:
I was thinking even smaller, like can we just dump the json with an indent value so that it gets pretty printed?

pared:
Ahh, ok, I guess my imagination went too far. Give me a sec...

pared:
@Benjamin-Etheredge would properly formatted json be good enough for start? Created PR for that.

pared:
Fixed in #78. We can revisit it if we need yaml support, though purely aesthetics reasoning is not enough to justify amount of changes needed to support different summary formats. (It would require configuration, which is currently not necessary for `dvclive`)

PR #78, opened 2021-05-18
json: use indent

pared:
Related to #75

### DIFF
--- a/dvclive/keras.py
+++ b/dvclive/keras.py
@@ -1,4 +1,6 @@
-from tensorflow.keras.callbacks import Callback
+from tensorflow.keras.callbacks import (  # pylint: disable=no-name-in-module
+    Callback,
+)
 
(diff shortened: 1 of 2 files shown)


dvclive/keras.py
```diff
@@ -1,4 +1,6 @@
-from tensorflow.keras.callbacks import Callback
+from tensorflow.keras.callbacks import (  # pylint: disable=no-name-in-module
```
dberenbaum, line 1: This seems like a weird change, but I'm guessing it's auto formatting? 🤷

dberenbaum (approved):
LGTM

dvclive/keras.py
pared: Seems like it started to occur recently, probably its some new rule for `pylint`. I need to research it, but don't want to block smaller changes.

PR #80, opened 2021-05-20
github actions: run tests for differnet os-es and python versions

pared:
Run tests for different python versions and different machines

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -25,8 +25,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -37,6 +37,10 @@ jobs:
       run: |
         pip install wheel
         pip install -e ".[tests]"
+    - name: install libomp
+      # required for xgboost
+      if: runner.os == 'macOS'
+      run: brew install libomp
```
pared, line 43: https://github.com/dmlc/xgboost/issues/4477

PR #96, opened 2021-06-17
Clean up summary and html files

### LINKED ISSUE #52: Clean up summary and html files
It seems like all the metrics files are cleaning up when `init()` is called. But summary and html files stay in the same place.
Should we clean them up as well?

daavoo:
Closes #52

### DIFF
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -42,11 +42,10 @@ class MetricLogger:
         else:
             shutil.rmtree(self.dir, ignore_errors=True)
 
-            try:
+            if os.path.exists(self.summary_path):
                 os.remove(self.summary_path)
+            if os.path.exists(self.html_path):
                 os.remove(self.html_path)
-            except OSError:
-                pass
 
             try:
                 os.makedirs(self.dir, exist_ok=True)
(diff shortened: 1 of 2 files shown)


dvclive/metrics.py
```diff
@@ -41,6 +41,13 @@ def __init__(
                 self._step = step
         else:
             shutil.rmtree(self.dir, ignore_errors=True)
+
+            try:
+                os.remove(self.summary_path)
+                os.remove(self.html_path)
+            except OSError:
```
pared, line 48: Maybe lets check if summary and html_path exists before deleting?
OSError might happen due to various reasons.

pared (changes requested):

dvclive/metrics.py
daavoo: What if we replace `OSError` with a more specific `FileNotFoundError`??

I used the `try`/`except` insted of `if` check because that's how it is internally handled in the standard library `pathlib`:

https://github.com/python/cpython/blob/main/Lib/pathlib.py#L1188

pared: Got it, no strong preference from me here, yet still: what if `summary_path` does not exist and `html_path` does? In this case we won't delete `html_path`.

daavoo: That's true. I will split the removal

pared (approved):

PR #97, opened 2021-06-18
Allow nested metrics

### LINKED ISSUE #95: Feature request: allow nested metric values via slash in metric name
There are situations when we may want to group metric values, for example, in "train", "val", "test":
1. We may need this to select what columns we want to show in Studio (e.g., turn on / turn of metrics for "val" completely).
2. Later we also can group plots in Studio/Vscode extension, like it is done in W&B, for example.

If “name” in dvclive.log(name, value) happens to be like “one/two” should we create a subfolder “one” with a filename “two”? Right now it raises an error:

```
FileNotFoundError: [Errno 2] No such file or directory: 'training_metrics/train/box_loss.tsv'
```

At the same time it should be handled in .json file with the final epoch - there we could have nested dictionaries.

It seems like wandb has a special way to handle this - maybe grouping metrics which starts with the same substring? Not sure how it works, actually.

### LINKED ISSUE #101: FileNotFoundError: [Errno 2] No such file or directory: '/home/xxx/codebases/dvclive/train/loss.tsv'
If the name contains `/`  like 'train/loss', `self.log` will raise FileNotFoundError

https://github.com/iterative/dvclive/blob/8b260ae19a834b693ca8853b6d12140ec695effc/dvclive/metrics.py#L108-L130

https://github.com/iterative/dvclive/blob/8b260ae19a834b693ca8853b6d12140ec695effc/dvclive/serialize.py#L7-L15

daavoo:
Closes #95 
Closes #101

### DIFF
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -4,7 +4,6 @@ import os
 import shutil
 import time
 from collections import OrderedDict
-from pathlib import Path
 from typing import Dict
 
 from .dvc import get_signal_file_path, make_checkpoint
@@ -123,12 +122,12 @@ class MetricLogger:
         if step:
             self._step = step
 
-        metric_history_path = (Path(self.history_path) / name).with_suffix(
-            ".tsv"
-        )
-        metric_history_path.parent.mkdir(parents=True, exist_ok=True)
+        metric_history_path = os.path.join(self.history_path, name + ".tsv")
+        os.makedirs(os.path.dirname(metric_history_path), exist_ok=True)
 
-        self._nested_set(self._metrics, Path(name).parts, val)
+        self._nested_set(
+            self._metrics, os.path.normpath(name).split(os.path.sep), val,
+        )
 
         ts = int(time.time() * 1000)
         d = OrderedDict([("timestamp", ts), ("step", self._step), (name, val)])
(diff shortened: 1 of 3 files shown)


dvclive/metrics.py
```diff
@@ -138,3 +142,9 @@ def read_step(self):
     def read_latest(self):
         with open(self.summary_path, "r") as fobj:
             return json.load(fobj)
+
+    @staticmethod
+    def _nested_set(d, keys, value):
+        for key in keys[:-1]:
+            d = d.setdefault(key, {})
+        d[keys[-1]] = value
```
pared, line 150: Should it be a part of logger? It seems to me like a dictionary manipulation method, we should probably get it out of the class as a separate function or even thing about utils package for the project. 

I like the `_nested_set` solution, thought if we will be tinkering more with `self.metrics` in the future, its worth remembering that we could add `dpath` as dependency which supports different nested operations with dict out of the box. But its probably not worh considering now, when we only need one operation.

pared (commented):
LGTM, one minor thing regarding method placement.

dvclive/metrics.py
daavoo: I opted for creating an `utils` module. `dpath` seems like a good choice if more operations are needed in the future

pared (approved):

PR #98, opened 2021-06-18
Add gitignore

daavoo:


shcheklein:
Curious - what are the benefits of using the default that includes a lot? I would probably go with a smaller file that we would add items as we need them. (e.g. Celery - I doubt that this project will have something like that)

.gitignore
```diff
@@ -113,26 +38,9 @@ ENV/
...
-.pyre/
-
-# pytype static type analyzer
-.pytype/
 
-# Cython debug symbols
-cython_debug/
+# Editor
+.vscode/
+.idea/
```
shcheklein, line 46: minor: include newline

pared (approved):
I have mixed feelings about it. Should we be shotgunning everything by default? Shouldn't we expect user to have global gitignore on her/his machine? 
Especially when we are talking about ide's - this is user specific configuration. It's hard to define what should be and what should not be here.

.gitignore
pared: @daavoo  should I merge, or do you intend to include this suggestion?

daavoo:
> I have mixed feelings about it. Should we be shotgunning everything by default? Shouldn't we expect user to have global gitignore on her/his machine?
> Especially when we are talking about ide's - this is user specific configuration. It's hard to define what should be and what should not be here.

I see your point. What would be the patterns that you would consider worth including?? Maybe only those strictly related with the repository and the tools explicitly used (i.e. Byte-compiled, python packaging and `pytest`) ?

pared:
@daavoo this is good point - I don't know how far we should go with ignoring :) 
hence the approval and concerns. Let's keep it as is for now. It doesn't seem it will harm us.

PR #106, opened 2021-07-13
Update log type hints

daavoo:
To match condition:

https://github.com/iterative/dvclive/blob/master/dvclive/metrics.py#L126

### DIFF
--- a/dvclive/__init__.py
+++ b/dvclive/__init__.py
@@ -1,3 +1,5 @@
+from typing import Union
+
 from dvclive.version import __version__  # noqa: F401
 
 from .error import InitializationError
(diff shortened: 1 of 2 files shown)


pared (approved):

dvclive/__init__.py
```diff
@@ -24,7 +26,7 @@ def init(
     return _metric_logger
 
 
-def log(name: str, val: float, step: int = None):
+def log(name: str, val: Union[int, float], step: int = None) -> None:
```
skshetry, line 29: Strictly speaking, you don't need `Union` for `int` with `float`. See [PEP 484](https://www.python.org/dev/peps/pep-0484/#the-numeric-tower). 

> this PEP proposes a straightforward shortcut that is almost as effective: when an argument is annotated as having type float, an argument of type int is acceptable;

PR #122, opened 2021-07-21
Safer cleanup

### LINKED ISSUE #102: my codebase is deleted when I set the path as my codebase directory
When I set the path as my codebase directory, a terrible thing happened. My codebase was deleted. 

https://github.com/iterative/dvclive/blob/8b260ae19a834b693ca8853b6d12140ec695effc/dvclive/metrics.py#L43

daavoo:
Changes:

- Added `.dvclive.tsv` prefix to *metrics logs* files. Does not affect visualizations with `dvc plots` or the HTML report.
- Replaced `shutil.rmdir` with `rglob`+`unlink()` to prevent unintentional deletions of user files.

Closes #102 

- [X] P.R. to `dvc.org` https://github.com/iterative/dvc.org/pull/2653

### DIFF
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -52,7 +52,7 @@ class MetricLogger:
     def _cleanup(self):
 
         for dvclive_file in Path(self.dir).rglob("*.dvclive.tsv"):
-            dvclive_file.unlink(missing_ok=True)
+            dvclive_file.unlink()
 
         if os.path.exists(self.summary_path):
             os.remove(self.summary_path)
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -138,9 +138,7 @@ class MetricLogger:
         if step:
             self._step = step
 
-        metric_history_path = os.path.join(
-            self.history_path, name + ".tsv"
-        )
+        metric_history_path = os.path.join(self.history_path, name + ".tsv")
         os.makedirs(os.path.dirname(metric_history_path), exist_ok=True)
 
         nested_set(
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -139,7 +139,7 @@ class MetricLogger:
             self._step = step
 
         metric_history_path = os.path.join(
-            self.history_path, name + ".dvclive.tsv"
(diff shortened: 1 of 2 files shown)


pared (commented):
LGTM, from technical point of view. Question is whether we want to add those long suffixes to each file, maybe `.lv.tsv`/`.dl.tsv` might do the trick?

daavoo:
> LGTM, from technical point of view. Question is whether we want to add those long suffixes to each file, maybe `.lv.tsv`/`.dl.tsv` might do the trick?

Sure. Probably even just `.tsv` and assume that it will be too much coincidence to have an user setting `path` to something unexpected and also having some relevant `.tsv` files inside.

pared:
>Sure. Probably even just .tsv and assume that it will be too much coincidence to have an user setting path to something unexpected and also having some relevant .tsv files inside.

Huh, that might be even better - we have to draw the line somewhere. We don't even expect user to put anything other than metrics in  `init` dir. So cleaning dir of `tsv` sounds good to me (if we really want to get have this safety mechanism)

daavoo:
> Huh, that might be even better - we have to draw the line somewhere. We don't even expect user to put anything other than metrics in `init` dir. So cleaning dir of `tsv` sounds good to me (if we really want to get have this safety mechanism)

I would say that it is worth it to include this safety mechanism. Deleting everything with force `rmdir` felt like having too much assumptions and problematic depending on how much the user knows about `dvclive`. It's not hard for me to imagine users setting `path` to places where they won't expect a `rm -rf`.

pared (approved):
What about [this comment](https://github.com/iterative/dvclive/issues/102#issuecomment-886246998). Do we want to save old files somewhere? Probably we can do that later, the current state of change is a proper improvement regarding files deletion behaviour.

daavoo:
> What about [this comment](https://github.com/iterative/dvclive/issues/102#issuecomment-886246998)? Do we want to save old files somewhere? Probably we can do that later, the current state of change is a proper improvement regarding files deletion behaviour.

Saving to temp sounds like a good enhancement. Probably for a subsequent change, like you said

PR #124, opened 2021-07-26
Extend summary

### LINKED ISSUE #89: summary: Add option to store **best** value?
By default the values for each metric at the latest step are being saved in the `summary`. It might be interesting to add an option to additionally save the **best** value for each metric.

An example use case would be a usual deep learning training loop where the model is being validated at the end of each epoch and the last epoch is not necessarily the one with the best performance. Having the best value saved in the `summary` could be more useful for comparing experiments (i.e. `dvc metrics diff --targets dvclive.json`)

A potential problem would be how to expose to the user (https://github.com/iterative/dvclive/issues/75#issuecomment-842186329) some options like whether to use the ¿`save_best`? or not and what to consider as better (i.e. higher or lower)

daavoo:
Closes #89 

This P.R. extends existing summary to have `last`, `max` and `min` sections.

Example:

```python
for value in [0.1, 0.2, 0.15]:
    dvclive.log("m", value)
    dvclive.next_step()
```

Currently generates:

```json
{
    "step": 2,
    "m": 0.15
}
```

After this P.R. would generate:

```json
{
    "last": {
        "m": 0.15,
        "step": 2
    },
    "max": {
        "m": 0.2,
        "step": 2
    },
    "min": {
        "m": 0.1,
        "step": 0
    }
}
```

It requires also changes in:

- [ ] P.R. to `dvc` 
- [ ] P.R. to `dvc.org`

Not sure how this kind of changes should be addressed on the `dvc` side (just breaks a few tests in `test_live`)  because it will require to have pinned versions of `dvclive` in C.I but tests that depend on `dvclive` are currently being skipped in `dvc` C.I.

### DIFF
--- a/dvclive/metrics.py
+++ b/dvclive/metrics.py
@@ -31,6 +31,8 @@ class MetricLogger:
         self._html: bool = html
         self._summary = summary
         self._metrics: Dict[str, float] = OrderedDict()
+        self._max_values: Dict[str, float] = OrderedDict()
+        self._min_values: Dict[str, float] = OrderedDict()
         self._checkpoint: bool = checkpoint
 
         if resume and self.exists:
@@ -103,9 +105,8 @@ class MetricLogger:
 
     def next_step(self):
         if self._summary:
-            metrics = OrderedDict({"step": self._step})
-            metrics.update(self._metrics)
-            write_json(metrics, self.summary_path)
+            summary = self._get_summary()
+            write_json(summary, self.summary_path)
 
         if self._html:
             signal_file_path = get_signal_file_path()
@@ -141,9 +142,7 @@ class MetricLogger:
         metric_history_path = os.path.join(self.history_path, name + ".tsv")
         os.makedirs(os.path.dirname(metric_history_path), exist_ok=True)
 
-        nested_set(
-            self._metrics, os.path.normpath(name).split(os.path.sep), val,
-        )
+        self._metrics[name] = val
 
         ts = int(time.time() * 1000)
         d = OrderedDict([("timestamp", ts), ("step", self._step), (name, val)])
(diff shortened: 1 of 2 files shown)


pared:
This one is tricky, We need to think about how to handle the summary - do we want `dvc metrics show/diff` to show `min`/`max` values? If not this change will require some adjustment on DVC side. Maybe we should turn it off by default but let the user turn it on with `init` call?

daavoo:
To clarify, I'm not convinced about the current state of the P.R. either. I opened the P.R. to move forward the discussion from the issue towards discussing implementation details.

> This one is tricky, We need to think about how to handle the summary - do we want `dvc metrics show/diff` to show `min`/`max` values? If not this change will require some adjustment on DVC side. Maybe we should turn it off by default but let the user turn it on with `init` call?

I might be biased because I was testing this feature with `dvc exp` instead of `dvc metrics`. As `dvc exp` includes the `--include-metrics` / `--exclude-metrics` flags it didn't bother me having some not interesting values.

For a regular workflow using checkpoints and `dvc apply`, disabling `min` and `max` as default probably makes sense. However I do find those values useful in a workflow without checkpoints.

dberenbaum:
What about a syntax like `dvclive.init(summary=["latest", "min"])`? Users can choose one or more ways to summarize. If a single summary method is chosen, maybe the summary gets collapsed back to the current format.

`dvclive.log("m", value, summary=["latest", "min"])` might be better since It's likely that min/max will only be useful for certain metrics, and that users will want to track only the min loss but the max accuracy, for example. What does the current PR look like for multiple metrics where the step is different for the min/max values for each metric?

daavoo:
> `dvclive.log("m", value, summary=["latest", "min"])` might be better since It's likely that min/max will only be useful for certain metrics, and that users will want to track only the min loss but the max accuracy, for example. 

I still think that even for this kind of granularity it should be better configured in `init` rather than in `log`, in most integrations `dvclive.log` is called inside a loop so you would still need to get the associated `summary` configuration for each metric from somewhere.

> What does the current PR look like for multiple metrics where the step is different for the min/max values for each metric?

The step value where the min/max values were logged is currently being lost.  

---

An alternative syntax for supporting granular configuration could allow users to specify the desired summary per metric.

If none of the metrics are configured, we colapse to the original behavior:

```json
{
    "step": 2,
    "m": 0.15
}
```

But we allow the user to have granular support over each metric and associated summary preferences:

```yaml
summary:
    m: max
```

Resulting in something like:

```json
{
    "last": {
        "step": 2,
        "m": 0.15
    },
    "m": {
        "max": 0.2
    }
}
```

It might get tricky but we could also preserve information about the corresponding step:

```json
{
    "last": {
        "step": 2,
        "m": 0.15
    },
    "m": {
        "max": {
            "value": 0.2,
            "step": 1
        }
    }
}
```

PR #125, opened 2021-07-28
Added MMCV integration

### LINKED ISSUE #110: Save model in `MMCV`

### LINKED ISSUE #120: Update `MMCV` integration
Even though `path` is optional in `dvclive` (https://dvc.org/doc/dvclive/api-reference/init#parameters) the existing `MMCV` integration requires this parameter (https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/logger/dvclive.py#L30).

The integration should be updated making `path` optional for consistency and to prevent users unfamiliar with `dvclive` from encountering unexpected behavior (i.e. #102)

daavoo:
I think that, at least for now, it makes more sense to maintain the integrations inside the `dvclive` repo in order to ensure consistency and testing without depending on opening P.R. to external repositories.

Closes #120 
Closes #110

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,10 @@ jobs:
     - name: install libomp
       # required for xgboost
       if: runner.os == 'macOS'
-      run: brew install libomp
+      run: |
+        wget https://raw.githubusercontent.com/Homebrew/homebrew-core/679923b4eb48a8dc7ecc1f05d06063cd79b3fc00/Formula/libomp.rb -O $(find $(brew --repository) -name libomp.rb)
+        brew install ninja libomp
+        brew pin libomp        
     - name: run tests
       run: py.test -v --cov dvclive --cov-report=xml
     - name: upload coverage report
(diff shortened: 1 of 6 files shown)


pared (approved):
In this case some parameters (like optimizer learning rate) will be saved as metrics?

daavoo:
> In this case some parameters (like optimizer learning rate) will be saved as metrics?

Yep. It's kind of tricky because the **initial** learning rate and momentum values could be considered parameters in terms of `dvc params`.

However, given that those values (optimizer `learning_rate` and `momentum`) are scalars than could potentially change during training (i.e. there are hooks for updating those values [`lr_updater`](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/lr_updater.py) and [`momentum_updater`](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/hooks/momentum_updater.py)) they are considered metrics by `MMCV`.

pared:
I restarted build, thinking it was some random error, but it persists, I copied the branch locally, and can confirm that we get segmentation fault on macos

daavoo:
> I restarted build, thinking it was some random error, but it persists, I copied the branch locally, and can confirm that we get segmentation fault on macos

Thanks @pared . Luckily, now I can also reproduce the macos error locally xD. It's probably related with `mmcv`. I will check

daavoo:
> > I restarted build, thinking it was some random error, but it persists, I copied the branch locally, and can confirm that we get segmentation fault on macos
> 
> Thanks @pared . Luckily, now I can also reproduce the macos error locally xD. It's probably related with `mmcv`. I will check

Fixed

PR #130, opened 2021-07-31
Fix step param in dvclive.log method 

### LINKED ISSUE #129: metrics bug: `dvclive.log` with 0 as step param value, save wrong step to log file
During creating pytorch lightning logger, I found small bug  in `dvclive.log` method. When call with `0` as step param value, logger use value from internal state of loger instead passed by param, and this cause corruption in my log wehen step/epoch start from 0.
Due to pytorch lightning specific I must use `dvclive. log` method with step param, instead `dvclive.next_step`.

Corrupted log example:
```tsv
timestamp	step	train_loss_epoch
1627723290964	1499	-0.6670725345611572
1627723307309	1	-0.7409120202064514
1627723323716	2	-0.7462224960327148
```

Expected log:
```tsv
timestamp	step	train_loss_epoch
1627723290964	0	-0.6670725345611572
1627723307309	1	-0.7409120202064514
1627723323716	2	-0.7462224960327148
```

This bug is caused by conditional expression in `dvclive/metrics.py`  line 145:

Current version:
```py
if step:
    self._step = step
```
Main problem is that `0` is considered as `False`. This expression should be true for all `not None` values. 

Fixed version:
```py
if step is not None:
    self._step = step
```

naibatsuteki:
Improve conditional expression checking  step param value.
 
Closes #129

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -184,8 +184,8 @@ def test_infer_next_step(tmp_dir, mocker):
 def test_custom_steps(tmp_dir):
     dvclive.init("logs")
 
-    steps = [0, 62, 1000]
-    metrics = [0.9, 0.8, 0.7]
+    steps = [10, 0, 62, 1000]
+    metrics = [1.0, 0.9, 0.8, 0.7]
 
     for step, metric in zip(steps, metrics):
         dvclive.log("m", metric, step=step)
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -184,13 +184,24 @@ def test_infer_next_step(tmp_dir, mocker):
 def test_custom_steps(tmp_dir):
     dvclive.init("logs")
 
-    steps = [10, 0, 62, 1000]
-    metrics = [1.0, 0.9, 0.8, 0.7]
+    steps = [0, 62, 1000]
+    metrics = [0.9, 0.8, 0.7]
 
     for step, metric in zip(steps, metrics):
         dvclive.log("m", metric, step=step)
 
     assert read_history("logs", "m") == (steps, metrics)
+
+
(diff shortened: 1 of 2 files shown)


daavoo (changes requested):
Thanks for the contribution! 

Just a minor thing, could you:

- [ ] Add a regression test (test that fails because of the bug and passes after the fix)

tests/test_main.py
```diff
@@ -184,8 +184,8 @@ def test_infer_next_step(tmp_dir, mocker):
 def test_custom_steps(tmp_dir):
     dvclive.init("logs")
 
-    steps = [0, 62, 1000]
-    metrics = [0.9, 0.8, 0.7]
+    steps = [10, 0, 62, 1000]
+    metrics = [1.0, 0.9, 0.8, 0.7]
```
daavoo, line 188:
```suggestion
    steps = [0, 62, 1000]
    metrics = [0.9, 0.8, 0.7]
```

Maybe we could add a separate test with a more realistic scenario, for example:

```python
def test_log_reset_with_step_0(tmp_dir):
    for i in range(3):
        dvclive.log("train_m", 1, step=i)

    for i in range(3):
        dvclive.log("val_m", 1, step=i)

    assert read_history("dvclive", "train_m") == ([0, 1, 2], [1, 1, 1])
    assert read_history("dvclive", "val_m") == ([0, 1, 2], [1, 1, 1])
```

naibatsuteki: You are right.
Looks much better. 🙂

naibatsuteki: I added  separate test according to your suggestion. (Your example 1:1)

daavoo (approved):

Issue #128, opened 2021-07-29
`dvclive.init`: Future of arguments


daavoo:
Discussion regarding `dvclive.init` and it's number of arguments have emerged across multiple issues (i.e. https://github.com/iterative/dvclive/pull/69, https://github.com/iterative/dvclive/issues/75, https://github.com/iterative/dvclive/issues/121, https://github.com/iterative/dvclive/pull/124), specially those regarding the addition of new features, so I opened this one to concentrate all discussions. 

From an user perspective I prefer having as many *explicit* arguments as needed, as long as they are useful and well documented, rather than "hiding" the arguments under `**kwargs**` or even a `config` file. Maybe I'm biased because I feel that using APIs with too many arguments  (~8+) is not a real problem if all are actually useful, the library uses things like proper naming and types and you are using a modern IDE.  

I would need to think about it and consider other perspectives to actually have a strong preference,. 

---

Regardless of the options for extending arguments, I think it could be also worth to take a step back and consider the general interface to check if some of the arugments already existing could be removed/revisited in order to "make room" for new ones:

- `html`

Given that this argument is only used when *DVCLive* is used alongside *DVC*. I think it could be removed from `dvclive.init` and just let it be configured by `env.DVCLIVE_HTML`.

- `step` / `resume`

I'm not really sure what would be the indented use case for this argument outside it's interaction with `resume`. Unless I'm missing some other use case, `resume` and `step` could be merged into a single argument `resume_from: Optional[int] = None`?

- `summary`

How much would this actually be used to disable the summary? Would someone complain if we remove the argument and just always create it? 

If we come to the conclusion that users might not really want to have a summary, we could maybe rethink the `log`/`next_step` interaction to allow them to skip the summary:

If the user doesn't really want a summary, we could "ask" to always call `dvclive.log` with a explicit `step`, otherwise we assume that calling`next_step` (either explicitly or implicitly by calling `dvclive.log` twice without `step`) means that the user wants a summary.

---

I know all the above options would imply a significant "shakeup" to the existing API but I think that kind of discussion has, in the end, the same goal to the one about how to extend the arguments.

daavoo:
`summary` could be reused for configuring something like: https://github.com/iterative/dvclive/pull/124

pared:
`html` - I agree that we could get rid of that
`resume/step` - i never used them separately, and I think they could be merged
[...]

Issue #61, opened 2021-03-10
Proposal for more visual output options for standalone dvclive


dberenbaum:
Currently, dvclive without dvc produces just a tsv output tracking metrics progress. This doesn't seem like enough to make it worthwhile for users. There are at least a few drawbacks to the current output:
* It's not visual - some kind of plot, progress bar, etc. would be easier to read
* It's in a separate file that needs to be opened and refreshed as training progresses

Ideally, it would be great to have a nice dynamic plot that shows up in stdout during training and also optionally gets saved to a file at the end. More realistically, we might have different output options that users could choose or that dvclive could default to depending on what's available:
* Use dvc plots to visualize if installed (although this is still a static separate file)
* Use matplotlib or other python plotting libraries to render and update as training progresses
* Use stdout to make a very simplistic plot or to print out the metrics being written to tsv

Hopefully, there are better suggestions than these! Interested in thoughts on the general concept, and individual output types can be opened in separate issues if needed. Keeping dvclive as lightweight as possible should also remain a priority.

pared:
related to https://github.com/iterative/dvc/discussions/5980

daavoo:
Revisiting this. 

Now that we are moving towards depending on [dvc-render](https://github.com/iterative/dvc-render), I'm assuming it makes more sense to implement these different plotting options on `dvc-render` side, right?

> Use stdout to make a very simplistic plot or to print out the metrics being written to tsv

Would this be covered by #206 ?

dberenbaum:
Once dvclive can generate plots without dvc, we can close this issue.

PR #133, opened 2021-08-03
Removed html from public API

daavoo:

Discussed in #128 

`dvc.org` P.R: https://github.com/iterative/dvc.org/pull/2683

### DIFF
--- a/dvclive/__init__.py
+++ b/dvclive/__init__.py
@@ -12,7 +12,6 @@ def init(
     resume: bool = False,
     step: int = 0,
     summary: bool = True,
-    html: bool = True,
 ) -> MetricLogger:
     global _metric_logger  # pylint: disable=global-statement
     _metric_logger = MetricLogger(
@@ -20,7 +19,6 @@ def init(
         resume=resume,
         step=step,
         summary=summary,
-        html=html,
     )
     return _metric_logger
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -106,13 +106,13 @@ def test_nested_logging(tmp_dir):
         (False, False, False),
     ],
 )
-def test_html(tmp_dir, dvc_repo, html, signal_exists):
+def test_html(tmp_dir, dvc_repo, html, signal_exists, monkeypatch):
     if dvc_repo:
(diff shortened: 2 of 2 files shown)


pared (approved):

dberenbaum:
What if users have dvc installed but aren't running their dvclive code via dvc (running `python train.py` where `train.py` includes `dvclive` calls)? Should dvclive default to generating html unless dvc isn't installed?

pared:
@dberenbaum hmm, if we are considering it, we should make a new issue for that. HTML creation is not available in the current state of dvclive (since #44, where we started to create a signal file for dvc). The mechanism is as follows:


- dvclive detects it is in `dvc` repo
- create signal file
- dvc picks up the file and makes HTML

So, even before this change HTML would not be created without dvc.

dberenbaum:
Sounds good. Makes sense for it to be strictly a benefit of dvc integration. In the early days of dvclive, I think we had plans for it to have more standalone functionality without dvc, but it seems better to focus on the experience with dvc now.

pared:
@dberenbaum you are right, and before #44 we actually used dvc to plot. But in our efforts to make `dvclive` as dependency-less as possible we lost ability to plot "from within" `dvclive`. I might have not stressed that enough before finishing #44 and later it was not brought up.

If we want this we could possibly check if `dvc` is installed and use it for plots. That way we could plot "from within" and not require `dvc` in project requirements.

daavoo:
> @dberenbaum you are right, and before #44 we actually used dvc to plot. But in our efforts to make `dvclive` as dependency-less as possible we lost ability to plot "from within" `dvclive`. I might have not stressed that enough before finishing #44 and later it was not brought up.
> 
> If we want this we could possibly check if `dvc` is installed and use it for plots. That way we could plot "from within" and not require `dvc` in project requirements.

Perhaps having `dvclive.show` and make `dvc` just one of the potential plotting "backends", similar to what has been discussed in https://github.com/iterative/dvclive/issues/61?

dberenbaum:
That's a possibility. For now, IMO it's more important to focus on the experience with DVC, so I'm happy to leave this as is.

PR #134, opened 2021-08-08
Added LightGBM 

### LINKED ISSUE #132: integrations: LightGBM
It would be nice to have a `dvclive` integration with [LightGBM](https://lightgbm.readthedocs.io/en/latest/).

> LightGBM is a gradient boosting framework that uses tree based learning algorithms.

Example of custom callback:

https://github.com/microsoft/LightGBM/blob/master/examples/python-guide/advanced_example.py#L199

pacifikus:
Closes #132

PR to dvc docs: iterative/dvc.org/pull/2699

### DIFF
--- a/dvclive/lgbm.py
+++ b/dvclive/lgbm.py
@@ -1,0 +1,17 @@
+import dvclive
+
+
+class DvcLiveCallback:
+    def __init__(self, model_file=None):
+        super().__init__()
+        self.model_file = model_file
+
+    def __call__(self, env):
+        for eval_result in env.evaluation_result_list:
+            metric = eval_result[1]
+            value = eval_result[2]
+            dvclive.log(metric, value)
+        dvclive.next_step()
+
+        if self.model_file and env.iteration == env.end_iteration - 1:
+            env.model.save_model(self.model_file)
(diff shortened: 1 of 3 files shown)


dvclive/lgbm.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __call__(self, env):
+        for eval_result in env.evaluation_result_list:
+            metric = eval_result[1]
+            value = eval_result[2]
+            dvclive.log(metric, value)
+        dvclive.next_step()
+
+        if self.model_file and env.iteration == env.end_iteration - 1:
+            env.model.save_model(self.model_file)
```
daavoo, line 17: Would this only sve the model in the last iteration?? Would it make sense to save it at the end of each iteration, like in xgb:

https://github.com/iterative/dvclive/blob/master/dvclive/xgb.py#L12

Regardless, [dvclive.next_step](https://dvc.org/doc/dvclive/api-reference/next_step) should be called after the model saving in order to work as intended when dvclive is used with DVC checkpoints.

```suggestion
        if self.model_file:
            env.model.save_model(self.model_file)
        dvclive.next_step()
```

daavoo (changes requested):
Thanks for the contribution @pacifikus ! 

I just have a small comment regarding model saving.

pacifikus:
> Thanks for the contribution @pacifikus !
> 
> I just have a small comment regarding model saving.

Thank you for your comment! Changes have been made.

daavoo (approved):

PR #135, opened 2021-08-10
get rid of DvcLiveError calls

zili55:

### DIFF
--- a/dvclive/error.py
+++ b/dvclive/error.py
@@ -25,3 +25,10 @@ class ConfigMismatchError(DvcLiveError):
             f"Dvclive initialized in '{ml.dir}' conflicts "
             f"with '{os.environ[env.DVCLIVE_PATH]}' provided by DVC."
         )
+
+
+class InvalidMetricTypeError(DvcLiveError):
+    def __init__(self, name, val):
+        self.name = name
+        self.val = val
+        super().__init__(f"Metrics '{name}' has not supported type {val}")
--- a/dvclive/error.py
+++ b/dvclive/error.py
@@ -37,4 +37,4 @@ class InvalidMetricTypeError(DvcLiveError):
 class CannotCreateLogDirError(DvcLiveError):
     def __init__(self, directory):
         self.directory = directory
-        super().__init__(f"dvc-live cannot create log dir - '{directory}'")
+        super().__init__(f"dvclive cannot create log dir - '{directory}'")
--- a/dvclive/error.py
+++ b/dvclive/error.py
@@ -32,9 +32,3 @@ class InvalidMetricTypeError(DvcLiveError):
         self.name = name
         self.val = val
         super().__init__(f"Metrics '{name}' has not supported type {val}")
-
-
-class CannotCreateLogDirError(DvcLiveError):
-    def __init__(self, directory):
-        self.directory = directory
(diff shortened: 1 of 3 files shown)


dvclive/error.py
```diff
@@ -25,3 +25,16 @@ def __init__(self, ml: "MetricLogger"):
...
+    def __init__(self, name, val):
+        self.name = name
+        self.val = val
+        super().__init__(f"Metrics '{name}' has not supported type {val}")
+
+
+class CannotCreateLogDirError(DvcLiveError):
+    def __init__(self, directory):
+        self.directory = directory
+        super().__init__(f"dvc-live cannot create log dir - '{directory}'")
```
daavoo, line 40:
```suggestion
        super().__init__(f"dvclive cannot create log dir - '{directory}'")
```

dvclive/metrics.py
```diff
@@ -45,9 +45,7 @@ def __init__(
             try:
                 os.makedirs(self.dir, exist_ok=True)
             except Exception as exception:
-                raise DvcLiveError(
-                    "dvc-live cannot create log dir - '{}'".format(self.dir),
-                ) from exception
+                raise CannotCreateLogDirError(self.dir) from exception
```
daavoo, line 48: Not sure if we really want to raise a custom exception here as it seems that it would be related to Python specific errors and not with any dvclive specific logic. 

What would be the scenarios where this exception will be raised?? I can think of having a `PermissionError` where I think the default Python exception would be already clear enough.

cc @pared

daavoo (commented):
Thanks for the contribution @zili55 !

I have some doubts regarding the addition of `CannotCreateLogDirError` but they are not strictly related with this changes because the same doubts apply to the previous `DvcLiveError`

daavoo (changes requested):
Edited.

- [x] Remove CannotCreateLogDirError
- [x] P.R. to `dvc.org` updating https://dvc.org/doc/dvclive/api-reference/log#exceptions

dvclive/metrics.py
zili55: yep, you're right, only `PermissionError` can occure

daavoo: So maybe we could remove `CannotCreateLogDirError` as part of this P.R. (still makes sense under the scope of "get rid of DvcLiveError calls")  and revisit it for considering a custom exception in the future

zili55: So we don't need to raise `PermissionError`, and write a test for it ?

daavoo: Yep. Just let that be handled by Python

daavoo (approved):
Thanks @zili55 ! Could you address: 

https://github.com/iterative/dvclive/pull/135#pullrequestreview-726115889

zili55:
@daavoo P.R. to dvc.org created: https://github.com/iterative/dvc.org/pull/2707

daavoo:
> @daavoo P.R. to dvc.org created: [iterative/dvc.org#2707](https://github.com/iterative/dvc.org/pull/2707)

Nice
> and maybe from here https://dvc.org/doc/dvclive/api-reference/init#exceptions we should remove that exception

Indeed, could you add that change in the same P.R.?

PR #137, opened 2021-08-13
Huggingface integration

### LINKED ISSUE #83: integrations: HuggingFace transformers
It would be nice to have a `dvclive` integration with the popular repository [transformers](https://github.com/huggingface/transformers/) from huggingface.

https://github.com/huggingface/transformers/blob/master/src/transformers/integrations.py
https://huggingface.co/transformers/main_classes/callback.html#available-callbacks

- [x] Update docs in [dvc.org](https://github.com/iterative/dvc.org)

pacifikus:
Closes #83

### DIFF
--- a/dvclive/huggingface.py
+++ b/dvclive/huggingface.py
@@ -13,17 +13,6 @@ class DvcLiveCallback(TrainerCallback):
         super().__init__()
         self.model_file = model_file
 
-    def on_evaluate(
-        self,
-        args: TrainingArguments,
-        state: TrainerState,
-        control: TrainerControl,
-        **kwargs
-    ):
-        metrics = kwargs["metrics"]
-        for key, value in metrics.items():
-            dvclive.log(key, value)
-
     def on_log(
         self,
         args: TrainingArguments,
@@ -33,8 +22,7 @@ class DvcLiveCallback(TrainerCallback):
     ):
         logs = kwargs["logs"]
         for key, value in logs.items():
-            if "train_" in key:
-                dvclive.log(key, value)
+            dvclive.log(key, value)
 
             if self.model_file:
                 model = kwargs["model"]
--- a/dvclive/huggingface.py
+++ b/dvclive/huggingface.py
@@ -24,8 +24,19 @@ class DvcLiveCallback(TrainerCallback):
         for key, value in metrics.items():
             dvclive.log(key, value)
 
-        if self.model_file:
-            model = kwargs["model"]
-            model.save_pretrained(self.model_file)
+    def on_log(
+        self,
+        args: TrainingArguments,
+        state: TrainerState,
+        control: TrainerControl,
+        **kwargs
+    ):
+        logs = kwargs["logs"]
+        for key, value in logs.items():
+            if "train_" in key:
+                dvclive.log(key, value)
 
+            if self.model_file:
+                model = kwargs["model"]
+                model.save_pretrained(self.model_file)
         dvclive.next_step()
--- a/dvclive/huggingface.py
+++ b/dvclive/huggingface.py
@@ -1,0 +1,31 @@
+from transformers import (
+    TrainerCallback,
+    TrainerControl,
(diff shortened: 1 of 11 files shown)


daavoo (changes requested):
Thanks for the contribution @pacifikus ! It looks great.

Could you:

- [x] Open a P.R. in https://github.com/iterative/dvc.org adding HuggingFace to the list of https://dvc.org/doc/dvclive/user-guide/ml-frameworks . You should be able to use some of the other pages as reference

dvclive/huggingface.py
```diff
@@ -0,0 +1,31 @@
...
+
+import dvclive
+
+
+class DvcLiveCallback(TrainerCallback):
+    def __init__(self, model_file=None):
+        super().__init__()
+        self.model_file = model_file
+
+    def on_evaluate(
```
daavoo, line 16: Btw, is there a reason why the Callbacks that are part of HuggingFace use `on_log` but we are using `on_evaluate`:

https://huggingface.co/transformers/_modules/transformers/integrations.html#CometCallback

This callback does look simpler than those other implementations, but I'm just curious about that

pacifikus: Thank you for your comment!

PR in dvc.org: https://github.com/iterative/dvc.org/pull/2718

Yes, there is a reason why the Callbacks that are part of HuggingFace use `on_log` instead of `on_evaluate`. As you can see in https://huggingface.co/transformers/main_classes/callback.html#trainercallback, metrics are only available in the `on_evaluate` event. Thus, we can get the loss values both at the train and eval from `on_log`, but eval metrics only from `on_evaluate`.

The `on_evaluate` event has its disadvantage: we cannot get the training loss during this event, only the evaluation loss.

daavoo: Thanks for the explanation. Would it make sense to extend `DvcLiveCallback` to  use both `on_log` (for trainand eval loss) and `on_evaluate` (for eval metrics) ?

daavoo (approved):

dvclive/huggingface.py
pacifikus: Yes, I think it's a good idea. I did it

daavoo: I think that the huggingface docs might be a little misleading. I just debug the test locally and it seems that the same values available as `metrics` inside `on_evaluate` are previously being passed as `logs` to `on_log`.

dvclive/huggingface.py
```diff
@@ -0,0 +1,42 @@
...
+    ):
+        logs = kwargs["logs"]
+        for key, value in logs.items():
+            if "train_" in key:
+                dvclive.log(key, value)
+
+            if self.model_file:
+                model = kwargs["model"]
+                model.save_pretrained(self.model_file)
+        dvclive.next_step()
```
daavoo, line 42: This produces the same result:

```suggestion
    def on_log(
        self,
        args: TrainingArguments,
        state: TrainerState,
        control: TrainerControl,
        **kwargs
    ):
        logs = kwargs["logs"]
        for key, value in logs.items():
            dvclive.log(key, value)

            if self.model_file:
                model = kwargs["model"]
                model.save_pretrained(self.model_file)
        dvclive.next_step()
```

dvclive/huggingface.py
pacifikus: Yes, you're right, it's strange, but metrics logs are also written in `on_log` event. Thank you for your comment!

pacifikus: Hmm, now for different operating systems, different `len(first(logs.values()))` values in the test are obtained

daavoo: Well that is strange, I will debug locally the macOS test case which is the one failing

daavoo: Still don't understand what the bug was. If I have to guess might be related with `first` behaving differently in MacOS. Anyway I made the test more explicit and it now passes 🎉

PR #139, opened 2021-08-23
Catalyst integration

### LINKED ISSUE #112: integrations: Catalyst
It would be nice to have a `dvclive` integration with [Catalyst](https://github.com/catalyst-team/catalyst):

> Catalyst is a PyTorch framework for Deep Learning Research and Development. It focuses on reproducibility, rapid experimentation, and codebase reuse so you can create something new rather than write yet another train loop.
Break the cycle – use the Catalyst!

https://github.com/catalyst-team/catalyst/tree/master/catalyst/loggers
https://catalyst-team.github.io/catalyst/core/logger.html

pacifikus:
Closes #112

PR PR in dvc.org: iterative/dvc.org#2748

### DIFF
--- a/dvclive/catalyst.py
+++ b/dvclive/catalyst.py
@@ -1,0 +1,28 @@
+from catalyst.core.callback import Callback, CallbackOrder
+
+import dvclive
+
+
+class DVCCallback(Callback):
+    def __init__(self, model_file=None):
+        super().__init__(order=CallbackOrder.external)
+        self.model_file = model_file
+
+    def on_epoch_end(self, runner) -> None:
+        step = runner.stage_epoch_step
+
+        for loader_key, per_loader_metrics in runner.epoch_metrics.items():
+            for key, value in per_loader_metrics.items():
+                key = key.replace("/", "_")
+                key += "_" + loader_key
+                dvclive.log(key, float(value), step)
+
+        if self.model_file:
+            checkpoint = runner.engine.pack_checkpoint(
+                model=runner.model,
+                criterion=runner.criterion,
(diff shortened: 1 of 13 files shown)


dvclive/catalyst.py
```diff
@@ -0,0 +1,28 @@
...
+        self.model_file = model_file
+
+    def on_epoch_end(self, runner) -> None:
+        step = runner.stage_epoch_step
+
+        for loader_key, per_loader_metrics in runner.epoch_metrics.items():
+            for key, value in per_loader_metrics.items():
+                key = key.replace("/", "_")
+                key += "_" + loader_key
+                dvclive.log(key, float(value), step)
```
daavoo, line 18: I think it would be nice to have `loader_key` ("train/" and "valid/" ) as prefix as some other frameworks also use this schema.

The `/` replace in `key` still needs to happen due to key duplication.

```suggestion
                key = key.replace("/", "_")
                dvclive.log(f"{loader_key}/{key}", float(value), step)
```

daavoo (commented):
Thanks for the contribution @pacifikus ! 

Just one minor comment regarding `loader_key` prefix

dvclive/catalyst.py
pacifikus: Yes, you're right, it's a good suggestion

daavoo:
> Thanks for the contribution @pacifikus !
> 
> Just one minor comment regarding `loader_key` prefix

Forgot to comment that tests should be updated to match the new behavior

daavoo (approved):

PR #166, opened 2021-09-20
Image support

### LINKED ISSUE #150: `dvclive.log`: Support images
Now that dvc has [support for logging images](https://github.com/iterative/dvc/pull/6431), it makes sense to also add support on the `dvclive` side.

A few questions arise:

## `dvclive.log` or `dvclive.log_image`

Should the existing `dvclive.log` method be extended to accept a new input type (Image-like) or should we add a new separate method ? 

For reference:

- wandb has a single [`log` method](https://docs.wandb.ai/guides/track/log) that handles every type.
- CometML has a separate [`log_image`](https://www.comet.ml/docs/python-sdk/APIExperiment/#apiexperimentlog_image)
- MLFlow uses a separate general method [`log_artifact`](https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.log_artifact)
- Neptune has a single [`log` method](https://docs.neptune.ai/api-reference/field-types#log-2) that can receive images (`File.as_image`).

## Valid inputs

What kind of objects should be accepted to log as images? How would be handled by `dvclive`?

On the dvc side, the only condition is that the image is stored in a [valid format ](https://github.com/iterative/dvc/blob/master/dvc/render/image.py#L61). 

On `dvclive` we should consider what we will accept as inputs and how to handle them. 

The more common Image-like objects I can think of are:

- 2D/3D array/tensor (i.e. Numpy, PyTorch) 
- Pillow Image
- Matplotlib Figure

In addition we need to consider if we want to also accept a string indicating a path to an existing image file (i.e. `dvclive.log_image("foo.png")`), for cases where the user takes care of saving the image by itself.

daavoo:

Close #150

- Added support for logging PIL Image and numpy arrays.


dvclive/data/image_pil.py
```diff
@@ -0,0 +1,36 @@
...
+    def dump(self, val, step) -> None:
+        super().dump(val, step)
+        output_path = Path(str(self.output_path).format(step=step))
+        output_path.parent.mkdir(exist_ok=True, parents=True)
+
+        val.save(output_path)
+
+    @property
+    def summary(self):
+        return {}
```
daavoo, line 36: Should image be reflected in the summary?

daavoo (commented):
I added an option for using a `step` placeholder in the output path of the image. With the current html layout, overriding the image at each step is kind of the only option that looks nice.

Overriding (`dvclive.log("image.png", img)`):

```python
r = 5
for i in range(r):

    img = Image.new('RGB', (200,200), (250,250,250))
    draw = ImageDraw.Draw(img)
    font = ImageFont.truetype("OpenSans-Regular.ttf", 100)
    draw.text((20, -5),str(i),(0,0,0), font=font)

    dvclive.log("image.png", img)
    dvclive.log("loss", (1-i/r) + (np.random.random() / 10))
    dvclive.log("accuracy", (i/r) + (np.random.random() / 10))
    dvclive.next_step()
```
ezgif-3-c8c7158d51cb

Using the placeholder (`dvclive.log("image_{step}.png", img)`):
```python
r = 5
for i in range(r):

    img = Image.new('RGB', (200,200), (250,250,250))
    draw = ImageDraw.Draw(img)
    font = ImageFont.truetype("OpenSans-Regular.ttf", 100)
    draw.text((20, -5),str(i),(0,0,0), font=font)

    dvclive.log("image_{step}.png", img)
    dvclive.log("loss", (1-i/r) + (np.random.random() / 10))
    dvclive.log("accuracy", (i/r) + (np.random.random() / 10))
    dvclive.next_step()
```
ezgif-3-693d6fdc97f8

dvclive/data/image_pil.py
pared: We could provide the URL of the latest.

dberenbaum:
We may want to wait on both dvc and studio to solve the problem of how to show multiple related images before prioritizing image support in dvclive.

I would focus more on where to save the files and how to organize them first. Save all the images in a way that makes sense and is easy to use with or without dvc. This fits with the stated dvclive principles. Visualization can be built later.

setup.py
```diff
@@ -43,8 +43,9 @@ def run(self):
 hugginface = ["transformers", "datasets"]
 catalyst = ["catalyst"]
 fastai = ["fastai"]
+image = ["pillow"]
```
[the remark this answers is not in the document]
daavoo: Added error throwing.

daavoo:
> We may want to wait on both dvc and studio to solve the problem of how to show multiple related images before prioritizing image support in dvclive.
> 
> I would focus more on where to save the files and how to organize them first. Save all the images in a way that makes sense and is easy to use with or without dvc. This fits with the stated dvclive principles. Visualization can be built later.

I feel like the current implementation of the P.R. kind of already covers this. 

The way I see it, any changes on DVC/Studio regarding how to visualize the images won't have an impact on the current state of the P.R. as it will be most likely based on where the images are saved. 

The `step` placeholder might be used by the user (or enforced by ourselves, if required) to achieve kind of all possible variations that DVC/Studio might require: override on step, use step prefix/suffix, use step folder.

dberenbaum:
So the image file is being overwritten at every step, right? That seems okay for DVC, but not great for non-DVC use cases since they will lose the images from all previous steps.

pared:
In order to let users write down images for each step we need to either make output for the particular key a directory or implement the structure that @daavoo proposes in iterative/dvc#6628.

daavoo:
> So the image file is being overwritten at every step, right? That seems okay for DVC, but not great for non-DVC use cases since they will lose the images from all previous steps.

If they don't include the `step` yes, but we can emphasize that option in the docs, right?. The following would work as expected:
```python 
dvclive.log(f"image_{step}.png", image)
```

Anyhow, we kind of already have a similar ux regarding `model_file` it gets overwritten as it is intended to be mainly used alongside DVC checkpoints.

> In order to let users write down images for each step we need to either make output for the particular key a directory or implement the structure that @daavoo proposes in [iterative/dvc#6628](https://github.com/iterative/dvc/pull/6628).

Not sure if I am missing something but why would be having a separate subdir a requirement?  Using `dvclive.log(f"image_{step}.png` or `dvclive.log(f"{step}/image.png")` would currently work, it is just not pretty when showed in the `html`.

daavoo:
I removed the subdir part from this P.R. as it is not a strict requirement and could be addressed in a separate P.R.

pared:
> Not sure if I am missing something but why would be having a separate subdir a requirement? Using dvclive.log(f"image_{step}.png or dvclive.log(f"{step}/image.png") would currently work, it is just not pretty when shown in the HTML.

Right, its not a strict requirement. If we document the `{step}` properly I think this is a  good way to let users decide how they want to store their images.

pared (approved):

dberenbaum:
TBH I'm pretty torn. The flexibility is nice, but it puts more burden on the user and moves away from a canonical tree structure. Would you recommend "image_{step}.png" for both DVC and non-DVC workflows?

> Anyhow, we kind of already have a similar ux regarding `model_file` it gets overwritten as it is intended to be mainly used alongside DVC checkpoints.

This seems slightly different to me:
* A model isn't an artifact that I would want to see in one place across multiple steps.
* Model file location isn't necessarily within the dvclive directory.
* Model saving is specific to certain frameworks as a convenience.

On the other hand, it might be nice to make model saving more similar so that, for example, integration with dvc could manage the model output instead of needing to specify this separately in `dvc.yaml`.

pared:
> On the other hand, it might be nice to make model saving more similar so that, for example, integration with dvc could manage the model output instead of needing to specify this separately in dvc.yaml.

This brings us to the question how would we save the model? Do we `log` it? if so, how do we make sure DVC understands its a model and not a plot/metric? Make a special subdir in `dvclive` results? Do we need a special method for saving assets that are not metrics? Auto-type detection seems risky in this case as models come in all shapes and colors.

daavoo:
I updated the code to use `{live.dir}/images/{live.step}/{filename}` as default in any case.

dberenbaum:
Is there a docs PR for this @daavoo?

daavoo:
> Is there a docs PR for this @daavoo?

https://github.com/iterative/dvc.org/issues/3013#issuecomment-964199190

PR #167, opened 2021-09-29
Removed public methods

### LINKED ISSUE #36: api: make dvclive thread safe
Currently only safe way to use `dvclive` in threads is to manually create `MericsLogger` class.

In other case, for example:
```
import dvclive
from concurrent.futures import ThreadPoolExecutor
from time import sleep

def log_some(index):
    dvclive.init(str(index))
    metrics = [index * i for i in range(10)]
    for m in metrics:
        dvclive.log(f"metric-{str(index)}",m)
        sleep(0.1)
        dvclive.next_step()


with ThreadPoolExecutor(max_workers=10) as e:
    e.map(log_some, [1,2,7,13,17])
```

we will get mixed up results. Research whether it is possible to keep current API functionality and be thread safe at the same time.

### LINKED ISSUE #156: `dvclive.log` "side effects"
I would like to discuss the current behavior of `dvclive.log` which, in certain situations, replaces `dvclive.init` and calls `dvclive.next_step`. 

It is not clear for me the tradeoff between benefits and potential problems of these "side effects" and would like to know about potential benefits I might be missing.

Assuming that the primary usage of `dvclive` would be through our provided integrations, end users won't really care about how we handle the workflow inside those integrations. For those using the API, the current behavior might be confusing (https://github.com/iterative/dvc.org/issues/2773).

I think that we should enforce a canonical workflow (`init` -> `log` -> `next_step`) in both docs and integrations by:

- [x] #159 

- [x] #162

### LINKED ISSUE #159: Public methods vs MetricsLogger instantiation
Not sure if this belongs in this issue, but what is the advantage of `dvclive.init()` over `logger = dvclive.MetricsLogger()` as the canonical workflow? Explicitly calling `MetricsLogger` seems more transparent, removes an extra layer of abstraction, and makes it easier to manage multiple loggers.

_Originally posted by @dberenbaum in https://github.com/iterative/dvclive/issues/156#issuecomment-916874901_

daavoo:

Closes #36
Closes #156 
Closes #159 

Related docs P.R: https://github.com/iterative/dvc.org/pull/2879

---

Removed `dvclive.*` methods and `lazy_init` side effect in favor of "canonical" workflow:

```python
dvclive = DVCLive()

for i in range(3):
    dvclive.log("foo", 1)
    dvclive.next_step()
```

Since instantiation is now mandatory, I changed `path` making it an optional argument so that: `dvclive = DVCLive()` will use the `path` specified by DVC (if exists) otherwise it will use the default (`dvclive`). 

`dvclive = DVCLive("custom_path")` still works as before, raising a `ConfigMismatchError` when appropriate.


pared:
So far looks good to me, I think we might want to ask @dmpetrov for opinion, as this changes initially planned usage of DVClive.

dberenbaum:
Looks good to me as well, and @dmpetrov confirmed he is fine with the change.

One really minor note is whether we actually want to change from `MetricLogger` -> `DVCLive`. I wonder if `MetricLogger` is more descriptive, and I worry a little that the capitalization of `DVCLive` (vs. `DvcLive`) is slightly trickier for people to remember if they don't have autocompletion. However, I'm fine with the change if you think it's better.

daavoo:
> Looks good to me as well, and @dmpetrov confirmed he is fine with the change.
> 
> One really minor note is whether we actually want to change from `MetricLogger` -> `DVCLive`. I wonder if `MetricLogger` is more descriptive, and I worry a little that the capitalization of `DVCLive` (vs. `DvcLive`) is slightly trickier for people to remember if they don't have autocompletion. However, I'm fine with the change if you think it's better.

Agree that the capitalization might be a little tricky. Switched back to MetricLogger, not a big fan of that but in the docs we can just remark the following to use `dvclive` as a convention for the name of the instance:

```python
from dvclive import MetricLogger

dvclive = MetricLogger()
```

dberenbaum:
Now I'm not sure how much better `MetricLogger()` is since I could easily see people confusing it with `MetricsLogger()` 🤦  . Other ideas: `LiveLogger()` or `Live()`. Again, I'm throwing out ideas and my rationale, but I'm fine with whatever you think is best.

daavoo:
> Now I'm not sure how much better `MetricLogger()` is since I could easily see people confusing it with `MetricsLogger()` 🤦 . Other ideas: `LiveLogger()` or `Live()`. Again, I'm throwing out ideas and my rationale, but I'm fine with whatever you think is best.

🤣 I have just accidentally typed `MetricsLogger` in the docs and was trying to find out why the linker wasn't working

I like `Live` it is short and matches the `dvc.yaml` section 🤔

pared (approved):
I personally prefer `MetricLogger` as it's quite explanatory. On the other hand, `Live` is short, sweet, and easy to write (no problem with "Was it MetricsLogger or MetricLogger?" So it has the upper hand.

daavoo:
> I personally prefer `MetricLogger` as it's quite explanatory. On the other hand, `Live` is short, sweet, and easy to write (no problem with "Was it MetricsLogger or MetricLogger?" So it has the upper hand.

Let's see how it goes, as we commented in `0.X` land we can always go back and change `Live` for other thing

PR #168, opened 2021-09-30
Add new callback to DVCLive

### LINKED ISSUE #85: integrations: pytorch lightning
It would be nice to have a `dvclive` integration with the popular training framework [PyTorch Lightning](https://www.pytorchlightning.ai/)

https://github.com/PyTorchLightning/pytorch-lightning/tree/master/pytorch_lightning/loggers
https://pytorch-lightning.readthedocs.io/en/1.2.9_a/common/loggers.html

sirily:

Close #85 
- Added callback for pytorch lightning
- Didn't update docs. I'll do it when you accept my changes. It's my first PR, so I'm not sure, that I'm doing it right ¯\_(ツ)_/¯

### DIFF
--- a/dvclive/lightning.py
+++ b/dvclive/lightning.py
@@ -18,7 +18,6 @@ class DvcLiveLogger(LightningLoggerBase):
         experiment=None,
     ):
 
-
         super().__init__()
         self._prefix = prefix
         self._dvclive_init = dict(
--- a/dvclive/lightning.py
+++ b/dvclive/lightning.py
@@ -10,11 +10,7 @@ import dvclive
 
 class DvcLiveLogger(LightningLoggerBase):
     def __init__(
-        self,
-        run_name: Union[str, int],
-        prefix="",
-        experiment=None,
-        **kwargs
+        self, run_name: Union[str, int], prefix="", experiment=None, **kwargs
     ):
 
         super().__init__()
--- a/dvclive/lightning.py
+++ b/dvclive/lightning.py
@@ -1,0 +1,80 @@
+from typing import Dict, Optional, Union
+
+from pytorch_lightning.loggers import LightningLoggerBase
+from pytorch_lightning.loggers.base import rank_zero_experiment
+from pytorch_lightning.utilities import rank_zero_only
+from torch import is_tensor
+
+import dvclive
+
+
+class DvcLiveLogger(LightningLoggerBase):
+    def __init__(
+        self,
+        run_name: Union[str, int],
+        save_dir: Optional[str] = None,
+        summary: Optional[bool] = True,
+        prefix="",
+        experiment=None,
+    ):
+
+        if dvclive is None:
+            raise ImportError(
+                "You want to use `dvclive` logger which is not installed yet,"
+                " install it with `pip install dvclive`."
+            )
+
+        super().__init__()
+        self._prefix = prefix
+        self._dvclive_init = dict(
+            path=save_dir,
+            summary=summary,
+        )
+        self._experiment = experiment
+        self._version = run_name
+
+    @property
+    def name(self):
+        return "DvcLiveLogger"
+
+    @rank_zero_only
+    def log_hyperparams(self, params, *args, **kwargs):
+        pass
+
+    @property
+    @rank_zero_experiment
+    def experiment(self):
+        r"""
+        Actual DVCLive object. To use DVCLive features in your
+        :class:`~LightningModule` do the following.
+        Example::
+            self.logger.experiment.some_dvclive_function()
+        """
+        if self._experiment is not None:
+            return self._experiment
+        else:
+            assert (
+                rank_zero_only.rank == 0
+            ), "tried to init log dirs in non global_rank=0"
+            self._experiment = dvclive.init(**self._dvclive_init)
+
+        return self._experiment
+
+    @property
+    def version(self):
+        return self._version
+
+    @rank_zero_only
+    def log_metrics(
+        self, metrics: Dict[str, float], step: Optional[int] = None
+    ):
+        assert (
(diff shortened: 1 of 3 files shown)


dvclive/lightning.py
```diff
@@ -0,0 +1,80 @@
...
+        summary: Optional[bool] = True,
+        prefix="",
+        experiment=None,
+    ):
+
+        if dvclive is None:
+            raise ImportError(
+                "You want to use `dvclive` logger which is not installed yet,"
+                " install it with `pip install dvclive`."
+            )
```
daavoo, line 25: We can remove this given that the callback is distributed as part of `dvclive`.

```suggestion
```

dvclive/lightning.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class DvcLiveLogger(LightningLoggerBase):
+    def __init__(
+        self,
+        run_name: Union[str, int],
+        save_dir: Optional[str] = None,
+        summary: Optional[bool] = True,
+        prefix="",
+        experiment=None,
```
daavoo, line 18: In order to match other integrations are going to do (#167) I would just use `**kwargs` for `dvclive` args :

```suggestion
        prefix="",
        experiment=None,
        **kwargs
```

dvclive/lightning.py
```diff
@@ -0,0 +1,80 @@
...
+                "You want to use `dvclive` logger which is not installed yet,"
+                " install it with `pip install dvclive`."
+            )
+
+        super().__init__()
+        self._prefix = prefix
+        self._dvclive_init = dict(
+            path=save_dir,
+            summary=summary,
+        )
```
daavoo, line 32:
```suggestion
        self._dvclive_init = kwargs
```

dvclive/lightning.py
```diff
@@ -0,0 +1,80 @@
...
+    def version(self):
+        return self._version
+
+    @rank_zero_only
+    def log_metrics(
+        self, metrics: Dict[str, float], step: Optional[int] = None
+    ):
+        assert (
+            rank_zero_only.rank == 0
+        ), "experiment tried to log from global_rank != 0"
```
daavoo, line 73: Do we really need this assertions if we are using the `@rank_zero_only` decorator?

```suggestion
```

dvclive/lightning.py
```diff
@@ -0,0 +1,80 @@
...
+from pytorch_lightning.utilities import rank_zero_only
+from torch import is_tensor
+
+import dvclive
+
+
+class DvcLiveLogger(LightningLoggerBase):
+    def __init__(
+        self,
+        run_name: Union[str, int],
```
daavoo, line 14: What is this used for? 

I see that it is used for the `version` property, just wondering what Pytorch Lightning uses that property for  🤔

daavoo (changes requested):
Thanks for the contribution @sirily ! 🚀 

Just some minor changes required

dvclive/lightning.py
daavoo: Per the suggestion above

dvclive/lightning.py
sirily: I've seen this line in several Lightning loggers. For example: [wandb logger](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loggers/wandb.py#L212), [comet logger](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loggers/comet.py#L241), [neptune logger](https://github.com/PyTorchLightning/pytorch-lightning/blob/master/pytorch_lightning/loggers/neptune.py#L469)
I assume that unexpected situations may occur on distributed GPUs, so I left this assertion just in case

dvclive/lightning.py
sirily: Usually it is the name of the run, that can be seen in the UI of logger.
Inside Pytorch Lightning I've seen the use of this property only in a method that form data for progress bar. But I have never seen any version numbers in the progress bar itself in practice

dvclive/lightning.py
daavoo: Got it. Thanks!

tests/test_lightning.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def configure_optimizers(self):
+        return Adam(self.parameters(), lr=1e-3)
+
+
+def test_lightning_integration():
+    # init model
+    model = LitMNIST()
+    # init logger
+    dvclive_logger = DvcLiveLogger("test_run", save_dir="logs")
```
daavoo, line 77: Tests were failing per:
```suggestion
    dvclive_logger = DvcLiveLogger("test_run", path="logs")
```

Or we could just completely omit that arg and use the default (`"dvclive"`):

```python
    dvclive_logger = DvcLiveLogger("test_run")
```
And update lines bellow replacing `"logs"` with `"dvclive"`

dvclive/lightning.py
daavoo: Thanks for the explanation. 

TBH the property kind of feels useless in DVCLive (+ DVC) context as DVC already generates a unique experiment name for each run. 

I wonder if we could just make it optional with a reasonable default ("dvclive_run"?)?

sirily: I agree, it will be useless in most cases. Made it optional with "dvclive_run" default

tests/test_lightning.py
sirily: Changed this in new commit

daavoo (approved):

sirily:
Created [the issue](https://github.com/iterative/dvc.org/issues/2882) about documentation update. I'll create a PR to dvc.org tomorrow

PR #174, opened 2021-10-07
Keras load model

daavoo:

`keras` implementation of https://github.com/iterative/dvclive/issues/140

### DIFF
--- a/tests/test_keras.py
+++ b/tests/test_keras.py
@@ -107,9 +107,7 @@ def test_keras_load_model_on_resume(
     assert load_weights.call_count == save_weights_only
 
 
-def test_keras_no_resume_skip_load(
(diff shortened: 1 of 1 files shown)


dvclive/keras.py
```diff
@@ -14,6 +19,17 @@ def __init__(
         self.save_weights_only = save_weights_only
         self.dvclive = Live(**kwargs)
 
+    def on_train_begin(self, logs=None):
+        if (
+            self.dvclive._resume
+            and self.model_file is not None
+            and os.path.exists(self.model_file)
```
dberenbaum, line 26: Maybe it would be good to also have tests for each of these conditions (for example, model loading isn't called if `dvclive._resume == False`)?

dberenbaum (approved):

pared (approved):

Issue #5, opened 2020-11-19
integrations: sklearn


pared:
Seems we should be supporting at least few popular frameworks.

Considering their popularity, we should probably start with:
- [x] keras - we have initial implementation
- [x] ~~sklearn~~
- [x] xgboost

Worth considering:
- [x] - ~FastAi~ - #136 
- [x] - pytorch lightning

`TF` and `PyTorch` - it seems to me that using their pure form is done when users need highly custom models, and probably in that cases they will be able to handle `dvclive` by hand.
@dmpetrov did I miss some popular framework?

EDIT:
crossing out FastAi as it has its own issue now

pared:
@dberenbaum Yes, after digging through documentation, it seems to me that in general, learning algorithms divide to those which utilize `fit` method and both `fit` and `partial_fit`. It does not seem to me that we can provide integration for "only `fit`" models, and in case of `partial_fit` models, the workflow will probably look more like `torch` one, which in my opinion does not require any integration, as its created manually.

The only place I could probably see some integration is methods accepting `scoring` param which can be `Callable` but it seems to me it would be really hard to define how such integration could work.

daavoo:
I am considering to work on the integration with `pytorch-lightning` but I'm not sure about where to contribute the new logger (i.e. this repository or `pytorch-lightning` itself). See https://github.com/iterative/dvclive/issues/70#issuecomment-811868255

daavoo:
I added an integration with [mmcv](https://github.com/open-mmlab/mmcv):

https://github.com/open-mmlab/mmcv/pull/1075

pared:
@daavoo Thats a great news! Can we do something to help with that pull request?

daavoo:
> @daavoo Thats a great news! Can we do something to help with that pull request?

It has been already approved so I think it will be merged soon, thanks!

daavoo:
I think it might be a good idea to have separated issues for each integration in order to better track the progress and have specific discussions for each one. 

I.e: https://github.com/iterative/dvclive/issues/83

pared:
@daavoo  That is right, in the beggining we intended it to be an umbrella issue, since singular implementations seemed like easy tasks. As `sklerarn` example shows, we should probably track each integration separately.

For future reference:
Changing the name of the issue for `sklearn`. Other integrations issues should be created as separate issues.

Issue #82, opened 2021-05-31
`dvclive.log`: log array, tensors and similar objects


aguschin:
AFAIK, some experiment management libraries (e.g., w&b) can log not only simple python objects, but also more complex, few of them being numpy arrays, torch tensors, tensorflow tensors, etc. It seems like a general functionality which can help a lot of users and make the logging process more convenient for them.

Also, surprisingly to me, w&b client library doesn't have numpy/torch/tensorflow dependencies: https://github.com/wandb/client/blob/master/requirements.txt

Could we implement something similar for the dvclive? If the answer is positive, I'm gonna do a little research and post more examples on what custom types would be great to support.

dberenbaum:
Logging more data types sounds good. I found this link for wandb:

https://docs.wandb.ai/guides/track/log

Is this what you were thinking? Did you have a specific scenario in mind, and how would you like that data logged?

---

> Also, surprisingly to me, w&b client library doesn't have numpy/torch/tensorflow dependencies: https://github.com/wandb/client/blob/master/requirements.txt

They seem to leave it to the user to have the correct dependencies installed, which is similar to the mlflow approach (see https://github.com/iterative/dvclive/issues/25#issuecomment-768623284). Maybe we should have a separate issue to discuss dependency management.

aguschin:
> Is this what you were thinking?

Yes, exactly! The page you found has a great summary. There are a lot of thing W&B can log, I guess few of them fits good in the current functionality, like logging arrays and tensors of length 1. But tracking other types of data would require major extension of the current functionality, I guess. Did we have the similar discussion before somewhere?

>  you have a specific scenario in mind, and how would you like that data logged?

The specific scenario I first came up with is the following: to log tensors/arrays of length==1 with dvclive, I need to convert them to simple types. To avoid repeating this every time I've written a wrapper, which I'd like to avoid.

More advanced scenario would be to try plotting ROC AUC (TPR vs FPR), or Precision vs Recall, or Precision vs Threshold, or something llke this. To plot this I'd need to store these arrays. If I already use dvclive to store metrics for checkpoints, why not use it to store these arrays to later compare 1st epoch versus 10th epoch versus 100th epoch?

Essentially, user may want to store any kind of artefact over epochs to compare them between each other (few more examples - images with bounding boxes produced by NN, pictures produced by GAN). The question is the same: if I already use dvclive, why not store those artefacts with it?

[...]

PR #189, opened 2021-11-08
sklearn: Add basic logging methods.

daavoo:

---

Per https://github.com/iterative/dvclive/issues/5#issuecomment-953295791 added a few utilities:

- log_classification_report
- log_confusion_matrix
- log_precision_recall_curve
- log_roc_curve

---

Example:

<details>
<summary>foo.py</summary>

```python
from dvclive.sklearn import (
    log_classification_report,
    log_confusion_matrix,
    log_precision_recall_curve,
    log_roc_curve,
)

from sklearn.datasets import make_classification
from sklearn.ensemble import RandomForestClassifier
from sklearn.model_selection import train_test_split

X, y = make_classification(random_state=0)
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
clf = RandomForestClassifier(random_state=0)
clf.fit(X_train, y_train)

y_pred = clf.predict(X_test)
y_score = clf.predict_proba(X_test)[:, 1]

classification_report = log_classification_report(y_test, y_pred, "classification_report.json")
cm = log_confusion_matrix(y_test, y_pred, "confusion_matrix.png")
prc = log_precision_recall_curve(y_test, y_score, "prc.json")
roc = log_roc_curve(y_test, y_score, "roc.json")
```

</details>

```yaml
stages:
  foo:
    cmd: python foo.py
    metrics:
      - classification_report.json
    plots:
    - confusion_matrix.png
    - prc.json:
        cache: false
        x: recall
        y: precision
        title: Precision Recall Curve
    - roc.json:
        cache: false
        x: fpr
        y: tpr
        title: ROC curve
```

`dvc exp show`

| Experiment   | Created   | 0.precision   | 0.recall   | 0.f1-score   | 0.support   | 1.precision   | 1.recall   | 1.f1-score   | 1.support   | accuracy   | macro avg.precision   | macro avg.recall   | macro avg.f1-score   | macro avg.support   | weighted avg.precision   | weighted avg.recall   | weighted avg.f1-score   | weighted avg.support   |
|--------------|-----------|---------------|------------|--------------|-------------|---------------|------------|--------------|-------------|------------|-----------------------|--------------------|----------------------|---------------------|--------------------------|-----------------------|-------------------------|------------------------|
| workspace    | -         | 0.92308       | 0.92308    | 0.92308      | 13          | 0.91667       | 0.91667    | 0.91667      | 12          | 0.92       | 0.91987               | 0.91987            | 0.91987              | 25                  | 0.92                     | 0.92                  | 0.92                    | 25                     |
| master       | 09:10 PM  | 0.92308       | 0.92308    | 0.92308      | 13          | 0.91667       | 0.91667    | 0.91667      | 12          | 0.92       | 0.91987               | 0.91987            | 0.91987              | 25                  | 0.92                     | 0.92                  | 0.92                    | 25                     |

`dvc plots show`

Screenshot 2021-11-08 at 21-13-05 DVC Plot

### DIFF
--- a/dvclive/sklearn.py
+++ b/dvclive/sklearn.py
@@ -1,0 +1,83 @@
+import json
+import math
+
+import numpy as np
+from sklearn import metrics
+
+
+def _check_binary(y_true):
+    if len(np.unique(y_true)) > 2:
+        raise ValueError(
+            "Only binary classification is supported for roc_curve."
+        )
+
+
+def _dump(content, output_file):
+    with open(output_file, "w") as f:
+        json.dump(content, f, indent=4)
+
+
+def log_classification_report(y_true, y_pred, output_file, **kwargs):
+    kwargs["output_dict"] = True
+
+    classification_report = metrics.classification_report(
+        y_true=y_true, y_pred=y_pred, **kwargs
+    )
+
+    _dump(classification_report, output_file)
+
+    return classification_report
+
+
+def log_roc_curve(y_true, y_score, output_file, **kwargs):
+    _check_binary(y_true)
+
+    fpr, tpr, roc_thresholds = metrics.roc_curve(
+        y_true=y_true, y_score=y_score, **kwargs
+    )
+
+    roc = {
+        "roc": [
+            {"fpr": fp, "tpr": tp, "threshold": t}
+            for fp, tp, t in zip(fpr, tpr, roc_thresholds)
+        ]
+    }
+
+    _dump(roc, output_file)
+
+    return roc
+
+
+def log_precision_recall_curve(y_true, probas_pred, output_file, **kwargs):
+    _check_binary(y_true)
+
+    precision, recall, prc_thresholds = metrics.precision_recall_curve(
+        y_true=y_true, probas_pred=probas_pred, **kwargs
+    )
+
+    # ROC has a drop_intermediate arg that reduces the number of points.
+    # https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html#sklearn.metrics.roc_curve.
+    # PRC lacks this arg, so we manually reduce to 1000 as a rough estimate.
+    nth_point = math.ceil(len(prc_thresholds) / 1000)
+    prc_points = list(zip(precision, recall, prc_thresholds))[::nth_point]
+
+    prc = {
+        "prc": [
+            {"precision": p, "recall": r, "threshold": t}
+            for p, r, t in prc_points
+        ]
+    }
+    _dump(prc, output_file)
+
+    return prc
+
+
+def log_confusion_matrix(y_true, y_pred, output_file, **kwargs):
+    confusion_matrix_disp = metrics.ConfusionMatrixDisplay.from_predictions(
+        y_true=y_true, y_pred=y_pred, **kwargs
+    )
+
+    confusion_matrix_disp.plot()
+    confusion_matrix_disp.figure_.savefig(output_file)
+
+    return confusion_matrix_disp
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -182,7 +182,7 @@ def test_require_step_update(tmp_dir, metric):
         dvclive.log(metric, 2.0)
 
 
-def test_custom_steps(tmp_dir, mocker):
+def test_custom_steps(tmp_dir):
     dvclive = Live("logs")
 
     steps = [0, 62, 1000]
--- a/tests/test_sklearn.py
+++ b/tests/test_sklearn.py
@@ -1,0 +1,91 @@
+import json
+from pathlib import Path
+
+import pytest
+
+from dvclive.sklearn import (
+    log_classification_report,
+    log_confusion_matrix,
+    log_precision_recall_curve,
+    log_roc_curve,
+)
+
+# pylint: disable=redefined-outer-name, unused-argument
+
+
+@pytest.fixture
+def y_true_y_pred_y_score():
+    from sklearn.datasets import make_classification
+    from sklearn.ensemble import RandomForestClassifier
+    from sklearn.model_selection import train_test_split
+
+    X, y = make_classification(random_state=0)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
+    clf = RandomForestClassifier(random_state=0)
+    clf.fit(X_train, y_train)
+
+    y_pred = clf.predict(X_test)
+    y_score = clf.predict_proba(X_test)[:, 1]
+
+    return y_test, y_pred, y_score
+
+
+def test_log_classification_report(tmp_dir, y_true_y_pred_y_score, mocker):
+    from dvclive.sklearn import metrics
+
+    y_true, y_pred, _ = y_true_y_pred_y_score
+
+    spy = mocker.spy(metrics, "classification_report")
+
+    classification_report = log_classification_report(
+        y_true=y_true, y_pred=y_pred, output_file="classification_report.json"
+    )
+
+    spy.assert_called_once_with(y_true, y_pred, output_dict=True)
+
+    assert (
+        json.dumps(classification_report, indent=4)
+        == Path("classification_report.json").read_text()
+    )
+
+
+def test_log_roc_curve(tmp_dir, y_true_y_pred_y_score, mocker):
+    from dvclive.sklearn import metrics
+
+    y_true, _, y_score = y_true_y_pred_y_score
+
+    spy = mocker.spy(metrics, "roc_curve")
+
+    roc = log_roc_curve(y_true=y_true, y_score=y_score, output_file="roc.json")
+
+    spy.assert_called_once_with(y_true, y_score)
+
+    assert json.dumps(roc, indent=4) == Path("roc.json").read_text()
+
+
+def test_log_prc_curve(tmp_dir, y_true_y_pred_y_score, mocker):
+    from dvclive.sklearn import metrics
+
+    y_true, _, y_score = y_true_y_pred_y_score
+
+    spy = mocker.spy(metrics, "precision_recall_curve")
+
+    prc = log_precision_recall_curve(
+        y_true=y_true, probas_pred=y_score, output_file="prc.json"
+    )
+
+    spy.assert_called_once_with(y_true, y_score)
+
+    assert json.dumps(prc, indent=4) == Path("prc.json").read_text()
+
+
+def test_log_confusion_matrix(tmp_dir, y_true_y_pred_y_score, mocker):
+    y_true, y_pred, _ = y_true_y_pred_y_score
+
+    mocker.patch("sklearn.metrics.ConfusionMatrixDisplay.from_predictions")
+
+    mock = log_confusion_matrix(
+        y_true=y_true, y_pred=y_pred, output_file="confusion_matrix.png"
+    )
+
+    mock.figure_.savefig.assert_called_with("confusion_matrix.png")


pared:
The integration looks alright. One question that I have is whether we want to promote logging confusion matrix as image file. While we do support images on DVC side, I am not sure we should promote that. 
Alternatives:
- just dump matrix data, but that will only work with dvc/visualization lib that we might extract from DVC
- dump the confusion matrix as vega spec, start supporting that on dvc side - that is probably worse than previous since we are thinking about supporting other libraries

pared:
Well, maybe it's another reason to reconsider iterative/dvc#6944 - maybe we should discuss somehow providing config for each image which would define how to plot it?

dvclive/sklearn.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def log_precision_recall_curve(y_true, probas_pred, output_file, **kwargs):
+    _check_binary(y_true)
+
+    precision, recall, prc_thresholds = metrics.precision_recall_curve(
+        y_true=y_true, probas_pred=probas_pred, **kwargs
+    )
+
+    # ROC has a drop_intermediate arg that reduces the number of points.
```
dberenbaum, line 58: This was a really hacky solution I introduced for the example repo. Not sure it makes sense to include it as is.

dberenbaum:
Agree with @pared about the confusion matrix and plots in general. Seems better to save the data and work on making it easier to specify plots properties.

For the other plots, it seems awkward to have to call them in dvclive and configure them in dvc. 

Also, I'm not sure how much value there is in these functions that lightly wrap existing sklearn functions. Seems more useful to do one or both of:
* `log_data()` to log any of these results. We can work on making it easy to show those results as plots, and it gives users flexibility to keep the raw results to process themselves.
* Some sort of auto-logging at a higher level for sklearn (similar to https://www.mlflow.org/docs/latest/python_api/mlflow.sklearn.html#mlflow.sklearn.autolog) so users have an easy default logger.

What do you think?

daavoo:
Revisiting this.

> Also, I'm not sure how much value there is in these functions that lightly wrap existing sklearn functions. Seems more useful to do one or both of:
> 
>     * `log_data()` to log any of these results. We can work on making it easy to show those results as plots, and it gives users flexibility to keep the raw results to process themselves.

I don't fully get `log_data` (perhaps we discussed offline and I forgot 😓 ) . Could you describe would it work (on a high-level)?

>     * Some sort of auto-logging at a higher level for sklearn (similar to https://www.mlflow.org/docs/latest/python_api/mlflow.sklearn.html#mlflow.sklearn.autolog) so users have an easy default logger.
> 
> 
> What do you think?

About the value of these light wraps, I don't really know. 

My view is that there are many stages between this light wraps and automagic logging "a la mlflow". 

I would say that a good intermediate would be `log_classifier` ("a la wandb" https://docs.wandb.ai/guides/integrations/scikit).

However, it seems that it's up to user taste and there is no real need to don't support any stage. Automagic methods are going to be using something similar to (if not exactly) these light wraps so, why not expose all different levels of granularity?

dberenbaum:
I'm still unsure about this. My hesitations are:

1. The included methods are a pretty arbitrary selection. How do we decide what to include, and do we want to potentially support wrappers for every kind of sklearn metric/plot?
2. There's no obvious way to include this (or any sklearn integration) as a sort of integrated callback like other frameworks that automatically logs info as the model trains.
3. If we eventually want something like `log_classifier`, the closest thing AFAIK is https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report. This wouldn't even utilize these methods I don't think. 

Another option is to add support solely via docs (once https://github.com/iterative/dvclive/pull/203 is merged), similar to https://dvc.org/doc/dvclive/ml-frameworks/tensorflow.

How much value do you think the current PR adds, and how much added value do you see in further integration?

daavoo:
>     1. The included methods are a pretty arbitrary selection. How do we decide what to include, and do we want to potentially support wrappers for every kind of sklearn metric/plot?

I would not say it's arbitrary. It's not complete, for sure, but includes 3 out of the 6 visualizations supported in scikit-learn (https://scikit-learn.org/stable/visualizations.html). 

I would support any scikit-learn visualization that can be seamlessly integrated with `dvc plots` usage. Under that condition, the only missing from the list would be the `det_curve` (could be added in the P.R, I just forgot)

It's a similar selection to the integrations in https://docs.wandb.ai/guides/integrations/scikit . The difference is that some of those plots types are not directly integrable with `dvc plots` and thus I don't see much sense in adding support.

>     2. There's no obvious way to include this (or any sklearn integration) as a sort of integrated callback like other frameworks that automatically logs info as the model trains.

Indeed, but it's a common limitation for any logger that doesn't perform the "magic" patching.  This is addressable on docs, similar to the existing barebones PyTorch and TensorFlow guides.
  
>     3. If we eventually want something like `log_classifier`, the closest thing AFAIK is https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html#sklearn.metrics.classification_report. This wouldn't even utilize these methods I don't think.

With `log_classifier/log_regressor` I was thinking something more in the lines of https://docs.wandb.ai/guides/integrations/scikit#or-visualize-all-plots-at-once which would call internally this `log_{X}` functions. 

Potentially, after #203,  this could be extended to include `live.log` for multiple classification/regression metrics but I don't see how these are mutually exclusive.

> Another option is to add support solely via docs (once #203 is merged), similar to https://dvc.org/doc/dvclive/ml-frameworks/tensorflow.
> 
> How much value do you think the current PR adds, and how much added value do you see in further integration?

I see immediate value to be used in https://github.com/iterative/example-get-started to reduce code and promote `dvclive`.
I think it can prevent users from having to be aware of the format `dvc plots` expects.
 
Added value in further integration should move towards "magic" patching and/or `log_classifier`. These functions can still have value there both for internal usage but also for users preferring more fine-grained control on what to log.

---

Given that the inputs are arrays of `y_true`/`y_pred`, my main doubt is whether we should explicitly save this under `sklearn` or consider this a new `plots` module that has `sklearn` as an optional dependency.

dberenbaum:
> I would not say it's arbitrary. It's not complete, for sure, but includes 3 out of the 6 visualizations supported in scikit-learn (https://scikit-learn.org/stable/visualizations.html).
> 
> I would support any scikit-learn visualization that can be seamlessly integrated with `dvc plots` usage. Under that condition, the only missing from the list would be the `det_curve` (could be added in the P.R, I just forgot)
> 
> It's a similar selection to the integrations in https://docs.wandb.ai/guides/integrations/scikit . The difference is that some of those plots types are not directly integrable with `dvc plots` and thus I don't see much sense in adding support.

Thanks for the explanation. That makes sense.

Why can't calibration curves and partial dependence plots also be supported in DVC? They are all linear unless I'm missing something.

> With `log_classifier/log_regressor` I was thinking something more in the lines of https://docs.wandb.ai/guides/integrations/scikit#or-visualize-all-plots-at-once which would call internally this `log_{X}` functions.

I think this is where we had different ideas for sklearn integration. I thought we would first need to capture the scalar metrics before worrying about plots. I guess you are thinking that the plots are harder for the user to log, and they can easily log scalar metrics already? Still, it would be nice to automatically capture all of the relevant scalar metrics (in addition to plots) if there is some higher-level integration.

> Given that the inputs are arrays of `y_true`/`y_pred`, my main doubt is whether we should explicitly save this under `sklearn` or consider this a new `plots` module that has `sklearn` as an optional dependency.

Are you suggesting that these could be generic functions rather than specific to `sklearn`? I have previously kept a mini-library of classification metrics/plots to have a consistent, lightweight way to evaluate models across ML frameworks. Maybe that's more useful here than any `sklearn`-specific integration. It's definitely a different direction for dvclive, but it might be worth considering.

daavoo:
> Why can't calibration curves and partial dependence plots also be supported in DVC? They are all linear unless I'm missing something.

You are right, calibration curves can be supported. And also partial dependence, but only in "average" mode.

> Still, it would be nice to automatically capture all of the relevant scalar metrics (in addition to plots) if there is some higher-level integration.

I agree, after #203 , a higher level `log_classifier` would also log scalar metrics.

> Are you suggesting that these could be generic functions rather than specific to `sklearn`? I have previously kept a mini-library of classification metrics/plots to have a consistent, lightweight way to evaluate models across ML frameworks. Maybe that's more useful here than any `sklearn`-specific integration. It's definitely a different direction for dvclive, but it might be worth considering.

We can start documenting as part of sklearn integration but I can see the plots being useful for basically any other framework that supports some classification task (so, pretty much, all).

dberenbaum:
Should these be logged inside the dvclive dir?

pared (approved):

dvclive/plots.py
```diff
@@ -0,0 +1,102 @@
...
+            {"precision": p, "recall": r, "threshold": t}
+            for p, r, t in zip(precision, recall, prc_thresholds)
+        ]
+    }
+    _dump(prc, output_file)
+
+    return prc
+
+
+def log_confusion_matrix(y_true, y_pred, output_file):
```
dberenbaum, line 56: Why no `**kwargs` here?

dberenbaum:
Ultimately, I think the workflow for this should depend on #82 and #203. Right now, it's impossible to log from within a `Live()` instance, which makes it a bit awkward to use. I would expect a workflow like:

```
live = dvclive.Live()
live.log_calibration(y_test, y_score, "calibration.json")
```

Do you think it's worth merging as is and then modifying it after? I think it will require breaking changes to address those issues.

daavoo:
> Ultimately, I think the workflow for this should depend on #82 and #203. Right now, it's impossible to log from within a Live() instance, which makes it a bit awkward to use.

> Do you think it's worth merging as is and then modifying it after? I think it will require breaking changes to address those issues.

Do you mention #82 because you consider that these plots should be loggable at every step?

Originally, I was considering a non-step workflow for these plots. 

Agree that #203 and this should be documented and released together. I was already considering exposing the functions as `Live` methods because of https://github.com/iterative/dvclive/pull/189#issuecomment-1010139910 , given that a non-step workflow would include #203, I guess it makes even more sense to make the functions part of `Live`.

dvclive/plots.py
daavoo: This is the only case where we are not calling `sklearn` internal function but just reformating the data into DVC format . There were kwargs in previous version because we were calling `sklearn.metrics.confusion_matrix` and saving the plot as image.

dberenbaum:
> Do you mention #82 because you consider that these plots should be loggable at every step?

I mention it because this PR seems like a specific application of #82. Similar to #166, it seems that we will need to decide on the canonical workflow and directory structure for this type of data in both non-step and step-based workflows. So, we can try to decide that now, or we can merge knowing we will need to break it later.

daavoo (commented):
Moved all the logic to `Live.log` and made `plots` a new "data type". 

So the idea is that `Live.log` can log any of the 3 supported data types: scalars, images, and plots.

- Scalar

```python
live.log("accuracy", 0.9)
```
- Image

```python
img = np.ones((500, 500, 3), np.uint8)
live.log("image.png", img)
```

- Plot

```python
live.log("roc_curve.json", (y_true, y_score, "roc"))
live.log("cm.json", (y_true, y_pred, "confusion_matrix"))
```

The internals make sense for me and the no step / step logic is resolved equally for images and plots.

However, I'm not sure how intuitive the API for plots is. 

It could be a matter of documentation or it could be better to have explicitly separated methods for each data type `log_metric`, `log_image`, `log_plot`. 

I kind of prefer separate methods.

pared:
I think that separate methods make sense too, as the distinction in use will be seen at first sight. I am not sure we should rename `log` to `log_metric`.

dberenbaum:
Nice! I agree with @pared that it seems best to have separate methods but leave `log` for metrics.

dberenbaum:
> Do you mention #82 because you consider that these plots should be loggable at every step?

Returning to this, yes, I think it should be possible to see how the ROC curve changes at each step, for example. Right now, saving the data works, but it's almost impossible to configure `dvc.yaml` to show the plots, and in default mode they make it hard to even see the regular metrics history plots.

Screen Shot 2022-01-28 at 10 38 05 AM

daavoo:
> Returning to this, yes, I think it should be possible to see how the ROC curve changes at each step, for example. Right now, saving the data works, but it's almost impossible to configure dvc.yaml to show the plots, and in default mode they make it hard to even see the regular metrics history plots.

The initial scope for this kind of plot was for non-step. 

Properly supporting multi-step curves would require a new template would be required on DVC side using `facet` for comparing revision.

dberenbaum:
> The initial scope for this kind of plot was for non-step.
> 
> Properly supporting multi-step curves would require a new template would be required on DVC side using `facet` for comparing revision.

Okay, let's discuss separately in #82 or elsewhere the per-step scenario. Why support logging at each step then in this PR? Should an error be thrown if using `log_plot` per step?

daavoo:
> Why support logging at each step then in this PR? Should an error be thrown if using `log_plot` per step?

No strong reason, it just reuses the logic used for images. Not sure about raising exception, there is nothing wrong with how the data is being stored apart from lack of support by default `dvc plots` commands (i.e. each individual plot file is usable on its own if passed as individual target).

In my mind, it's a matter of extending `dvc plots` to support per-step plots (will open issue to discuss) and find out if a  different format is needed or try to support the existing folder-based format.

dberenbaum:
Sorry for the confusion over step/no-step scenarios. Limiting to no-step makes sense.

dberenbaum (approved):

dvclive/live.py
```diff
@@ -134,28 +144,51 @@ def next_step(self):
...
+            self._scalars[name] = data
+
         data.dump(val, self._step)
 
         if self._summary:
             self.make_summary()
 
+    def log_image(self, name: str, val):
+        if not Image.could_log(val):
+            raise InvalidDataTypeError(name, type(val))
```
jorgeorpinel, line 163: Maybe `val` should be renamed. `val` made sense for scalars (`log()`) but `image_data`** or something more descriptive would be better here IMO.

dvclive/live.py
```diff
@@ -134,28 +144,51 @@ def next_step(self):
...
+
+        if name in self._images:
+            data = self._images[name]
+        else:
+            data = Image(name, self.dir)
+            self._images[name] = data
```
jorgeorpinel, line 169: As for `name`, should it be `filename` or `out` instead? Again for clarity

dvclive/live.py
```diff
@@ -134,28 +144,51 @@ def next_step(self):
...
+
+        data.dump(val, self._step)
+
+    def log_plot(self, name, labels, predictions, **kwargs):
+        val = (labels, predictions)
```
jorgeorpinel, line 174: Same here. `plot_data`?

dvclive/live.py
jorgeorpinel, line 169: As for `name`, should it be `filename` or `out` instead? Again for clarity

dvclive/live.py
```diff
@@ -134,28 +144,51 @@ def next_step(self):
...
+
+        if name in self._plots:
+            data = self._plots[name]
+        elif name in PLOTS and PLOTS[name].could_log(val):
+            data = PLOTS[name](name, self.dir)
+            self._plots[name] = data
+        else:
+            raise InvalidPlotTypeError(name)
```
jorgeorpinel, line 182: Same. `filename`?

jorgeorpinel: Or I guess `plot_type` would be most accurate here.

PR #197, opened 2021-12-06
update metric summary data of arbitrary depth

### LINKED ISSUE #196: bad summary for 2 level deep nested metric
The `dict.update` method used here: https://github.com/iterative/dvclive/blob/master/dvclive/live.py#L146 fails to correctly aggregate multiple metrics with 2 or more nested level.

I've quickfixed the code with a [stackoverflow answer](https://stackoverflow.com/a/3233356) to update an arbitrarily deep nested dict.
I'll make pull request on monday.

kwon-young:
The default `dict.update` method only update the 1st level key of the
dict. When using metric with with 2 or more nested levels, only the
last update of each nested key will be kept.
For example:

>> summary_data = {'valid': {'accuracy': 0.6}}
>> summary = {'valid': {'precision': 0.5}}
>> summary_data.update(summary)
>>> summary_data
{'valid': {'precision': 0.5}}

The function `update_nesteddict` will correctly update an arbitrarily
nested dict, so that:

>> summary_data = {'valid': {'accuracy': 0.6}}
>> summary = {'valid': {'precision': 0.5}}
>> summary_data = update_nesteddict(summary)
>>> summary_data
{'valid': {'precision': 0.5, 'accuracy': 0.6}}

Fixes #196 

* [X] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -3,7 +3,7 @@ import logging
 import os
 import shutil
 from collections import OrderedDict
-from collections.abs import Mapping
+from collections.abc import Mapping
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -3,25 +3,15 @@ import logging
 import os
 import shutil
 from collections import OrderedDict
-from collections.abc import Mapping
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 from .data import DATA_TYPES
 from .dvc import make_checkpoint, make_html
+from .utils import nested_update
 from .error import ConfigMismatchError, InvalidDataTypeError
 
 logger = logging.getLogger(__name__)
-
-
-def update_nesteddict(d, u):
-    """Update values of a nested dictionnary of varying depth"""
-    for k, v in u.items():
-        if isinstance(v, Mapping):
-            d[k] = update_nesteddict(d.get(k, {}), v)
-        else:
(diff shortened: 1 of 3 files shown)


dvclive/live.py
```diff
@@ -3,6 +3,7 @@
 import os
 import shutil
 from collections import OrderedDict
+from collections.abs import Mapping
```
pared, line 6:
```suggestion
from collections.abc import Mapping
```

pared (changes requested):
Thank you very much for the report and this change!

We would need a test for that.
We either have to write a test specifically for this functionality or update
`test_main.py::test_nested_logging`,
log another nested value and assert that the resulting `.json` has both nested values.

@kwon-young would you like to write a test, or should we take over?

daavoo (changes requested):
Thanks for the contribution @kwon-young !

In addition to @pared comments, could you move the `update_nesteddict` function from `dvclive/live.py` to `dvclive/utils.py`. And, for consistency, perhaps rename it to `nested_update` (as we already have one utility named `nested_set`). 

Thanks!

kwon-young:
Let me know if there is something else left to do to merge this.

daavoo (approved):
Thanks for the contribution and addressing all comments 🚀

Issue #182, opened 2021-10-27
`log`: Make plot generation optional


daavoo:
I think that one of the easiest ways to start exploring how dvclive could provide value in ML Scenarios where there are no training steps (i.e. Most `sklearn` algorithms) would be to make the generation of the `.tsv` plots optional.

Now that both `summary` and `plots` generation happen at `log` level, it should be not complicated to expose an argument equivalent to `summary` (i.e. `plots`?) for toggling the generation of plots.

Even though it won't add much value, at least it would be useful for logging arbitrary metrics in those "stepless" scenarios without creating a single line `.tsv`.

```python
live.log("avg_prec", average_precision_score(labels, predictions))
```

daavoo:
> What about making `step` optional rather than adding a new argument? If `step` is `None`, do not log any step-based info.

That's a very good idea. 

I need to do some research on DVC integration (not sure how it would handle the non-existence of plots).

PR #203, opened 2021-12-28
live: Implicit no_step

daavoo:
Disable `step` related functionality. Useful for traditional ml scenarios.

Closes https://github.com/iterative/dvclive/issues/182

`dvc.org`: https://github.com/iterative/dvc.org/pull/3147


tests/conftest.py
```diff
@@ -5,9 +5,6 @@
 
 @pytest.fixture
 def tmp_dir(tmp_path, monkeypatch):
-    import dvclive
-
-    dvclive._metric_logger = None  # pylint: disable=protected-access
```
pared, line 10: +1

pared (approved):

pared:
@daavoo code-check fails

dvclive/error.py
```diff
@@ -33,3 +33,8 @@ def __init__(self, name, step):
         super().__init__(
             f"Data '{name}' has already being logged whith step '{step}'"
         )
+
+
+class NoStepError(DvcLiveError):
+    def __init__(self, func):
+        super().__init__(f"'{func}' can't be used if 'step=False'")
```
dberenbaum, line 40:
```suggestion
        super().__init__(f"'{func}' can't be used if 'no_step=True'")
```

dberenbaum (commented):
Seems like this isn't compatible right now with https://github.com/iterative/dvclive/pull/166?

dberenbaum:
More thoughts:

* How should `resume = True` and `summary = False` interact with `no_step = True`?
* Out of curiosity, would it be difficult to make this workflow implicit (assume a no-step scenario until `next_step` is called)? I assume this would be more complicated to implement, and since everything else in dvclive is very explicit now, I don't see it as an issue, but it would be good to understand since the implicit workflow is simpler.

daavoo (commented):
> Out of curiosity, would it be difficult to make this workflow implicit (assume a no-step scenario until next_step is called)? I assume this would be more complicated to implement, and since everything else in dvclive is very explicit now, I don't see it as an issue, but it would be good to understand since the implicit workflow is simpler.

Implemented this implicit `no_step`. No plots are generated until `step` is updated (either `next_step` or `set_step`.

> Seems like this isn't compatible right now with #166?

Fixed. First `log` will save image to `{dvclive.dir}/{image_name}`, if next_step is called, that first logged  image is renamed to `{dvclive.dir}/0/{image_name}` and subsequent calls will save in `{dvclive.dir}/{step}/{image_name}`.

> How should resume = True and summary = False interact with no_step = True?

In the current implicit implementation, `resume` won't have any effect unless `step` is updated. `summary=False` will result in "log" calls doing nothing until `step` is updated the first time.

dberenbaum:
So what do you think about this implicit workflow? Any concerns of side effects like #156? Do you think it will be confusing to move the output? Or hard to handle when there are more output types?

If you're comfortable with the implicit workflow, it looks good to me.

dberenbaum (approved):

daavoo:
> So what do you think about this implicit workflow? 

I think it makes sense from the user's perspective.  However, I'm not that comfortable with the internal changes required though. Right now it's a little messy, but there might be a simpler way of doing that I haven't thought.

> Any concerns of side effects like #156?

All I can think at the moment is that we should remove the `summary` argument and make it mandatory.

> Do you think it will be confusing to move the output? Or hard to handle when there are more output types?

I think that moving the image output is a little awkward. Would it be confusing to have a different output pattern for the first step? (i.e. first log would be `{live.dir}/images/{filename}` and subsequent go to {live.dir}/images/{live.step}/{filename}`)

In https://github.com/iterative/dvclive/pull/166 there were not really strong opinions for picking `{live.dir}/images/{live.step}/{filename}` over other alternatives.

Given that image support is yet to be documented, perhaps there is a better pattern that doesn't require moving

dberenbaum:
> However, I'm not that comfortable with the internal changes required.

Anything besides the points made in the comment above?

> Given that image support is yet to be documented, perhaps there is a better pattern that doesn't require moving

Any ideas?

As I mentioned before, I think it was fine to have an explicit workflow since that's what we have opted for elsewhere, so I have no objections if you prefer to revert to that workflow.

PR #204, opened 2021-12-30
Add `patch_tensorboard`

daavoo:

---

Add `patch_tensorboard` function.

It can override or extend existing tensorboard logging calls by redirecting those to `Live.log`/`Live.log_image` calls.

This allows to patch both raw tensorboard calls but also existing callbacks using tensorboard:

```python

```

```python
from dvclive.tensorboard import patch_tensorboard
 
patch_tensorboard()

. . .
model.fit(
    x,
    y,
    epochs=2,
    callbacks=[tf.keras.callbacks.TensorBoard()],
)
```

`patch_tensorboard(override=False)` would generate DVCLive outputs while preserving the original behavior.

### DIFF
--- a/dvclive/tensorboard.py
+++ b/dvclive/tensorboard.py
@@ -1,0 +1,53 @@
+import gorilla
+import tensorflow as tf
+
+from dvclive import Live
+
+# pylint: disable=unused-argument, no-member
+
+
+def patch_tensorboard(override: bool = True, **kwargs):
+    live = Live(**kwargs)
+    settings = gorilla.Settings(allow_hit=True, store_hit=True)
+
+    original_scalar = gorilla.Patch(
+        tf.summary, "original_scalar", tf.summary.scalar, settings=settings
+    )
+    gorilla.apply(original_scalar)
+
+    def log_scalar(name, data, step=None, description=None):
+        if step is not None:
+            if step != live.get_step():
+                live.set_step(step)
+        live.log(name, data)
+        if not override:
+            tf.summary.original_scalar(name, data, step=None, description=None)
+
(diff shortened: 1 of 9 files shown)


dvclive/tensorboard.py
```diff
@@ -0,0 +1,53 @@
+import gorilla
```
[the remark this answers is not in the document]
dberenbaum: It looks like mlflow uses it but also has their own modified versions of it in https://github.com/mlflow/mlflow/blob/master/mlflow/utils/gorilla.py? Maybe one of the engineers can weigh in.

tests/test_keras.py
```diff
@@ -179,3 +179,37 @@ def test_keras_None_model_file_skip_load(
...
+    assert not scalar.call_count
+    assert not image.call_count
+
+    assert os.path.exists("logs")
+    logs, _ = read_logs("logs/scalars")
+
+    assert "epoch_accuracy" in logs
+    assert len(logs["epoch_accuracy"]) == 2
+
+    for patch in patches:
```
pared, line 214: So in case of using `dvclive` with `tensorboard` we would need to revert the patches after the execution in order to not use modified `tensorboard` later in our project? 

Shouldn't `patch_tensorboard` be a context manager that would revert the patch after exiting the context?

daavoo: I need to test it but I think that the patch doesn't persist across python executions.

pared: What about the same process? I think safest way would be to let the user control when the `dvclive` is patching the tensorboard.

dberenbaum:
Discussed with @daavoo that we will hold off on merging this.

PR #210, opened 2022-01-27
live: Raise exception if resume and not summary

daavoo:
Per https://github.com/iterative/dvclive/pull/209#discussion_r789986730

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -23,7 +23,8 @@ class Live:
         resume: bool = False,
         summary: bool = True,
     ):
-
+        if resume and not summary:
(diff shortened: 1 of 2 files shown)


dberenbaum:
At what point do we make the summary required? It seems it's taking substantial effort to keep working around it being optional.

daavoo:
I will add a DeprecationWarning message in the upcoming release and implement the actual removal for the release after that. Is this anything blocking this P.R. @dberenbaum ?

dberenbaum:
Nope, looks good! Just thinking it's a shame to keep adding logic that needs to be remove in the near future.

dberenbaum (approved):

Issue #91, opened 2021-06-15
integrations: CML


daavoo:
From a high level user perspective, I think it would be interesting to have some kind of integration between `dvclive` and `CML`.

Just like there is already a custom `dvclive` behavior when used along with `DVC` (i.e. generating `checkpoints`) maybe a similar approach could be used when `dvclive` is also used along with `CML`.

I was thinking on a use case like `dvclive` taking care of using `cml-send-comment` inside `next_step` if `CML` is detected. 

Conceptually i think that it would be something similar to the existing `make_checkpoint`:

https://github.com/iterative/dvclive/blob/master/dvclive/metrics.py#L105  

This might be just an specific use case of https://github.com/iterative/dvclive/issues/90

dberenbaum:
This seems like a general need for integration between cml and dvc checkpoints. Do you see any functionality as being specific to dvclive?

daavoo:
> This seems like a general need for integration between cml and dvc checkpoints. Do you see any functionality as being specific to dvclive?

Given the discussion so far in https://github.com/iterative/dvc/issues/6182, it looks like the `exp push` integration won't require any changes from the `dvclive` side. 

I probably didn't express it right but I was also taking about an additional integration, which will be something like using `cml-send-comment` in the context of a long running experiment (i.e. deep learning) as a way to periodically report the status of the training. 

From an user perspective what I would want is, by using `dvclive` in an environment where CML is installed, getting a P.R. comment reporting the `dvclive` metrics *summary* **each** time an epoch is finished (which would translate to each time `dvclive.next_step` is called). 

This is a feature (sending the metrics to somewhere) that I have been repeatedly implementing through workarounds when using other *ML Loggers* and I though that ot could be nice to have it working out of the box with `dvclive<>CML`.

The feature probably doesn't make too much sense in the context of training a lot of short experiments and it's probably not a deal breaker for attracting new users.

dberenbaum:
> The feature probably doesn't make too much sense in the context of training a lot of short experiments and it's probably not a deal breaker for attracting new users.

Yes, this could be really useful, and absolutely dvclive should address issues for long-running experiments! https://github.com/iterative/dvc/issues/6182 will at least make it possible to pull the latest checkpoint and see the associated metrics and plots, but we could make the experience smoother and better integrate cml.

Seems like this has 2 requirements:

1. Generate some default report to send.
[...]

Issue #90, opened 2021-06-15
logger: Add notifier to `next_step`?


daavoo:
Depending on the type of model to be trained, the time in between calls to `next_step`  may vary significantly. In common deep learning scenarios, i.e. the [keras callback](https://github.com/iterative/dvclive/blob/master/dvclive/keras.py), `next_step`  is being called at the end of an epoch which could result in long times (maybe hours) in between calls.

It could be useful to have built-in support for optionally sending a notification each time `next_step` is being called. 

Without changing `dvclive`, the user could just call a custom library (i.e. https://github.com/liiight/notifiers) after next_step:

```python
class MetricsCallback(Callback):
    def on_epoch_end(self, epoch: int, logs: dict = None):
        logs = logs or {}
        for metric, value in logs.items():
            dvclive.log(metric, value)
        dvclive.next_step()
        notify('pushover', user='foo', token='bar', message=f'epoch: {epoch}')
```

But having the notification step built inside `MetricLogger` would have some benefits like access to internals (i.e. `_metrics`) and configuration options in addition to hiding complexity to the end user.

However, I'm not sure if it is worth to implement this feature inside `dvclive` or if it would be better to keep `dvclive` as lightweight as possible.

dmpetrov:
@daavoo I'm trying to understand the motivation behind this? 😄  
could you please elaborate on this? o you want to update the files more often? What `notify()` means?

daavoo:
> @daavoo I'm trying to understand the motivation behind this? smile
> could you please elaborate on this? o you want to update the files more often? What `notify()` means?

Sorry about the lack of clarity. 

The motivation comes from working with deep learning models that take a lot of time to train (i.e. hours or days). When working under that circumstances we always ended up writing some sort of "notification" code to complement or integrate into the ml logger. 

This notification code takes care of sending a message to some platform (i.e. e-mail, slack / discord / telegram channel, etc) containing information like the number of finished epoch (a.k.a `step` in `dvclive`) and associated metrics. We also used it to inform when exceptions occurred during the training loop.  

`notify()` is usually a function that sends useful information as a message to an app. 

> 
> EDIT: do you have any references in other ml logger frameworks to this functionality?

I think that in other ml loggers we usually have an associated UI with a view that is automatically being updated as the plots/information are being logged (Related with this Studio issue: https://github.com/iterative/studio-support/issues/13)

In addition to that, some ml loggers also provide "notification" utilities:

[...]

PR #215, opened 2022-02-01
live: Add `report` option.

daavoo:
If True and CML is available, use CML to build and publish
a markdown report with metrics and plots on each step.

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -183,9 +183,7 @@ class Live:
         with TemporaryDirectory() as tmpdir:
             file = Path(tmpdir) / "report.md"
             file.write_text(report)
-            subprocess.run(
-                ["cml", "send-comment", str(file)], check=True
-            )
+            subprocess.run(["cml", "send-comment", str(file)], check=True)
         return True
 
     def read_step(self):
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -184,7 +184,7 @@ class Live:
             file = Path(tmpdir) / "report.md"
             file.write_text(report)
             subprocess.run(
-                ["cml", "send-comment", "--update", str(file)], check=True
+                ["cml", "send-comment", str(file)], check=True
             )
         return True
 
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -2,14 +2,16 @@ import json
 import logging
 import os
 import shutil
+import subprocess
 from collections import OrderedDict
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Any, Dict, Optional, Union
 
 from .data import DATA_TYPES
 from .dvc import make_checkpoint, make_html
 from .error import ConfigMismatchError, InvalidDataTypeError
-from .utils import nested_update
+from .utils import is_cml_available, nested_update
 
 logger = logging.getLogger(__name__)
 
@@ -18,7 +20,11 @@ class Live:
     DEFAULT_DIR = "dvclive"
 
     def __init__(
-        self, path: Optional[str] = None, resume: bool = False, **kwargs
+        self,
+        path: Optional[str] = None,
+        resume: bool = False,
+        report: bool = True,
+        **kwargs,
     ):
         if "summary" in kwargs:
             logger.warning(
@@ -31,6 +37,7 @@ class Live:
             raise ValueError("`resume` can't be used without `summary`")
         self._path: Optional[str] = path
         self._resume: bool = resume
+        self._report: bool = report
         self._summary: bool = summary
         self._html: bool = True
         self._checkpoint: bool = False
@@ -133,6 +140,9 @@ class Live:
         if self._checkpoint:
             make_checkpoint()
 
+        if self._report and is_cml_available():
+            self.make_report()
+
         self._step = step
 
     def next_step(self):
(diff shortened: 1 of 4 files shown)


dvclive/report.py
```diff
@@ -0,0 +1,56 @@
+import os
+import subprocess
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from dvclive.utils import _parse_tsv, is_cml_available
+
+
+def build_report(plots_folder):
```
dberenbaum, line 9: It seems more generally useful for `dvc plots show` (or wherever we extract the plots rendering library) to have a markdown format option. This would also make it easier in general to generate reports in CML instead of needing to use tools like vl2png. Not sure how we can also include metrics or if that should stay separate 🤔 . cc @pared

pared: I think that ideally we should have an option to define a report both for both DVC and `dvclive` - including metrics plots and maybe even params - though that would require some research - including metrics and params as a markdown table seems doable, I don't know about vega plots - how to handle image generation, which currently leverages HTML and `<script>`.

dberenbaum: Hmm, looks like markdown plots generation might require altair and additional packages to do in vega-lite: https://altair-viz.github.io/user_guide/saving_charts.html#png-svg-and-pdf-format. Alternatives would be to have other rendering options like matplotlib (what @daavoo uses here) or plotly (which looks to be the most lightweight option).

Re: including metrics and maybe params, I'm not sure whether this belongs in dvc, dvclive, or both. I think the current table makes sense for dvclive since it only knows about those metrics, and dvc doesn't know about steps. I could see value in including metrics, params, and plots altogether in dvc outputs, but this might have to be separate issue.

daavoo: > It seems more generally useful for dvc plots show (or wherever we extract the plots rendering library) to have a markdown format option.

I see 2 different scopes. 

On the one hand, this P.R. focuses on monitoring a running stage only covering the DVCLive scope (current workspace and metrics and plots logged by `Live`). 

On the other, it seems to me that the discussion is more on the lines of extending/simplifying the existing [CML/DVC integration](https://cml.dev/doc/cml-with-dvc) which happens after a stage has been finished.

IMO, having `--md` options in `dvc metrics`, `dvc params` and `dvc exp` commands is better than trying to join all together in a single output. It allows users to structure their reports on demand. 

Adding `--md` option to `dvc plots` command would definitely smooth the CML integration. It's not clear how to do with the current vega-based rendering but it seems doable if we extend the rendering options.

dvclive/live.py
```diff
@@ -23,7 +25,11 @@ class Live:
     DEFAULT_DIR = "dvclive"
 
     def __init__(
-        self, path: Optional[str] = None, resume: bool = False, **kwargs
+        self,
+        path: Optional[str] = None,
+        resume: bool = False,
+        report: bool = True,
```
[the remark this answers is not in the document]
dberenbaum: See also https://github.com/iterative/dvclive/issues/91#issuecomment-1029388277. Users shouldn't have to touch their code when going from local to cml, so I think we probably want this set through an environment variable or config file.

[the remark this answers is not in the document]
daavoo: > See also https://github.com/iterative/dvclive/issues/91#issuecomment-1029388277. Users shouldn't have to touch their code when going from local to cml, so I think we probably want this set through an environment variable or config file.

`html` currently uses env var because it relies on DVC integration. Once we extract the render and DVCLive is in charge of generating the HTML, would you still expect it to be handled by env var?

> I think we should look at local html reports and cml markdown reports as two different report formats, which could be extended to others. See https://github.com/iterative/dvclive/issues/90#issuecomment-1035120185.

I like the approach of considering`html` and `cml` as just different formats of `report`. However, I have some doubts, like would it make sense to support multiple report formats in a single `Live` instance?

I'm not sure about the env var, I understand the reasoning but we could make code the top priority while still supporting the env var scenario for CI/CD with CML:

```
report: Optional[typing.Literal["cml", "html"] ] = None
. . . 
if report is None:
    report = os.getenv("DVCLIVE_REPORT", "html")
```

This would preserve current behavior locally and allow to enable CML report via env var on CI/CD

daavoo:
After offline discussions we decided to went with a different direction

Hommus:
@daavoo I'd love to hear about where this is going in the future!

daavoo:
> @daavoo I'd love to hear about where this is going in the future!

Hi @Hommus , apologies for not adding more context.

The idea is that , instead of `dvclive` calling `cml`, `dvclive` will write a report containing the same info as in the [current `html` report](https://dvc.org/doc/dvclive/get-started#html-report) but in `markdown` format (i.e. `dvclive/report.md`.

The logic for writing this `markdown` report will be mostly implemented in the library we currently rely for the `html` report (https://github.com/iterative/dvc-render).

CML on their side will add a new feature (WIP https://github.com/iterative/cml/pull/1026 ) to "watch" for updates on this file and automatically call `cml send-comment` and `cml publish` (for the images)/

dberenbaum:
@Hommus How would you plan to use the feature?

Hommus:
@dberenbaum It was suggested in [#13](https://github.com/iterative/studio-support/issues/13) that this might be a possible solution for tracking metrics during an experiment run, through a web interface like Iterative Studio. The functionality would be similar to that of tensorboard or aimstack.io.

Issue #219, opened 2022-02-14
catalyst: Update integration


daavoo:
The latest version breaks current integration https://github.com/catalyst-team/catalyst/releases/tag/v22.02

Failure:

https://github.com/iterative/dvclive/runs/5183975198?check_suite_focus=true#step:6:156

PR #220, opened 2022-02-14
catalyst: restrict version

pared:

For `catalyst==22.02` our builds fail, need to investigate that, restricting the version for now.
Temporary workaround for #219

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ tf = ["tensorflow"]
 xgb = ["xgboost"]
 lgbm = ["lightgbm"]
 hugginface = ["transformers", "datasets"]
-catalyst = ["catalyst"]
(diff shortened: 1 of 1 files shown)


daavoo (approved):
Should we start doing this for all frameworks?

pared:
@daavoo I think that this is quite a challenge for us to decide. 
1. On one hand we don't want to restrict the dependencies, because that could interfere with users workspace, and `dvclive` is probably not that important to force user to install particular version of the DL framework.

2. On the other hand, we need to maintain our project and the wide range of supported frameworks makes it hard to be able to know which versions of particular frameworks we support. Solution to that would be restricting the supported versions, but that contradicts with 1. We can discuss this on our next meeting.

Issue #195, opened 2021-12-01
`html`: nest inside dvclive dir


dberenbaum:
See the complaint in https://discord.com/channels/485586884165107732/485596304961962003/915439735256801352, which more generally complains of too many dvc-related paths, but specifically calls out `dvclive_dvc_plots`. While they can configure the `dvclive` part, they will still have a path like `{dir}_dvc_plots`. It would be nice to nest the plots directory inside `{dir}` and not have to include `dvc` anywhere in the path since these are supposed to be general purpose outputs.

daavoo:
This has to be solved on DVC side, unless/until https://github.com/iterative/dvc/issues/6944 is implemented

daavoo:
The problem with the existing DVC integration would be that anything inside `{dvclive.dir}` is tracked equally by DVC

dberenbaum:
Even `{dir}_plots` seems nicer than `{dir}_dvc_plots`. What do you think?

 > This has to be solved on DVC side

True. We can transfer or keep discussion here and transfer/open a new issue once there is a decision.

PR #227, opened 2022-03-07
live: `html` report without DVC.

### LINKED ISSUE #213: html generation: make dvclive independent of dvc
In iterative/dvc#6944 we are considering extracting the rendering outside of DVC.
This issue would go nicely with #163 - because we could let both DVC and dvclive use the same rendering backend, which would make dvclive more functional. 

This unravels some problems:
1. Plot is not only some data, but also a configuration
2. DVC has a capability to store the configuration, dvclive does not

In order to provide such functionality, we need to let `dvclive` provide configuration to the potential plots library.
We discussed this issue with @dberenbaum and @daavoo and it seems that we could go two ways about it:

1. Let the `dvclive` render the data only at the runtime - we know what method has been called and can provide proper config from within the library - this however would make visualizing previously logged data hard. We could avoid that by saving the config  on DVC side (similar to what we do with `html` and `summary` now) - though that would make it work only in DVC + `dvclive` use case. Standalone `dvclive` would be limited.

2. Let the `dvclive` store the configuration - that seems like our desired behavior, though it poses a problem: how to structure the plot config and how to store it so that both standalone `dvclive` and `dvvclive` + DVC use cases work properly?
Plot config would probably be defined by iterative/dvc#7086
In the same issue we are intending to define plots specification inside `dvc.yaml` - but if we want `dvclive` to render too, we would need to store it somewhere non-DVC related. Initial idea would be to store it in `dvclive_path/.config`, but in this case, when we are using `dvclive` alongside DVC we will need to make sure DVC reads additional `plots` definitions from within `dvclive` dirs.

Maybe we could let `plots` section inside `dvc.yaml` address other file by path?
eg:
```
# dvc.yaml
plots:
   plots_configs: ['dvclive/.config.yaml']
   plot1:
   ...
   plot2:
   ...
   
```

Which would let DVC know where to look for additional plots configurations and "expand it" on `dvc plots show/diff`. That would also make `dvclive` independent of DVC in terms of plots generation.
An additional problem for this use case is templates: do we want to let users define their own plots templates? 
We will have to store them somewhere.

Did I miss something? cc @dberenbaum @daavoo

daavoo:
- Uses `dvc_render`.
- Add new method `Live.make_report` that users can call from a `Live` instance at any point. 

  This method is still called by default on `next_step` if the report has been enabled. I'm not sure about preserving that behavior and would move towards explicit `make_report` calls.
  
- Add new entry point `dvclive-show` that can generate the latest HTML report.

Defaults to the previous behavior, for now. `html` report only gets generated if enabled on the `dvc.yaml` side . However, users can manually enable `html` without DVC by passing:

```python
live = Live(report="html")
```

Closes #213 #195

### DIFF
--- a/dvclive/data/plot.py
+++ b/dvclive/data/plot.py
@@ -46,8 +46,22 @@ class Plot(Data):
             "DVCLive plots can only be used in no-step mode."
         )
 
+    @staticmethod
+    def get_properties():
+        raise NotImplementedError
+
 
 class Roc(Plot):
+    @staticmethod
+    def get_properties():
+        return {
+            "x": "fpr",
+            "y": "tpr",
+            "title": "Receiver operating characteristic (ROC)",
+            "x_label": "False Positive Rate",
+            "y_label": "True Positive Rate",
+        }
+
     def no_step_dump(self) -> int:
         from sklearn import metrics
 
@@ -64,6 +78,16 @@ class Roc(Plot):
 
 
 class PrecisionRecall(Plot):
+    @staticmethod
+    def get_properties():
+        return {
+            "x": "recall",
+            "y": "precision",
+            "title": "Precision-Recall Curve",
+            "x_label": "Recall",
+            "y_label": "Precision",
+        }
+
     def no_step_dump(self) -> int:
         from sklearn import metrics
 
@@ -72,7 +96,7 @@ class PrecisionRecall(Plot):
         )
 
         prc = {
-            "prc": [
+            "precision_recall": [
                 {"precision": p, "recall": r, "threshold": t}
                 for p, r, t in zip(precision, recall, prc_thresholds)
             ]
@@ -81,6 +105,16 @@ class PrecisionRecall(Plot):
 
 
 class Det(Plot):
+    @staticmethod
+    def get_properties():
+        return {
+            "x": "fpr",
+            "y": "fnr",
+            "title": "Detection error tradeoff (DET)",
+            "x_label": "False Positive Rate",
+            "y_label": "False Negative Rate",
+        }
+
     def no_step_dump(self) -> int:
         from sklearn import metrics
 
@@ -98,6 +132,17 @@ class Det(Plot):
 
 
 class ConfusionMatrix(Plot):
+    @staticmethod
+    def get_properties():
+        return {
+            "template": "confusion",
+            "x": "actual",
+            "y": "predicted",
+            "title": "Confusion Matrix",
+            "x_label": "True Label",
+            "y_label": "Predicted Label",
+        }
+
     def no_step_dump(self) -> int:
         cm = [
             {"actual": str(actual), "predicted": str(predicted)}
@@ -107,6 +152,16 @@ class ConfusionMatrix(Plot):
 
 
 class Calibration(Plot):
+    @staticmethod
+    def get_properties():
+        return {
+            "x": "prob_pred",
+            "y": "prob_true",
+            "title": "Calibration Curve",
+            "x_label": "Mean Predicted Probability",
+            "y_label": "Fraction of Positives",
+        }
+
     def no_step_dump(self) -> int:
         from sklearn import calibration
 
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -191,9 +191,9 @@ class Live:
     def make_report(self):
         if self._report == "html":
             html_report(self.dir, self.summary_path, self.html_path)
-        if self._auto_open:
-            open_file_in_browser(self.html_path)
-            self._auto_open = False
+            if self._auto_open:
+                open_file_in_browser(self.html_path)
+                self._auto_open = False
 
     def read_step(self):
         if Path(self.summary_path).exists():
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -15,7 +15,7 @@ from .error import (
     InvalidPlotTypeError,
 )
 from .report import html_report
-from .utils import nested_update
+from .utils import nested_update, open_file_in_browser
 
 logger = logging.getLogger(__name__)
 
@@ -28,12 +28,14 @@ class Live:
         path: Optional[str] = None,
         resume: bool = False,
         report: Optional[str] = "html",
+        auto_open: bool = True,
     ):
 
         self._path: Optional[str] = path
         self._resume: bool = resume
         self._report: str = report
         self._checkpoint: bool = False
+        self._auto_open: bool = auto_open
 
         self.init_from_env()
 
@@ -189,6 +191,9 @@ class Live:
     def make_report(self):
         if self._report == "html":
             html_report(self.dir, self.summary_path, self.html_path)
+        if self._auto_open:
+            open_file_in_browser(self.html_path)
+            self._auto_open = False
 
     def read_step(self):
         if Path(self.summary_path).exists():
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -8,12 +8,13 @@ from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
 from .data import DATA_TYPES, PLOTS, Image, Scalar
-from .dvc import make_checkpoint, make_html
+from .dvc import make_checkpoint
 from .error import (
     ConfigMismatchError,
     InvalidDataTypeError,
     InvalidPlotTypeError,
 )
+from .report import html_report
 from .utils import nested_update
 
 logger = logging.getLogger(__name__)
@@ -22,16 +23,26 @@ logger = logging.getLogger(__name__)
 class Live:
     DEFAULT_DIR = "dvclive"
 
-    def __init__(self, path: Optional[str] = None, resume: bool = False):
+    def __init__(
+        self,
+        path: Optional[str] = None,
+        resume: bool = False,
+        report: Optional[str] = "html",
+    ):
+
         self._path: Optional[str] = path
         self._resume: bool = resume
-        self._html: bool = True
+        self._report: str = report
         self._checkpoint: bool = False
 
         self.init_from_env()
 
         if self._path is None:
             self._path = self.DEFAULT_DIR
+
+        if self._report is not None:
+            out = Path(self.html_path).resolve()
+            print(f"Report will be saved at {out}")
 
         self._step: Optional[int] = None
         self._scalars: Dict[str, Any] = OrderedDict()
@@ -52,18 +63,12 @@ class Live:
                 Path(self.dir) / data_type.subfolder, ignore_errors=True
             )
 
-        if os.path.exists(self.summary_path):
-            os.remove(self.summary_path)
-
-        if os.path.exists(self.html_path):
-            shutil.rmtree(Path(self.html_path).parent, ignore_errors=True)
+        for f in {self.summary_path, self.html_path}:
+            if os.path.exists(f):
+                os.remove(f)
 
     def _init_paths(self):
-        if self._step is not None:
-            os.makedirs(self.dir, exist_ok=True)
-            if self._html:
-                os.makedirs(Path(self.html_path).parent, exist_ok=True)
-        os.makedirs(Path(self.summary_path).parent, exist_ok=True)
+        os.makedirs(self.dir, exist_ok=True)
 
     def init_from_env(self) -> None:
         from . import env
@@ -75,12 +80,14 @@ class Live:
 
             env_config = {
                 "_path": os.environ.get(env.DVCLIVE_PATH),
-                "_html": bool(int(os.environ.get(env.DVCLIVE_HTML, "0"))),
                 "_checkpoint": bool(
                     int(os.environ.get(env.DVC_CHECKPOINT, "0"))
                 ),
                 "_resume": bool(int(os.environ.get(env.DVCLIVE_RESUME, "0"))),
             }
+            if not bool(int(os.environ.get(env.DVCLIVE_HTML, "0"))):
+                env_config["_report"] = None
+
             for k, v in env_config.items():
                 if getattr(self, k) != v:
                     logger.info(
@@ -102,7 +109,7 @@ class Live:
 
     @property
     def html_path(self):
-        return str(self.dir) + "_dvc_plots/index.html"
+        return os.path.join(self.dir, "report.html")
 
     def get_step(self) -> int:
         return self._step or 0
@@ -119,8 +126,7 @@ class Live:
                 data.dump(data.val, self._step)
             self.make_summary()
 
-        if self._html:
-            make_html()
+        self.make_report()
 
         if self._checkpoint:
             make_checkpoint()
@@ -179,6 +185,10 @@ class Live:
 
         with open(self.summary_path, "w") as f:
             json.dump(summary_data, f, indent=4)
+
+    def make_report(self):
+        if self._report == "html":
+            html_report(self.dir, self.summary_path, self.html_path)
 
     def read_step(self):
         if Path(self.summary_path).exists():
--- a/dvclive/report.py
+++ b/dvclive/report.py
@@ -1,0 +1,82 @@
+import json
+from pathlib import Path
+
+from dvc_render.html import render_html
+from dvc_render.image import ImageRenderer
+from dvc_render.vega import VegaRenderer
+
+from dvclive.data import PLOTS, Image, Scalar
+from dvclive.data.plot import Plot
+from dvclive.utils import parse_tsv, to_base64_url
+
+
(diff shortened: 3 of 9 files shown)


dvclive/data/plot.py
```diff
@@ -46,8 +46,36 @@ def step_dump(self) -> None:
...
+    def get_properties(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def dump_template(cls, output_file):
+        template = cls.get_template()
+        content = template.content
+        for k, v in cls.get_properties().items():
+            content = template.fill_anchor(content, k, v)
+        Path(output_file).write_text(content)
```
daavoo, line 65: All this logic is to somehow allow preserving behavior with and without DVC by serializing templates. So users can set in DVC.yaml:

```
stages:
  test-dvclive:
    cmd: python test_dvclive.py
    plots:
      - dvclive/plots/confusion_matrix.json:
          template: dvclive/plots/confusion_matrix.template
```

DVCLive would be able to work without the serialized template.

Nor sure if the templates should be dumped to a separate subfolder (i.e. `dvclive/templates` instead of `dvclive/plots`) in order to be preserved across runs:

```
stages:
  test-dvclive:
    cmd: python test_dvclive.py
    plots:
      - dvclive/plots/confusion_matrix.json:
          template: dvclive/templates/confusion_matrix.template
```

dvclive/live.py
```diff
@@ -102,7 +105,7 @@ def summary_path(self):
 
     @property
     def html_path(self):
-        return str(self.dir) + "_dvc_plots/index.html"
+        return os.path.join(self.dir, "report.html")
```
daavoo, line 108: This is what closes #195 . Report is now generated in `dvclive/report.html`.

dvclive/live.py
```diff
@@ -180,6 +183,10 @@ def make_summary(self):
         with open(self.summary_path, "w") as f:
             json.dump(summary_data, f, indent=4)
 
+    def make_report(self, mode: str = "html"):
```
daavoo, line 186: Users can manually call this at any point from a `Live` instance

dvclive/report.py
```diff
@@ -0,0 +1,88 @@
...
+    dvclive_path = Path(dvclive_folder)
+    renderers = []
+    renderers.extend(get_scalar_renderers(dvclive_path / Scalar.subfolder))
+    renderers.extend(get_image_renderers(dvclive_path / Image.subfolder))
+    renderers.extend(get_plot_renderers(dvclive_path / Plot.subfolder))
+
+    render_html(renderers, output_html_path, refresh_seconds=5)
+
+
+def make_report():
```
daavoo, line 69: Added `CLI` utility that can reproduce the latest `html` report.

dvclive/data/plot.py
dberenbaum: Is this only necessary for `dvc plots`? Is serialization needed for dvclive to make the report?

dvclive/live.py
```diff
@@ -22,10 +23,16 @@
 class Live:
     DEFAULT_DIR = "dvclive"
 
-    def __init__(self, path: Optional[str] = None, resume: bool = False):
+    def __init__(
+        self,
+        path: Optional[str] = None,
+        resume: bool = False,
+        report: Optional[str] = None,
```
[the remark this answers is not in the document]
dberenbaum: Should we assert and raise an error if `report` is something other than `html`?

dvclive/live.py
dberenbaum: Hm, it seems odd that `report` is kwarg in `Live()`, but it doesn't get passed in `live.make_report()` calls (unlike `live.next_step()`). Seems like it should either default to whatever is passed in `Live()`, or `report` kwarg should be removed from `Live()`.

Related: It seems odd to pass `Live(report="html")` in no-step scenarios and have that kwarg do nothing. So maybe that's not the best place for it? Not sure 🤔 .

dvclive/report.py
dberenbaum: This is very cool and we may need it at some point, but I think it might be overkill in this PR. Does it add enough value yet over documenting `report.html_report()` and keeping everything in the Python API?

dvclive/report.py
```diff
@@ -0,0 +1,88 @@
...
+                data = data[name]
+            for row in data:
+                row["rev"] = "workspace"
+            plot.dump_template(file.with_suffix(".template"))
+            properties = {"template": file.with_suffix(".template")}
+            renderers.append(VegaRenderer(data, name, **properties))
+    return renderers
+
+
+def html_report(dvclive_folder, output_html_path):
```
dberenbaum, line 59: Not sure where to address this, but one thing that's missing from the dvc implementation is an output link and/or auto open of the report.

dberenbaum (commented):
🔥 This is great @daavoo. Still some big open questions, and I want to see how it works in DVC once that is updated, but exciting to see reports working without dvc.

dvclive/live.py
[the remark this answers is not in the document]
daavoo: > Should html be the default, at least for now? It matches current behavior when run with dvc.

It already matches the behavior when run with DVC, (at least while the `live` section is not deprecated) with:

https://github.com/iterative/dvclive/blob/90c46ea602a2bd04052b334fab57e6ba04940c36/dvclive/live.py#L84-L85

dvclive/data/plot.py
daavoo: > Is this only necessary for dvc plots?

Yes

> Is serialization needed for dvclive to make the report?

It's not needed, DVCLive could generate the report without dumping templates.

dvclive/report.py
daavoo: > Does it add enough value yet over documenting report.html_report() and keeping everything in the Python API?

Probably it doesn't.

dvclive/live.py
daavoo: I'm not sure either.

It might be better to remove `mode` from `make_report` and make the `report` arg of `Live` the only source of truth.

One problem with the alternative (removing `report` from `Live`) is that integrations become a little more complicated because right now users can customize `Live` args but in this case we would need to handle `report` separately

dvclive/data/plot.py
dberenbaum: IMO it might be better to at least put this in a follow-up PR because it might need more discussion or a different approach might come up. Even without it, we are achieving parity with current DVC behavior, right?

dvclive/live.py
dberenbaum: > One problem with the alternative (removing `report` from `Live`) is that integrations become a little more complicated because right now users can customize `Live` args but in this case we would need to handle `report` separately

Got it, makes sense.

> It might be better to remove `mode` from `make_report` and make the `report` arg of `Live` the only source of truth.

Since it's a method of `Live`, I would expect it to use `self._report` unless I override it, so what about something like this?

```
    def make_report(self, mode: Optional[str] = None):
           if not mode:
               mode = self._report
```

Also, maybe we should rename in one place or the other so it is consistently either `report` or `mode`?

dvclive/report.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def get_image_renderers(images_folder):
+    renderers = []
+    for suffix in Image.suffixes:
+        for file in Path(images_folder).rglob(f"*{suffix}"):
+            name = str(file.relative_to(images_folder))
+            data = [
+                {
+                    ImageRenderer.SRC_FIELD: to_base64_url(file),
```
dberenbaum, line 34: So this changes current dvc behavior, right? @pared Do you remember why we decided to include refs to all image files instead of encoding them directly into the HTML?

dvclive/live.py
dberenbaum: Will it be an issue if the dvclive dir is marked as a plots output and there is already an html file inside it?

daavoo: > Will it be an issue if the dvclive dir is marked as a plots output and there is already an html file inside it?

Not sure if an issue but the HTML will be tracked by whichever config they set for the `plots` output (so HTML might be added to DVC cache).

It's not strictly related with this P.R. but one of the directions I also wanted with deprecating `live` was to encourage more granular output config for DVCLive outputs which is not currently possible:

```yaml
...
plots:
  - dvclive/scalars
      cache: false
  - dvclive/images
metrics:
  # I mean to open issue to discuss also nesting summary
  - dvclive/summary.json:
      cache: false
```

```
#.gitignore
dvclive/report.html
```

dberenbaum: Also, looks like the dvc output will be wrong once this is merged:

```
Live summary will be created at:
/Users/dave/Code/dvc-checkpoints-mnist/dvclive_dvc_plots/index.html
```

How do we sequence these PRs to keep everything working as expected? Would it be better to close #295 in a follow-up PR?

dvclive/live.py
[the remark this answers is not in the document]
dberenbaum: > It already matches the behavior when run with DVC, (at least while the `live` section is not deprecated) with:
> 
> [](https://github.com/iterative/dvclive/blob/90c46ea602a2bd04052b334fab57e6ba04940c36/dvclive/live.py#L84-L85)

True. Does it cause an issue to default to `report = "html"`? Any other reason you wouldn't want it to be the default?

AFAIK the only reason we didn't always generate it before was because we couldn't without dvc. We used to have `html = True` by default.

dvclive/live.py
daavoo: A subsequent DVC P.R. should be created to fix that message but more importantly to deactivate the HTML generation on DVC side. 

`dvclive` is pinned in DVC so I see the sequence as:

Merging https://github.com/iterative/dvc/pull/7409 > Approval of this P.R. > New P.R. in DVC fixing message and disabling generation.

dberenbaum: One other discrepancy I noticed:

https://github.com/iterative/dvc/blob/6a6955e27cf69490c7f694980dadfd3f7a5f9949/dvc/repo/live.py#L22

The dvc-generated html includes metrics, but this PR does not. Is there a plan to include metrics in a future PR?

dberenbaum: > deactivate the HTML generation on DVC side

Should html still be generated now in `dvclive_dvc_plots`? I'm not seeing that happen.

dberenbaum: I'm confused what's happening now in `dvclive_dvc_plots` with these changes. It doesn't look like it's being updated live anymore after this PR.

dvclive/report.py
pared: There wasn't any necessity that would make us use ``. We decided to go this way to avoid producing gigantic htmls which would be troubling to parse/inspect.

dvclive/live.py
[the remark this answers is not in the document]
daavoo: Made `html` default for all cases

dvclive/report.py
daavoo: Removed

dvclive/report.py
daavoo: Added message to `Live` init. Need to add auto open

dvclive/live.py
daavoo: > I'm confused what's happening now in dvclive_dvc_plots with these changes. It doesn't look like it's being updated live anymore after this PR.

If you install this P.R. and https://github.com/iterative/dvc/pull/7514 (follow up for DVC once this P.R. gets merged), then :

- `dvclive_dvc_plots` should not be generated at all
- The new HTML report (with correct refreshing) should be generated at `{dvclive_dir}/report.html`.

dvclive/data/plot.py
daavoo: > IMO it might be better to at least put this in a follow-up PR because it might need more discussion or a different approach might come up. Even without it, we are achieving parity with current DVC behavior, right?

Removed from this P.R.

dvclive/live.py
pared: If we are separating this, we might as well remove `dvc/commands/live` functionality. We created for visualizing `dvclive` results, but  now `dvclive` will be able to do it by itself.

dvclive/live.py
```diff
@@ -180,6 +188,13 @@ def make_summary(self):
         with open(self.summary_path, "w") as f:
             json.dump(summary_data, f, indent=4)
 
+    def make_report(self):
+        if self._report == "html":
+            html_report(self.dir, self.summary_path, self.html_path)
+        if self._auto_open:
```
pared, line 194: We should probably do it only in case of existing `self._report`. For example:
```
python -c "from dvclive import Live;logger=Live('ble',report=None);logger.log('value', 10);logger.next_step()"
```
will raise some errors, which I wouldn't expect, since I specified that I don't want  a report.

setup.py
```diff
@@ -36,6 +36,7 @@ def run(self):
         _build_py.run(self)
 
 
+render = ["dvc_render"]
```
dberenbaum, line 39: Should this be optional? It seems like this is the basic functionality of dvclive.

setup.py
```diff
@@ -69,6 +82,7 @@ def run(self):
     packages=find_packages(exclude="tests"),
     description="Metric logger for ML projects.",
     long_description=open("README.rst", "r", encoding="UTF-8").read(),
+    install_requires=render,
```
pared, line 85: So I cannot say that `dvclive` has no dependencies anymore? :(

dvclive/report.py
```diff
@@ -0,0 +1,82 @@
...
+    if summary_path.exists():
+        return {
+            "": {
+                "data": {
+                    summary_path.name: {
+                        "data": json.loads(summary_path.read_text())
+                    }
+                }
+            }
+        }
```
pared, line 68: `dvclive` gets polluted with the internal `dvc` data structure, I will create an issue on the dvc-render side to get it fixed.

pared (changes requested):
One thing to fix.

dvclive/report.py
pared: Well, i won't, its already there:
https://github.com/iterative/dvc-render/issues/19

setup.py
dberenbaum: As long as we can say it's "lightweight" and people are able to quickly install it on any system, I don't think we should worry about it.

dvclive/report.py
dberenbaum: Looking at this again, it seems like it could result in giant htmls if there's enough image data. It also duplicates what's already saved in `dvclive/images`. Can we refer to those image files instead? Is there a reason to prioritize encoding them?

daavoo: > Can we refer to those image files instead? 

Yes, we can.

> Is there a reason to prioritize encoding them?

No strong reason, just making the HTML standalone, similar to how the plot data is embedded. If we refer to the image folder the report might break if someone remove/alter images from the folder.

TBH I think the vega embedding takes proportionally much more space per item than the base64 encoded string for each image.

dvclive/live.py
dberenbaum: > It seems odd to pass Live(report="html") in no-step scenarios and have that kwarg do nothing. 

Not sure if it should be part of this PR, but currently, I can use the `live` section in `dvc.yaml` to get a report even in no-step scenarios, but it doesn't seem like it's possible without the `live` section. Is there some way to still generate the report by default?

dvclive/report.py
daavoo: To clarify, the duplication is not introduced in this P.R. 

In the current DVC HTML, image data is being duplicated and saved in a separate directory (https://github.com/iterative/dvc/issues/6665)

dvclive/live.py
daavoo: >  Is there some way to still generate the report by default?

When using the `live` section in no-step scenario, the report gets generated by the monitor task after the stage has ended. 
Any implementation on DVCLive side would behave differently, as the `Live` object doesn't have info of when the "stage" has ended (the closest would be to implement the logic in `__del__`) .
IMO would be better to document `Live.make_report()` and emphasize it's usage for non step scenarios

setup.py
pared: Yeah, I'm just joking. Though, to be honest, there was something magical to the "It has NO dependencies". :)

pared (approved):

dvclive/report.py
dberenbaum: True, although in this case we already have everything consolidated into a folder. Anyway, we can keep as is for now. No need to block the PR.

dvclive/live.py
dberenbaum: Extracted to #231

dberenbaum (approved):

Issue #201, opened 2021-12-15
`dvc`: Add option to disable automatic open


daavoo:
Users might want to have html generation without automatically open the browser.

shcheklein:
Question, what was the initial motivation to make this behavior default? It feels a bit too "intrusive" since it changes my context automatically.

shcheklein:
And a bit more context, having this option is critical for VS Code. And probably we'll want to have a way to override the default behavior (e.g. env var).

daavoo:
> Question, what was the initial motivation to make this behavior default? It feels a bit too "intrusive" since it changes my context automatically.

No strong motivation. We decided to make it default and wait for users asking for an option to opt out.
My personal bias was that if you generate the live HTML it is to look at it during training (otherwise you could just use DVC plots). I was aware it could be a too aggressive default.

 > And a bit more context, having this option is critical for VS Code. And probably we'll want to have a way to override the default behavior (e.g. env var).

I think HTML generation doesn't make sense for users of the VSCode extension (live plots of extension replace it). Based on that assumption, VSCode could be already re-using the existing env var `DVCLIVE_HTML` (it already defaults to false) to disable HTML generation at all.  @shcheklein @iterative/vs-code ?

Regardless of that env var, I would add a specific env var for the web browser open.

Issue #180, opened 2021-10-19
`dvc`: Option to automatically open the html report


daavoo:
`dvc plots {show|diff}` have an `--open` option that "opens the generated plot in the browser automatically".

Coming from discussions in https://github.com/iterative/dvc/pull/6815 it looks like having a similar option to automatically open the DVCLive HTML report on the browser would be nice to have.

In addition to whether adding this option is useful (I think it is), questions are where and how would the option be added.

From the top of my head, an additional property to the field in `live` section of `dvc.yaml` would be nice:

```yaml
stages:
   stages:
  train:
    cmd: python train.py
    live:
      dvclive:
        summary: true
        html: true
        open: true
```

As a drawback, this would probably require adding yet another parameter to the long list of `dvc stage add` / `dvc run` options (`--live-open`?).

pared:
I am not a user, but the inability to turn this off is tiring. My use case is somewhat special because I just run reproduction scripts and mostly don't care about visualization besides the fact whether it renders or not. 

I am not sure that dvclive should be opening created HTML. AFAIK `--open` for the `show` was added as a convenience method. Now we start to make users do something, without explicit commands. I think this is controversial.

daavoo:
> I am not sure that dvclive should be opening created HTML. AFAIK `--open` for the `show` was added as a convenience method. Now we start to make users do something, without explicit commands. I think this is controversial and a breaking change.

WDYT about  adding `open` option to `live` section in `dvc.yaml`, defaulting to `true`? That would give a way for users to opt out.

Somehow related: https://github.com/iterative/dvc/issues/6091

My personal experience was always using `--open` so I might be biased to open by default,  but I agree that it could be considered by some users as too aggressive/breaking change.

pared:
> WDYT about adding open option to live section in dvc.yaml, defaulting to true? That would give a way for users to opt out.

If we want to make it a default behavior, I would rather consider adding this as DVC option - because it will be an option related to one's workflow rather than particular output. And this is where the problems start - should we introduce a settings option, in DVC, related to dvclive? Sounds bad. But then, if someone wants to turn it off, he/she will have to go through all `dvclive` outputs and make them `open: false`. It's hard to determine until users start to complain. I think I would read `DVC_TEST` in `ui` module and disable call if it's true, that would also solve iterative/dvc#7014.

PR #243, opened 2022-04-20
live: Handle auto_open via env.

### LINKED ISSUE #241: `auto_open`: how to handle it
During discussion with @mattseddon, it came up that changing `auto_open` default to `True` broke the VS Code extension. This has since been reverted in #240, but there are still questions about how to prevent this behavior and when to automatically open the report.

Some issues with `auto_open` behavior:
* How do we ensure this doesn't happen again in VS Code or in other scenarios like CI where it's not needed? Should we have a way to override it through something like an environment variable, or always respect whether dvclive is being run in a tty?
* If we will have non-html report formats, will `auto_open` make sense?
* Should there be an explicit method to open the report (see #156 and #231)?
* How can we keep simple the basic scenario of generating a live-updated metrics report?

daavoo:
Remove `auto_open` arg.
Add DVCLIVE_OPEN env var.
Add env2bool utils.

Closes #241


tests/test_report.py
```diff
@@ -68,14 +71,17 @@ def test_make_report_open(tmp_dir, mocker):
...
+
     live = Live(report=None)
     live.log("foo", 1)
     live.next_step()
 
     assert not mocked_open.called
 
+    monkeypatch.setenv(DVCLIVE_OPEN, True)
+
     mocked_open = mocker.patch("webbrowser.open")
```
pared, line 83: Do we need to patch it a second time? Seems like the one from line 73 should be enough.

dvclive/env.py
```diff
@@ -1,3 +1,4 @@
+DVCLIVE_OPEN = "DVCLIVE_OPEN"
 DVCLIVE_PATH = "DVCLIVE_PATH"
 DVCLIVE_SUMMARY = "DVCLIVE_SUMMARY"
 DVCLIVE_HTML = "DVCLIVE_HTML"
```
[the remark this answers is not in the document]
dberenbaum: Ah, I guess it's needed for backwards compatibility, right?

dvclive/live.py
```diff
@@ -196,9 +191,8 @@ def make_summary(self):
     def make_report(self):
         if self._report == "html":
             html_report(self.dir, self.summary_path, self.html_path)
-            if self._auto_open:
+            if env2bool(env.DVCLIVE_OPEN):
```
dberenbaum, line 194: What should the default be? Should we try to detect tty if no environment variable is set? Are there ever cases where the report was auto opened by default before?

dvclive/env.py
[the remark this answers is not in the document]
daavoo: > Ah, I guess it's needed for backwards compatibility, right?

Yes. We need it until 3.X where live support is fully dropped

dvclive/live.py
daavoo: > What should the default be?

It's currently False and I guess that's what should be

> Should we try to detect tty if no environment variable is set?

I'm not sure if it's worth the effort. We provide a way for users to handle the option in different environments, let's be explicit in docs about how to use it.

> Are there ever cases where the report was auto opened by default before?

Only if `dvc config plots.auto_open` was set to True.

tests/test_report.py
daavoo: We don't, updated!

dvclive/live.py
dberenbaum: My worry is that the point of auto open is to make the simplest use case (showing the live report) as easy as possible, and this creates an additional step. It almost seems pointless to have the environment variable, since it's mostly for people who don't read the docs or can't immediately figure out how to open the report.

Maybe it's not really an issue. I think it's fine to merge, but I'm curious about whether auto open is ever needed and if there's some way to make this scenario useful.

daavoo: @mattseddon Would it be okay for you if we make the default `True`? vscode would need to always set the `DVCLIVE_OPEN` var in order to prevent the behavior

dberenbaum: I realized we have had this discussion before (multiple times 😅):

* https://github.com/iterative/dvc/pull/5584
* https://github.com/iterative/dvc/issues/6091
* https://github.com/iterative/dvclive/issues/180
* https://github.com/iterative/dvclive/issues/201

I don't think it's worth revisiting, so let's not open by default.

It might be enough to hint about the environment variable like in https://github.com/iterative/dvc/blob/f84f9cf1846554a13f388693220378edb8d17be6/dvc/commands/plots.py#L111.

dberenbaum: > It might be enough to hint about the environment variable like in https://github.com/iterative/dvc/blob/f84f9cf1846554a13f388693220378edb8d17be6/dvc/commands/plots.py#L111.

@daavoo What do you think about this idea?

PR #244, opened 2022-04-26
Checkpoint without live

daavoo:
Oversight from `live` deprecation.

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -32,9 +32,8 @@ class Live:
     ):
 
         self._path: Optional[str] = path
-        self._resume: bool = resume
+        self._resume: bool = resume or env2bool(env.DVCLIVE_RESUME)
         self._report: str = report
-        self._checkpoint: bool = False
         self.html_path = None
 
         self.init_from_env()
@@ -83,8 +82,6 @@ class Live:
 
             env_config = {
                 "_path": os.getenv(env.DVCLIVE_PATH),
-                "_checkpoint": env2bool(env.DVC_CHECKPOINT),
-                "_resume": env2bool(env.DVCLIVE_RESUME),
             }
 
             # Keeping backward compatibility with `live` section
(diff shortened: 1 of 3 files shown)


dvclive/live.py
```diff
@@ -130,8 +133,7 @@ def set_step(self, step: int) -> None:
 
         self.make_report()
 
-        if self._checkpoint:
-            make_checkpoint()
+        self.make_checkpoint()
```
daavoo, line 136: Always call it, the `env.DVC_CHECKPOINT` is what would determine if we actually generate checkpoint or not. This allows to work with checkpoints even without the `live` section

dvclive/live.py
```diff
@@ -194,6 +196,10 @@ def make_report(self):
             if env2bool(env.DVCLIVE_OPEN):
                 open_file_in_browser(self.html_path)
 
+    def make_checkpoint(self):
+        if env2bool(env.DVC_CHECKPOINT) and dvc_checkpoint:
+            dvc_checkpoint()
```
daavoo, line 201: We can't really reach this point if we are not in DVC context so we can use `dvc.api` instead of our implementation.

---
`dvc.api.make_checkpoint` also checks if `env.DVC_CHECKPOINT` exists or not but I prefer to have this explicit check here even if it's redundant (also in case `dvc.api` changes).

pared (commented):
LGTM, though its worth remembering that relying on `dvc` to exist to use `DVC_CHECKPOINTS` prevents us from "cross-installation" usage. Eg. DVC is installed as binary, and we have a special virtual env for training. Now we will require users to use the same env to use DVC and `dvclive`. Not sure how popular this use case is.

daavoo:
> LGTM, though its worth remembering that relying on `dvc` to exist to use `DVC_CHECKPOINTS` prevents us from "cross-installation" usage. Eg. DVC is installed as binary, and we have a special virtual env for training. Now we will require users to use the same env to use DVC and `dvclive`. Not sure how popular this use case is.

Good point. I missed that. Re-added the `dvc` module and only implemented the fixing logic in the P.R.

Issue #177, opened 2021-10-14
`dvc`: prevent printing html path


sirily:
I'm using dvclive with pytorch lightning in dvc pipeline. While training there is a string with the path to metrics html, and it breaks progress bar and fills the console with useless messages. Is there a way to disable printing this line?
> Epoch 10:   8%|█▊                      | 103/1361 [00:11<02:13,  9.42it/s, loss=1, v_num=_run, train_positive_step=0.988]
> file:///path/index.html
> Epoch 10:  15%|███▋                    | 209/1361 [00:22<02:01,  9.51it/s, loss=1, v_num=_run, train_positive_step=0.985]
> file:///path/index.html
> Epoch 10:  22%|█████▍                  | 305/1361 [00:32<01:50,  9.53it/s, loss=1, v_num=_run, train_positive_step=0.991]
> file:///path/index.html
> Epoch 10:  29%|███████                 | 400/1361 [00:42<01:40,  9.53it/s, loss=1, v_num=_run, train_positive_step=0.984]
> file:///path/index.html
> Epoch 10:  37%|████████▉               | 505/1361 [00:53<01:29,  9.53it/s, loss=1, v_num=_run, train_positive_step=0.990]
> file:///path/index.html

Inside train stage logger is defined like this:
```
logger = DvcLiveLogger()
model = Trainer(hparams, model_path)
trainer = pl.Trainer(
                    gpus=1,
                    logger=logger,
                    max_epochs=hparams['num_epochs'],
                    log_every_n_steps=hparams['log_every_n_steps'],
                    callbacks=[checkpoint_callback, early_stopping],
                    stochastic_weight_avg=True)
```
in dvc.yaml stage looks like this:
```
cmd: python3 pipeline/train.py --task ${item.task} --inp_dir ${item.in} --out_dir ${item.out}
deps:
- data/processed/train_${item.task}_df.par
- data/processed/val_${item.task}_df.par
- models/${item.in}
- pipeline/train.py
- src/trainer.py
params:
- ${item.task}_train
outs:
- models/${item.out}
live:
  metrics/${item.out}_train:
    cache: false
    summary: true
    html: true
```
It seems like nothing bad happens in dvclive, maybe it prints because of summary path [here](https://github.com/iterative/dvc/blob/e5cf9b203df01f13e5568c745906d74bc131235c/dvc/repo/live.py#L25). But @shcheklein asked me to open an issue here.

dvc doctor:

> DVC version: 2.7.4 (pip)

[...]

PR #247, opened 2022-04-29
Removes noisy and misleading `print()` in init

shcheklein:
- It's not very expected from a lib to have log messages, especially using `prints`
- `Report will be saved at` is printed in some cases but actual report is not created (e.g. if we use dvclive only as a tool to write plots, metrics, images)

----------------

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -44,10 +44,6 @@ class Live:
         if self.html_path is None:
             self.html_path = os.path.join(self.dir, "report.html")
 
-        if self._report is not None:
-            out = Path(self.html_path).resolve()
-            print(f"Report will be saved at {out}")
-
         self._step: Optional[int] = None
         self._scalars: Dict[str, Any] = OrderedDict()
         self._images: Dict[str, Any] = OrderedDict()


daavoo (commented):
I'm ok with this. 

If I remember correctly, we added this to maintain parity with the previous behavior when the HTML was generated by DVC, where DVC will print this message at the beginning of running the stage.

@dberenbaum WDYT?

dberenbaum:
> * It's not very expected from a lib to have log messages

IMO the primary scenario for dvclive is seeing a live-updated report of the model being trained. We have already made some compromises about the ease of seeing this report by not opening it by default (see the discussion in https://github.com/iterative/dvclive/pull/243#discussion_r856281838). If we don't provide the user with the report path, it will be harder for users to even know the report is being generated and to know the path of the report. Do you have an alternate suggestion?

> especially using `prints`

> * `Report will be saved at` is printed in some cases but actual report is not created (e.g. if we use dvclive only as a tool to write plots, metrics, images)

These seem like implementation details/bugs rather than reasons to remove the feature. Do you still think it should be removed if these are resolved @shcheklein?

shcheklein:
>  Do you have an alternate suggestion?

At least have a clear way to silence it? Usually it's fine for libs to have loggers, but there is a path to toggle them on / off.

The cleanest way to communicate back, would be probably some callback or some env variable that _any_ consumer can read and decide what has to be printed.

> Do you still think it should be removed if these are resolved @shcheklein?

If bugs are solved I don't feel it's that critical to remove this. You decide folks, I can probably see if I can rework this to solve the bug itself, not removing the whole thing.

daavoo:
> At least have a clear way to silence it? Usually it's fine for libs to have loggers, but there is a path to toggle them on / off.
> 
> The cleanest way to communicate back, would be probably some callback or some env variable that any consumer can read and decide what has to be printed.

Added logger and env var in #250 

> Report will be saved at is printed in some cases but actual report is not created (e.g. if we use dvclive only as a tool to write plots, metrics, images)
> These seem like implementation details/bugs rather than reasons to remove the feature.

The message is only logged if `report` has been passed to `Live` instantiation. This results in a lie when users don't call `next_step` or `make_report`.

However, showing the message when we know for a fact that report has been created is tricky as it can mess up with the ML frameworks progress bars, that's why we moved the message on DVC from the end of `step` to instantiation (https://github.com/iterative/dvclive/issues/177)

shcheklein:
>  that's why we moved the message on DVC from the end of step to instantiation 

Change the message then at least to be more explicit? "Report path (if generated): .... See more on report here ...."?

shcheklein:
Looks even uglier now, folks :(

`INFO:dvclive:Report path (if generated): /Users/ivan/Projects/example-get-started/evaluation/report.html`

PR #250, opened 2022-05-03
live: Add dvclive logger and `DVCLIVE_LOGLEVEL`.

### LINKED ISSUE #206: `log`: Verbose mode
It would be nice to have a way to see what dvclive is doing instead of having it operate silently, so that stdout/loggers could optionally have info about when dvclive calls are made and what they are logging. This would be especially helpful in debugging.

daavoo:
Closes #206


pared:
LGTM. Tested on a actual repo using `matplotlib` too. Seems like setting the `DEBUG` level will result in logs from this lib too. Not sure how to approach that. Probably we should not propagate it down to other used libraries.

daavoo:
> LGTM. Tested on a actual repo using `matplotlib` too. Seems like setting the `DEBUG` level will result in logs from this lib too. Not sure how to approach that. Probably we should not propagate it down to other used libraries.

Updated P.R. It should no longer affect other libs, could you confirm?

dvclive/live.py
```diff
@@ -46,7 +47,7 @@ def __init__(
 
         if self._report is not None:
             out = Path(self.html_path).resolve()
-            print(f"Report will be saved at {out}")
+            logger.info(f"Report will be saved at {out}")
```
daavoo, line 50: Can be a good time to address https://github.com/iterative/dvclive/pull/247#issuecomment-1116425371 . Any suggestions?

> Report path (if generated): 

^ Reads fine to me

pared:
I can confirm that now I see only `dvclive` logs for `DEBUG` option.

pared (approved):

dvclive/live.py
dberenbaum: Yup, I think that's fine.

dberenbaum (approved):

PR #259, opened 2022-06-09
integrations: Standardize metric name to group under subfolders.

daavoo:
This attempts to standardize outputs across frameworks using a format that is more convenient for our UIs (Studio, VScode)

Pytorch Lightning example:

- Before

imagen

- After

imagen

### DIFF
--- a/dvclive/utils.py
+++ b/dvclive/utils.py
@@ -73,3 +73,32 @@ def env2bool(var, undefined=False):
     if var is None:
         return undefined
     return bool(re.search("1|y|yes|true", var, flags=re.I))
+
+
+def standardize_metric_name(metric_name: str, framework: str) -> str:
+    """Map framework-specific format to DVCLive standard.
+
+    Use `{split}/` as prefix in order to seperate by subfolders.
+    Use `{train|eval}` as split name.
+    """
+    if framework == "dvclive.fastai":
+        metric_name = metric_name.replace("train_", "train/")
+        metric_name = metric_name.replace("valid_", "eval/")
+
+    elif framework == "dvclive.huggingface":
+        for split in {"train", "eval"}:
+            metric_name = metric_name.replace(f"{split}_", f"{split}/")
+
+    elif framework == "dvclive.keras":
+        if "val_" in metric_name:
+            metric_name = metric_name.replace("val_", "eval/")
+        else:
+            metric_name = f"train/{metric_name}"
+
+    elif framework == "dvclive.lightning":
+        parts = metric_name.split("_")
+        if len(parts) > 2:
+            split, *rest, freq = parts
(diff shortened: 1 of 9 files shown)


dvclive/utils.py
```diff
@@ -73,3 +73,32 @@ def env2bool(var, undefined=False):
     if var is None:
         return undefined
     return bool(re.search("1|y|yes|true", var, flags=re.I))
+
+
+def standardize_metric_name(metric_name: str, framework: str) -> str:
+    """Map framework-specific format to DVCLive standard.
+
+    Use `{split}/` as prefix in order to seperate by subfolders.
+    Use `{train|eval}` as split name.
```
dberenbaum, line 82: Changing separators from `_` to `/` seems like a pretty straightforward and transparent change. Trying to standardize split names across all frameworks might be a bit too opinionated and unexpected. Is there some scenario where the metrics are a mess if we don't do this standardization?

daavoo: > Trying to standardize split names across all frameworks might be a bit too opinionated and unexpected. 
I just saw the opportunity to apply some friendly tyranny 😅 after seeing that many other ML loggers don't respect the original name passed by the framework.

> Is there some scenario where the metrics are a mess if we don't do this standardization?

Not really although for some (i.e. pytorch lightning) we do need to apply some "unexpected" refactor because of the `_{epoch|step}` suffix

dberenbaum: > I just saw the opportunity to apply some friendly tyranny 😅 after seeing that many other ML loggers don't respect the original name passed by the framework.

Ah, I didn't realize that was the case.

I guess replacing all `_` with `/` will be too aggressive, even if documented, so maybe we are stuck with some decisions on when to split. I'm still not sure we should automatically rename `valid` to `eval` or add `train` prefixes, for example. If you feel strongly about it, though, I don't think it's worth blocking the PR over.

dberenbaum (approved):

pared (commented):
In the loggers that we modify - is it possible to log your own metrics?
What if I want to log `this_is_very_informative_name_that_should_probably_not_be_split`?

daavoo:
> Ah right, sorry, should have read it more in depth. Then what about the situation when in one project I log `train_very_complicated_name` and in another `training_some_other_name`? 

Well, **before** this P.R, if you log those names in PyTorch lightning, you will end with `train_train_very_complicated_name` and `train_training_some_other_name` which is probably why manually adding prefixes explicitly to the metric name is not a good idea when using ML logger integrations. 

> I do realize that we can document it properly, but the behavior for the mentioned example could be confusing.

I can add an option for disabling the behavior.

pared (approved):

Issue #231, opened 2022-04-01
`report`: how to make report in no-step scenarios


dberenbaum:
> It seems odd to pass Live(report="html") in no-step scenarios and have that kwarg do nothing. 

Not sure if it should be part of this PR, but currently, I can use the `live` section in `dvc.yaml` to get a report even in no-step scenarios, but it doesn't seem like it's possible without the `live` section. Is there some way to still generate the report by default?

_Originally posted by @dberenbaum in https://github.com/iterative/dvclive/pull/227#discussion_r839895032_

dberenbaum:
>  Is there some way to still generate the report by default?

When using the `live` section in no-step scenario, the report gets generated by the monitor task after the stage has ended. 
Any implementation on DVCLive side would behave differently, as the `Live` object doesn't have info of when the "stage" has ended (the closest would be to implement the logic in `__del__`) .
IMO would be better to document `Live.make_report()` and emphasize it's usage for non step scenarios

_Originally posted by @daavoo in https://github.com/iterative/dvclive/pull/227#discussion_r840416323_

dberenbaum:
It seems clunky to me to have the report generated automatically in step scenarios but not in no-step scenarios.

Possible ways to make the experiences consistent:

1. Never auto-generate reports and require users to call `Live.make_report()` after every `Live.next_step()` call. This fits with the explicit canonical workflow emphasized in #156, but it's starting to get to be a lot of methods that need to be called for what used to be possible with just `dvclive.log()`.
2. Invoke `make_report` whenever `Live.log()` is called and `Live._step is None`. We already have some special logic for how to log things in no-step scenarios, so this might not be too different, although partial reports may be generated, which might seem awkward. It also could make it more complicated to determine when to send reports for things like #91.

Would either of these improve the experience? Are there other ideas?

daavoo:
> Would either of these improve the experience? Are there other ideas?

I would like to go with `1`.

> it's starting to get to be a lot of methods that need to be called for what used to be possible with just `dvclive.log()`.

I think that the main DVCLive use case would be based on integrations for step scenarios. In that use case, we will still be creating the report "automatically" by calling `make_report` at the end of each epoch.

For scenarios where users need to manually call the API, my personal preference is to have explicit methods, each one doing a single thing. 

We would have 3 methods for 3 different actions (logging data, updating step, generating report). IMO opinion this is a better API than having a method called `log` but having to explain that it also does 3 other things (counting checkpoints).

Issue #223, opened 2022-02-22
1.0 checklist


pared:
- [x] #225 
- [x] #272
- [x] #246 
- [x] #273 
- [x] #326
- [x] #332
- [x] #231
- [x] #323
- [x] #340
- [x] https://github.com/iterative/dvclive/issues/238
- [x] Context manager https://github.com/iterative/dvclive/pull/354
- [x] #172 
- [x] #103

Nice to have: https://github.com/iterative/dvclive/issues/223#issuecomment-1285109650

dberenbaum:
>     * [ ]   [integrations: How to manage versions #221](https://github.com/iterative/dvclive/issues/221)
>
>     * [ ]  [`step`: Remove methods in favor of `@property`. #172](https://github.com/iterative/dvclive/issues/172)
> 
>     * [ ]  [Complete examples for ML frameworks #283](https://github.com/iterative/dvclive/issues/283)
> 
>     * [ ]  [Decouple epoch metrics logging and checkpoints #103](https://github.com/iterative/dvclive/issues/103)

Do you think any of these are a priority for 1.0 @daavoo?

daavoo:
Nice to have:

- [ ]  #221 
- [ ] #172
- [ ] #283 
- [ ] #103

dberenbaum:
More nice to have:

- [ ] https://github.com/iterative/dvclive/issues/238
- [ ] https://github.com/iterative/dvclive/issues/310
- [ ] https://github.com/iterative/dvclive/issues/311
- [ ] https://github.com/iterative/dvclive/issues/271
- [ ] https://github.com/iterative/dvclive/issues/224

@daavoo Thoughts on these?

dberenbaum:
@daavoo Should #238 be added to the checklist? I think everything else besides docs is complete.

daavoo:
> @daavoo Should https://github.com/iterative/dvclive/issues/238 be added to the checklist? I think everything else besides docs is complete.

Moved to the top checklist and there is a P.R. #353 
> @daavoo Thoughts on these?

All look like non-breaking changes that can be prioritized but shouldn't block 1.0

dberenbaum:
What about https://github.com/iterative/dvclive/issues/172 and https://github.com/iterative/dvclive/pull/354?

PR #266, opened 2022-07-07
report: Introduce `markdown` support.

### LINKED ISSUE #91: integrations: CML

daavoo:
Depends on https://github.com/iterative/dvc-render/pull/69
Closes #91 

The new default value for `report` is `auto`. 
If `CI` env var is set, use `md` else use `html`.

```python
import random
import time

from dvclive import Live
from PIL import Image
from PIL import ImageDraw 

live = Live(report="md")

for i in range(2):
    live.log("foo", i + random.random())
    live.log("bar", i + random.random())

    img = Image.new("RGB", (50, 50), (255, 255, 255))
    draw = ImageDraw.Draw(img)
    draw.text((20, 20), f"{i  + random.random():.1f}", (0,0,0))
    live.log_image("img.png", img)
    live.next_step()
    time.sleep(5)
```

### DIFF
--- a/dvclive/live.py
+++ b/dvclive/live.py
@@ -15,7 +15,7 @@ from .error import (
     InvalidDataTypeError,
     InvalidPlotTypeError,
 )
-from .report import html_report
+from .report import make_report
 from .utils import env2bool, nested_update, open_file_in_browser
 
 logging.basicConfig()
@@ -30,23 +30,34 @@ class Live:
         self,
         path: Optional[str] = None,
         resume: bool = False,
-        report: Optional[str] = "html",
+        report: Optional[str] = "auto",
     ):
         self._path: Optional[str] = path
         self._resume: bool = resume or env2bool(env.DVCLIVE_RESUME)
-        self._report: str = report
-        self.html_path = None
+
+        if report == "auto":
+            if env2bool("CI"):
+                report = "md"
+            else:
+                report = "html"
+        else:
+            if report not in {None, "html", "md"}:
+                raise ValueError(
+                    "`report` can only be `None`, `auto`, `html` or `md`"
+                )
+
+        self._report: Optional[str] = report
+        self.report_path = ""
 
         self.init_from_env()
 
         if self._path is None:
             self._path = self.DEFAULT_DIR
 
-        if self.html_path is None:
-            self.html_path = os.path.join(self.dir, "report.html")
-
         if self._report is not None:
-            out = Path(self.html_path).resolve()
+            if not self.report_path:
+                self.report_path = os.path.join(self.dir, f"report.{report}")
+            out = Path(self.report_path).resolve()
             logger.info(f"Report path (if generated): {out}")
 
         self._step: Optional[int] = None
@@ -69,7 +80,7 @@ class Live:
                 Path(self.dir) / data_type.subfolder, ignore_errors=True
             )
 
-        for f in {self.summary_path, self.html_path}:
+        for f in {self.summary_path, self.report_path}:
             if os.path.exists(f):
                 os.remove(f)
 
@@ -90,8 +101,9 @@ class Live:
             if not env2bool(env.DVCLIVE_HTML, "0"):
                 env_config["_report"] = None
             else:
+                env_config["_report"] = "html"
                 path = str(env_config["_path"])
-                self.html_path = path + "_dvc_plots/index.html"
+                self.report_path = path + "_dvc_plots/index.html"
 
             for k, v in env_config.items():
                 if getattr(self, k) != v:
@@ -191,11 +203,13 @@ class Live:
             json.dump(summary_data, f, indent=4, cls=NumpyEncoder)
 
     def make_report(self):
-        if self._report == "html":
-            html_report(self.dir, self.summary_path, self.html_path)
+        if self._report is not None:
+            make_report(
+                self.dir, self.summary_path, self.report_path, self._report
+            )
 
-            if env2bool(env.DVCLIVE_OPEN):
-                open_file_in_browser(self.html_path)
+            if self._report == "html" and env2bool(env.DVCLIVE_OPEN):
+                open_file_in_browser(self.report_path)
 
     def make_checkpoint(self):
         if env2bool(env.DVC_CHECKPOINT):
--- a/dvclive/report.py
+++ b/dvclive/report.py
@@ -3,12 +3,13 @@ from pathlib import Path
 
 from dvc_render.html import render_html
 from dvc_render.image import ImageRenderer
+from dvc_render.markdown import render_markdown
 from dvc_render.table import TableRenderer
 from dvc_render.vega import VegaRenderer
 
 from dvclive.data import PLOTS, Image, Scalar
 from dvclive.data.plot import Plot
-from dvclive.utils import parse_tsv, to_base64_url
+from dvclive.utils import parse_tsv
 
 
 def get_scalar_renderers(scalars_folder):
@@ -18,21 +19,30 @@ def get_scalar_renderers(scalars_folder):
             data = parse_tsv(file)
             for row in data:
                 row["rev"] = "workspace"
-            rel = file.relative_to(scalars_folder).with_suffix("")
-            name = str(rel.as_posix())
-            properties = {"x": "step", "y": name}
+
+            y = file.relative_to(scalars_folder).with_suffix("")
+            y = y.as_posix()
+
+            name = file.relative_to(scalars_folder.parent).with_suffix("")
+            name = name.as_posix()
+            name = name.replace(scalars_folder.name, "static")
+
+            properties = {"x": "step", "y": y}
             renderers.append(VegaRenderer(data, name, **properties))
     return renderers
 
 
 def get_image_renderers(images_folder):
+    dvclive_path = images_folder.parent
     renderers = []
     for suffix in Image.suffixes:
-        for file in Path(images_folder).rglob(f"*{suffix}"):
+        all_images = Path(images_folder).rglob(f"*{suffix}")
+        for file in sorted(all_images):
+            src = str(file.relative_to(dvclive_path))
             name = str(file.relative_to(images_folder))
             data = [
                 {
-                    ImageRenderer.SRC_FIELD: to_base64_url(file),
+                    ImageRenderer.SRC_FIELD: src,
                     ImageRenderer.TITLE_FIELD: name,
                 }
             ]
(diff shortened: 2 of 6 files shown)


dberenbaum:
Does the markdown report get generated automatically on `live.next_step()`? Do we need to follow up on https://github.com/iterative/dvclive/issues/231 to clarify behavior in this scenario now that we have non-html reports?

dberenbaum:
PIL will fail when writing to paths where the intermediate dirs don't exist yet. This is causing problems especially when using the dvclive callbacks since they automatically use intermediate dirs. Traceback:

```
eback (most recent call last):
  File "/Users/dave/Code/lstm_seq2seq/train.py", line 96, in <module>
    model.fit(
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/keras/utils/traceback_utils.py", line 67, in error_handler
    raise e.with_traceback(filtered_tb) from None
  File "/Users/dave/Code/dvclive/dvclive/keras.py", line 49, in on_epoch_end
    self.dvclive.next_step()
  File "/Users/dave/Code/dvclive/dvclive/live.py", line 150, in next_step
    self.set_step(self.get_step() + 1)
  File "/Users/dave/Code/dvclive/dvclive/live.py", line 142, in set_step
    self.make_report()
  File "/Users/dave/Code/dvclive/dvclive/live.py", line 207, in make_report
    make_report(
  File "/Users/dave/Code/dvclive/dvclive/report.py", line 85, in make_report
    render_markdown(renderers, output_path)
  File "/Users/dave/Code/dvc-render/src/dvc_render/markdown.py", line 68, in render_markdown
    renderer.generate_markdown(output_path=output_path)
  File "/Users/dave/Code/dvc-render/src/dvc_render/vega.py", line 118, in generate_markdown
    plt.savefig(output_file)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/pyplot.py", line 979, in savefig
    res = fig.savefig(*args, **kwargs)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/figure.py", line 3046, in savefig
    self.canvas.print_figure(fname, **kwargs)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 2319, in print_figure
    result = print_method(
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/backend_bases.py", line 1648, in wrapper
    return func(*args, **kwargs)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/_api/deprecation.py", line 415, in wrapper
    return func(*inner_args, **inner_kwargs)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/backends/backend_agg.py", line 541, in print_png
    mpl.image.imsave(
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/matplotlib/image.py", line 1675, in imsave
    image.save(fname, **pil_kwargs)
  File "/Users/dave/miniforge3/envs/tensorflow/lib/python3.10/site-packages/PIL/Image.py", line 2297, in save
    fp = builtins.open(filename, "w+b")
FileNotFoundError: [Errno 2] No such file or directory: 'results/train/loss.png'
```

dberenbaum:
Can we reorganize the structure of the markdown output and images? Using the keras callback, I end up with a dvclive dir that looks like:

```bash
$ tree results # name of the dvclive dir
results
├── eval
│   ├── accuracy.png
│   └── loss.png
├── report.md
├── scalars
│   ├── eval
│   │   ├── accuracy.tsv
│   │   └── loss.tsv
│   └── train
│       ├── accuracy.tsv
│       └── loss.tsv
└── train
    ├── accuracy.png
    └── loss.png
```

Could all of the md-related info be in one dir?

Also, is there any way we can create some consistency between where html and md get written?

dvclive/live.py
```diff
@@ -191,11 +203,13 @@ def make_summary(self):
             json.dump(summary_data, f, indent=4, cls=NumpyEncoder)
 
     def make_report(self):
-        if self._report == "html":
-            html_report(self.dir, self.summary_path, self.html_path)
+        if self._report is not None:
```
daavoo, line 206: @dberenbaum 

> Does the markdown report get generated automatically on live.next_step()? 

Yes.

> Do we need to follow up on https://github.com/iterative/dvclive/issues/231 to clarify behavior in this scenario now that we have non-html reports?

Not sure I follow what's the difference of considering non-html report in the discussion

dvclive/report.py
```diff
@@ -17,21 +19,24 @@ def get_scalar_renderers(scalars_folder):
             data = parse_tsv(file)
             for row in data:
                 row["rev"] = "workspace"
-            rel = file.relative_to(scalars_folder).with_suffix("")
-            name = str(rel.as_posix())
-            properties = {"x": "step", "y": name}
-            renderers.append(VegaRenderer(data, name, **properties))
+            y = file.relative_to(scalars_folder).with_suffix("")
+            name = file.relative_to(scalars_folder.parent).with_suffix("")
```
daavoo, line 23: @dberenbaum 

> Can we reorganize the structure of the markdown output and images? Using the keras callback, I end up with a dvclive dir that looks like:
> 
> ```shell
> $ tree results # name of the dvclive dir
> results
> ├── eval
> │   ├── accuracy.png
> │   └── loss.png
> ├── report.md
> ├── scalars
> │   ├── eval
> │   │   ├── accuracy.tsv
> │   │   └── loss.tsv
> │   └── train
> │       ├── accuracy.tsv
> │       └── loss.tsv
> └── train
>     ├── accuracy.png
>     └── loss.png
> ```
> 
> Could all of the md-related info be in one dir?
 
 The `.png` images corresponding to each `.tsv` plot are now generated in the same path:
 
 ```
 results
├── report.md
└── scalars
   ├── eval
   │   ├── accuracy.png
   │   ├── accuracy.tsv
   │   ├── loss.png
   │   └── loss.tsv
   └── train
       ├── accuracy.png
       ├── accuracy.tsv
       ├── loss.png
       └── loss.tsv
```

File: dvclive/live.py [hunk 3]
```diff
@@ -30,23 +30,34 @@ def __init__(
...
         if self._path is None:
             self._path = self.DEFAULT_DIR
 
-        if self.html_path is None:
-            self.html_path = os.path.join(self.dir, "report.html")
-
         if self._report is not None:
-            out = Path(self.html_path).resolve()
+            if not self.report_path:
+                self.report_path = os.path.join(self.dir, f"report.{report}")
```
daavoo, line 59: @dberenbaum 

> Also, is there any way we can create some consistency between where html and md get written?

They are both written to `results/report.{format}`

daavoo:
> `ValueError: `generate_markdown`can only be used with`LinearTemplate``
> 
> `live.log_plot("confusion_matrix")` is not supported?

It is not. The initial scope was meant for `step` scenarios and `log_plot` can't be used in those. We could open an issue to follow up adding markdown support for confusion matrix template in https://github.com/iterative/dvc-render

dvclive/live.py
dberenbaum: 👍 I was probably looking at stale HTML reports.

dvclive/report.py
dberenbaum: WDYT about putting the images into a `static` dir? That would be more consistent with `dvc plots` and keep `scalars` uncluttered.

daavoo: Updated

pared (approved):

0x2b3bfa0:
> The new default value for `report` is `auto`.
> _If `CI` env var is set, use `md` else use `html`._

Isn't that a bit extreme? 🤔 I would prefer a less [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) behavior.

daavoo:
> Isn't that a bit extreme? 🤔 I would prefer a less [astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) behavior.

Users have the ability to select the value for `report` between `{None, html, md, auto}`. 

In `auto`, we make the assumption that `html` format is useful for local runs and `md` useful for CI. If we don't have something like that, we would require users to change the code between local and CI runs manually.

We could make it less extreme by not making `auto` the default, so users have to explicitly opt in. 

Alternatively, `report` could be configured via an ENV var (i.e. `DVCLIVE_REPORT`) that overrides the value in Python code (I like the `auto` mode better than this)

> Could the format be detected instead by using the file extension?

DVCLive doesn't know about the file CML is watching, it only knows the value passed to the `report` argument in the Python code.

dberenbaum:
> Alternatively, `report` could be configured via an ENV var (i.e. `DVCLIVE_REPORT`) that overrides the value in Python code (I like the `auto` mode better than this)
> 
> Thoughts ^ @dberenbaum

Agree with your comment @daavoo. We could add this environment variable (let's wait to see if it's needed), but it seems like reasonable default behavior if we document it. It's not unusual at least in ML libraries like sklearn to have "auto" options with conditional logic like this.

0x2b3bfa0:
> _it only knows the value passed to the report argument in the Python code_

Wouldn't that be enough to infer the format? E.g. if `report` ends with `.md` ...

0x2b3bfa0:
> _We could make it less extreme by not making auto the default, so users have to explicitly opt in._

If users are going to copy/paste a snippet anyway from the documentation, requiring an explicit `format="md"` doesn't seem that bad to me. Anyway, just a nitpick. 💅🏼

dvclive/live.py
```diff
@@ -15,7 +15,7 @@
     InvalidDataTypeError,
     InvalidPlotTypeError,
 )
-from .report import html_report
+from .report import make_report
 from .utils import env2bool, nested_update, open_file_in_browser
```
jorgeorpinel, line 19: Is it OK to deprecate an attribute just like that? Should it stay hidden for backward compatibility instead?

dberenbaum: @daavoo WDYT? Since we are still pre-1.0 (we probably need to consider a 1.0 release), we can technically break anything. Not sure how you feel about keeping an alias for backward compatibility.

daavoo: In this case, i feel its not necesary. The public and documented method is/was `Live.make_report` which didn't change. This was an internal función.

Regarding 1.0 I added items to https://github.com/iterative/dvclive/issues/223

PR #292, opened 2022-09-12
add log_param/log_params

dtrifiro:
Provides two new methods which log parameters to a yaml file (`params.yaml`):

- `log_param(name, value)` logs the given parameter name/value
- `log_params(params)` logs the given set of parameters (dict).

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -7,20 +7,31 @@ from itertools import chain
 from pathlib import Path
 from typing import Any, Dict, Optional, Union
 
+from ruamel.yaml.representer import RepresenterError
+
 from . import env
 from .data import DATA_TYPES, PLOTS, Image, NumpyEncoder, Scalar
 from .dvc import make_checkpoint
 from .error import (
     ConfigMismatchError,
     InvalidDataTypeError,
+    InvalidParameterTypeError,
     InvalidPlotTypeError,
+    ParameterAlreadyLoggedError,
 )
 from .report import make_report
+from .serialize import dump_yaml, load_yaml
 from .utils import env2bool, nested_update, open_file_in_browser
 
 logging.basicConfig()
 logger = logging.getLogger("dvclive")
 logger.setLevel(os.getenv(env.DVCLIVE_LOGLEVEL, "INFO").upper())
+
+
+# Recursive type aliases are not yet supported by mypy (as of 0.971),
+# so we set type: ignore for ParamLike.
+#  See https://github.com/python/mypy/issues/731#issuecomment-1213482527
+ParamLike = Union[int, float, str, bool, Dict[str, "ParamLike"]]  # type: ignore # noqa
 
 
 class Live:
@@ -54,6 +65,8 @@ class Live:
         if self._path is None:
             self._path = self.DEFAULT_DIR
 
+        self._params_path = os.path.join(self._path, "params.yaml")
+
         if self._report is not None:
             if not self.report_path:
                 self.report_path = os.path.join(self.dir, f"report.{report}")
@@ -64,15 +77,17 @@ class Live:
         self._scalars: Dict[str, Any] = OrderedDict()
         self._images: Dict[str, Any] = OrderedDict()
         self._plots: Dict[str, Any] = OrderedDict()
+        self._params: Dict[str, Any] = OrderedDict()
 
+        self._init_paths()
         if self._resume:
+            self._read_params()
             self._step = self.read_step()
             if self._step != 0:
                 self._step += 1
             logger.info(f"Resumed from step {self._step}")
         else:
             self._cleanup()
-            self._init_paths()
 
     def _cleanup(self):
         for data_type in DATA_TYPES:
@@ -80,7 +95,7 @@ class Live:
                 Path(self.dir) / data_type.subfolder, ignore_errors=True
             )
 
-        for f in {self.summary_path, self.report_path}:
+        for f in (self.summary_path, self.report_path, self.params_path):
             if os.path.exists(f):
                 os.remove(f)
 
@@ -117,6 +132,10 @@ class Live:
         return self._path
 
     @property
+    def params_path(self):
+        return self._params_path
+
+    @property
     def exists(self):
         return os.path.isdir(self.dir)
 
@@ -130,7 +149,6 @@ class Live:
     def set_step(self, step: int) -> None:
         if self._step is None:
             self._step = 0
-            self._init_paths()
             for data in chain(
                 self._scalars.values(),
                 self._images.values(),
@@ -190,6 +208,43 @@ class Live:
 
         data.dump(val, self._step, **kwargs)
         logger.debug(f"Logged {name}")
+
+    def _read_params(self):
+        if os.path.isfile(self.params_path):
+            params = load_yaml(self.params_path)
+            self._params.update(params)
+
+    def _dump_params(self):
+        try:
+            dump_yaml(self.params_path, self._params)
+        except RepresenterError as exc:
+            raise InvalidParameterTypeError(exc.args) from exc
+
+    def log_params(self, params: Dict[str, ParamLike]):
+        """Saves the given set of parameters (dict) to yaml"""
+        if self._resume and self.get_step():
+            logger.info(
+                "Resuming previous dvclive session, not logging params."
+            )
+            return
+
+        for param_name, param_value in params.items():
+            if param_name in self._params:
+                raise ParameterAlreadyLoggedError(
+                    param_name, param_value, self._params[param_name]
+                )
+
+        self._params.update(params)
+        self._dump_params()
+        logger.debug(f"Logged {params} parameters to {self.params_path}")
+
+    def log_param(
+        self,
+        name: str,
+        val: ParamLike,
+    ):
+        """Saves the given parameter value to yaml"""
+        self.log_params({name: val})
 
     def make_summary(self):
         summary_data = {}
--- a/src/dvclive/serialize.py
+++ b/src/dvclive/serialize.py
@@ -1,0 +1,42 @@
+from collections import OrderedDict
+
+from dvclive.error import DvcLiveError
+
+
+class YAMLError(DvcLiveError):
+    pass
+
+
+class YAMLFileCorruptedError(YAMLError):
+    def __init__(self, path):
+        super().__init__(path, "YAML file structure is corrupted")
+
+
+def load_yaml(path, typ="safe"):
+    from ruamel.yaml import YAML
+    from ruamel.yaml import YAMLError as _YAMLError
+
+    yaml = YAML(typ=typ)
+    with open(path, encoding="utf-8") as fd:
+        try:
+            return yaml.load(fd.read())
+        except _YAMLError:
+            raise YAMLFileCorruptedError(path)
+
+
+def _get_yaml():
+    from ruamel.yaml import YAML
+
+    yaml = YAML()
+    yaml.default_flow_style = False
+
+    # tell Dumper to represent OrderedDict as normal dict
+    yaml_repr_cls = yaml.Representer
+    yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
+    return yaml
+
+
+def dump_yaml(path, data):
+    yaml = _get_yaml()
+    with open(path, "w", encoding="utf-8") as fd:
+        yaml.dump(data, fd)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,7 @@ addopts = "-ra"
 
 [tool.coverage.run]
 branch = true
-source = ["dvclive", "tests"]
+source = ["src/dvclive", "tests"]
 
 [tool.coverage.paths]
 source = ["src", "*/site-packages"]
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ package_dir=
 packages = find:
 install_requires=
     dvc_render[table]>=0.0.8
+    ruamel.yaml>=0.17.11
 
 [options.extras_require]
 tests =
--- a/src/dvclive/error.py
+++ b/src/dvclive/error.py
@@ -1,5 +1,5 @@
 import os
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Optional
 
 if TYPE_CHECKING:
     from .live import Live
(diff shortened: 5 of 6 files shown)


shcheklein:
Should we log into `params.yaml` ? To simplify the migration and so that we can show them as params in Studio/VS Code?

Otherwise, what is the plan to support this in both - Studio/VS Code?

src/dvclive/live.py
```diff
@@ -79,6 +85,7 @@ def _cleanup(self):
             shutil.rmtree(
                 Path(self.dir) / data_type.subfolder, ignore_errors=True
             )
+        shutil.rmtree(Path(self.dir) / self.PARAMS_DIR, ignore_errors=True)
```
daavoo, line 88: Why not handle whether or not to remove the dir base on `self._resume` here instead of inside the `log_param` method?

src/dvclive/live.py
```diff
@@ -191,6 +198,78 @@ def log_plot(self, name, labels, predictions, **kwargs):
...
+        """Saves the given set of parameters (dict) to json"""
+        if self._resume:
+            logger.debug("resume=True, not removing existing parameters.")
+            return
+
+        output_dir = Path(self.dir) / self.PARAMS_DIR
+        output_dir.mkdir(exist_ok=True)
+        params_path = output_dir / (
+            f"{name}.json" if not name.endswith(".json") else name
+        )
```
daavoo, line 213: (Implementation)
Not a blocker but, for consistency, I think would be nice if we implement a new `Data` subclass (i.e. `Param`) and move this logic there.

It would look more similar to the other `log_x` methods and could launch explicit exceptions for invalid data types, etc.

src/dvclive/live.py
```diff
@@ -191,6 +198,78 @@ def log_plot(self, name, labels, predictions, **kwargs):
         data.dump(val, self._step, **kwargs)
         logger.debug(f"Logged {name}")
 
+    def log_params(
+        self, params: Dict[str, ParamsLike], name: str = PARAMS_DEFAULT_NAME
+    ):
```
daavoo, line 203: (Product) I am not sure about including this one (at least until we also include equivalents for other data types). 

(Implementation) Regardless, is there a reason for this method to don't reuse `log_param` internally?

src/dvclive/live.py
```diff
@@ -191,6 +198,78 @@ def log_plot(self, name, labels, predictions, **kwargs):
...
+
+        output_dir = Path(self.dir) / self.PARAMS_DIR
+        output_dir.mkdir(exist_ok=True)
+        params_path = output_dir / (
+            f"{name}.json" if not name.endswith(".json") else name
+        )
+
+        if params_path.is_file():
+            with open(params_path, "r", encoding="utf-8") as fobj:
+                loaded_params = json.load(fobj)
```
daavoo, line 217: (Product) Should we load the existing params? I think we could just treat as the other things we log and overwrite unless `resume` is used

src/dvclive/live.py
```diff
@@ -191,6 +198,78 @@ def log_plot(self, name, labels, predictions, **kwargs):
...
+
+        logger.debug(
+            f"Logged {len(params.keys())} parameters to {params_path}"
+        )
+
+    def log_param(
+        self,
+        name: str,
+        val: Union[int, float, str, bool],
+        params_set_name: str = PARAMS_DEFAULT_NAME,
```
daavoo, line 239: (Product) I am not sure about providing this option. 

I believe we could start by "forcing" a fixed output file of our choice so that the Onboarding workflow gets easier to implement.

We can wait for users to ask for this unless we see a good reason to include the logic as part of the current onboarding proposal (right now feels a little like a deviation from what we have been discussing)

src/dvclive/live.py
```diff
@@ -22,9 +23,13 @@
 logger = logging.getLogger("dvclive")
 logger.setLevel(os.getenv(env.DVCLIVE_LOGLEVEL, "INFO").upper())
 
+ParamsLike = Union[int, float, str, bool]
```
daavoo, line 26: ❤️ . Would be awesome to also add this to the other methods `log`, `log_plot` `log_image` (not as part of this P.R)

src/dvclive/live.py
dberenbaum: I think it's okay to start here because we have to start somewhere and params are maybe the most likely scenario for having to log many values at once.

However, can we collapse into one method that accepts either? It's messier to implement but less for users to remember.

src/dvclive/live.py
```diff
@@ -191,6 +198,78 @@ def log_plot(self, name, labels, predictions, **kwargs):
         data.dump(val, self._step, **kwargs)
         logger.debug(f"Logged {name}")
 
+    def log_params(
+        self, params: Dict[str, ParamsLike], name: str = PARAMS_DEFAULT_NAME
```
dberenbaum, line 202: What about handling nested dicts? Similarly, can `log_param("train.lr", 0.1)` be saved as a nested dict?

dberenbaum:
> Should we log into `params.yaml` ? To simplify the migration and so that we can show them as params in Studio/VS Code?

I'm not sure if it simplifies migration to put them into `params.yaml`, since it then requires understanding that you need to remove your `log_param()` calls or end up overwriting your param dependencies. Saving them in some other YAML file might be less confusing, although it looks a bit odd to be saving data in both YAML and JSON (no strong opinion there). 

> Otherwise, what is the plan to support this in both - Studio/VS Code?

This is still under discussion, but a custom params file could be supported through `dvc.yaml` config or through some DVCLive convention.

src/dvclive/live.py
dtrifiro: We should be handling those, although I think this can become tricky: think of more complex objects, such as an optimizer (like in here https://keras.io/api/optimizers/). How are we supposed to capture these kind of parameters?

dtrifiro:
> I'm not sure if it simplifies migration to put them into params.yaml, since it then requires understanding that you need to remove your log_param() calls or end up overwriting your param dependencies.

+1 

> Saving them in some other YAML file might be less confusing, although it looks a bit odd to be saving data in both YAML and JSON (no strong opinion there).

One reason I see for preferring YAML is to provide users with an example on how a "standard" `params.yaml` file would look like, although this does not provide a clear path on how to go forward from there: "ok, I have a `params.yaml` file which I could use as a params dependency, although I still have to rewrite my code in order to read/use that file"

src/dvclive/live.py
dtrifiro: > (Implementation) Is there a reason for this method to don't reuse log_param internally?

These were slighty different initially, but the ended up converging, I'll change `log_param` so that it uses `log_params` internally.

src/dvclive/live.py
daavoo: I would expect people to log that kind of thing as follows:

`Live().log_params("train", {"lr": 0.01})`

Which works, although I don't know if unexpectedly 😅 as it doesn't match the `ParamsLike` type @dtrifiro

src/dvclive/live.py
dtrifiro: > However, can we collapse into one method that accepts either? It's messier to implement but less for users to remember.

Wouldn't it be confusing? As in: with a single method, if logging a `dict`, are you logging a single parameter that takes a `dict` value (this is somehow related to your nested params comment) or are you logging a group of parameters? Maybe it would be just enough to provide an optional `name` argument to distinguish between the two cases, although I stil feel naming would be odd.

Something like this, perhaps: 
```python
dvclive.log_param(name="optimizer_kwargs", {"learning_rate": 1e-3, "name": "adam"}) # this logs a dict parameter
dvclive.log_param({"learning_rate": 1e-3, "optimizer_name": "adam"}) # this logs a dict parameter
```

src/dvclive/live.py
dtrifiro: I thought about that, and that's the first approach I went for, although if you look at the `Data` class most of its (abstract) methods are tightly connected to the "stepped" workflow:  `step`, `no_step_output_path`, `first_step_dump`, `no_step_dump`, `step_dump`. Maybe have a new superclass from which `Data` and `Param` derive, although I think they would share only few methods/attributes.

I think a `Param` class is something we will have to add, especially in relation to the nested dictionaries/complex objects parameters support.

src/dvclive/live.py
dtrifiro: This depends on how we want to handle `log_param`/`log_params`. If using `log_param(name, val)` I would expect this to append the parameter to an existing file. Using `log_params` could avoid loading the parameters, although it could be confusing to have different behaviors. 

Note that the params directory is always cleared on dvclive's init and we are not dumping parameters when `resume==True`

src/dvclive/live.py
dtrifiro: That's why this is optional ("forcing" a fixed output). The idea is to let users log parameters to different files to keep a flatter structure in each file. Could get rid of this though.

src/dvclive/live.py
dtrifiro: Yep, `ParamsLike` should really look like this: `ParamsLike = Union[int, float, str, bool, "ParamsLike"]`, although mypy complains because of a cyclic definition. Will check this out later.

dberenbaum: Sorry, for some reason I thought the dot-notation was supported for logging metrics. That shouldn't be necessary, but handling nested dicts should be.

Some interesting reading:
* https://stackoverflow.com/questions/68167770/type-hints-for-variably-nested-dict-in-python
* https://github.com/python/mypy/issues/731

src/dvclive/live.py
dtrifiro: Not sure what you mean here

src/dvclive/live.py
dtrifiro: Any thoughts here?

src/dvclive/live.py
dtrifiro: Changed this slightly, now skipping `log_param/log_params` when `resume==True an step > 0`

src/dvclive/live.py
dtrifiro: Fixed `ParamsLike` to include the recursive definition. For now we support dumping any parameters that are supported by `json.dump`. Complex (custom) parameters are still not supported. Do we want to support those right away, or can we leave those to another iteration?

dtrifiro:
Note: linting failures on mac/win 3.10 seem unrelated

src/dvclive/live.py
dberenbaum: LGTM with `log_param` and `log_params`, thanks!

src/dvclive/live.py
dberenbaum: Seems like it deviates from the rest of dvclive where the output structure is fixed relative to the dvclive path, and it might be confusing to document, so my preference is to leave this out until we have a need.

src/dvclive/live.py
dberenbaum: Still seems a little odd because of how similar they both are now, but it's probably the most familiar approach for both dvclive and mlflow users, and it gives the most flexibility to users, so I'm good with it.

src/dvclive/live.py
dtrifiro: Got rid of it

dberenbaum:
> Should we log into `params.yaml` ?

@dtrifiro Do you want to migrate to a YAML file? I actually don't think `params.yaml` is that problematic as long as it's inside the dvclive dir.

src/dvclive/live.py
```diff
@@ -23,8 +24,16 @@
...
 
+# Recursive type aliases are not yet supported by mypy (as of 0.971),
+# so we set type: ignore for ParamLike.
+#  See https://github.com/python/mypy/issues/731#issuecomment-1213482527
+ParamLike = Union[int, float, str, bool, Dict[str, "ParamLike"]]  # type: ignore # noqa
+
+
 class Live:
     DEFAULT_DIR = "dvclive"
+    PARAMS_DIR = "params"
```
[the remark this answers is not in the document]
dtrifiro: Not if we do not want multiple parameters files, will remove

dberenbaum:
Discussed with @dtrifiro and decided that we should be able to implement basic YAML parsing without copying all of the DVC YAML serialization utilities.

src/dvclive/live.py
```diff
@@ -79,6 +93,8 @@ def _cleanup(self):
             shutil.rmtree(
                 Path(self.dir) / data_type.subfolder, ignore_errors=True
             )
+        if os.path.exists(self.params_path):
+            os.remove(self.params_path)
 
         for f in {self.summary_path, self.report_path}:
             if os.path.exists(f):
```
daavoo, line 100:
```suggestion
        for f in {self.summary_path, self.report_path, self.params_path}:
            if os.path.exists(f):
```

src/dvclive/live.py
```diff
@@ -191,6 +211,43 @@ def log_plot(self, name, labels, predictions, **kwargs):
         data.dump(val, self._step, **kwargs)
         logger.debug(f"Logged {name}")
 
+    def log_params(self, params: Dict[str, ParamLike]):
+        """Saves the given set of parameters (dict) to json"""
```
daavoo, line 215:
```suggestion
        """Saves the given set of parameters (dict) to `self.params_path`"""
```

src/dvclive/live.py
```diff
@@ -191,6 +211,43 @@ def log_plot(self, name, labels, predictions, **kwargs):
...
+        if self._resume and self._step:
+            logger.info(
+                "Resuming previous dvclive session, not logging params."
+            )
+            return
```
daavoo, line 220: Missing test for this

src/dvclive/serialize.py
```diff
@@ -0,0 +1,47 @@
...
+
+    # tell Dumper to represent OrderedDict as normal dict
+    yaml_repr_cls = yaml.Representer
+    yaml_repr_cls.add_representer(OrderedDict, yaml_repr_cls.represent_dict)
+    return yaml
+
+
+def _dump(data, stream):
+    yaml = _get_yaml()
+    return yaml.dump(data, stream)
```
daavoo, line 41: Unused?
```suggestion
```

daavoo (changes requested):
Some minor comments and good to go.

Main doubt:

(Product) I am missing the point of `YAML` vs `JSON` but I have no strong opinion. I just think the right direction would be to implement https://github.com/iterative/dvc/issues/7939 and save us the new dependency / extra utils code.

Must address:

(Implementation) @dtrifiro Need to either make all imports of `ruamel.yaml` optional with `try/except ImporError` or add as a `install_requires` dependency.

dberenbaum:
> (Product) I am missing the point of `YAML` vs `JSON` but I have no strong opinion. I just think the right direction would be to implement [iterative/dvc#7939](https://github.com/iterative/dvc/issues/7939) and save us the new dependency / extra utils code.

@daavoo Not sure I understand your point (I also have no strong opinion on `YAML` vs `JSON`). How does a configurable params file save code/dependencies here?

daavoo:
> @daavoo Not sure I understand your point (I also have no strong opinion on `YAML` vs `JSON`). How does a configurable params file save code/dependencies here?

If I understood the motivation for `params.yaml` was to `simplify the migration`, right? 
If we allow DVC to set custom default params file, migration would be equally simple with `json` format

dtrifiro:
Addressed the remaining comments. 

> If we allow DVC to set custom default params file
Do we really need this? We can already specify custom params files, although we "nudge" users towards using `params.yaml`

dberenbaum:
> If I understood the motivation for `params.yaml` was to `simplify the migration`, right?
> If we allow DVC to set custom default params file, migration would be equally simple with `json` format

Good point. I would say YAML is arguably more easily readable and editable and maybe more common for parameters, but JSON is not so different, has fewer issues, and is already used for metrics, so 🤷 . Let's merge this and we can come back to it if needed in the future.

daavoo (approved):

PR #297, opened 2022-09-14
Add `post_to_studio`

daavoo:
For MVP  https://github.com/iterative/studio/pull/4311


tests/test_studio.py
```diff
@@ -0,0 +1,109 @@
...
+            "rev": mocker.ANY,
+            "step": 1,
+            "timestamp": mocker.ANY,
+            "metrics": {"dvclive.json": {"data": {"step": 1, "foo": 2}}},
+            "plots": {
+                "dvclive/scalars/foo.tsv": {
+                    "data": [
+                        {"timestamp": mocker.ANY, "step": "0", "foo": "1"},
+                        {"timestamp": mocker.ANY, "step": "1", "foo": "2"},
+                    ]
```
daavoo, line 103: @Suor Is this what Studio expects when a previous request has failed? I just include the datapoints that were not sent

src/dvclive/studio.py
```diff
@@ -0,0 +1,51 @@
...
+    import requests
+
+    timestamp = datetime.now().isoformat()
+    if type == "data":
+        metrics, plots = _get_updates(live)
+    else:
+        metrics = plots = {}
+
+    return requests.post(
+        "https://0.0.0.0",
```
daavoo, line 39: @Suor What would be the endpoint?

tests/test_studio.py
Suor: Is `step` guaranteed to be there? Can I rely on that if I saved some plot row with `step == X` then if I receive it again it will be the same thing resend, so that I may ignore it safely? Why are those strings? May I convert them to integer for purposes of order? Is it the same step as in root of the event - which is needed to filter out resending if metrics too?

src/dvclive/studio.py
Suor: Most probably `https://studio.iterative.ai/api/live` for production. Staging `https://dev.iterative.ai/api/live`, local `http://localhost:8111/api/live`.

tests/test_studio.py
daavoo: > Is step guaranteed to be there?

Yes

> Can I rely on that if I saved some plot row with step == X then if I receive it again it will be the same thing resent to me, so that I may ignore it safely?

Yes. But I will only send older steps if Studio doesn't return a success code.

> Why are those strings? May I convert them to integer for purposes of order?

Just that the `csv` reader returns strings by default. I can cast to integer before posting.

> Is it the same step as in root of the event - which is needed to filter out resending if metrics too?

The highest step number in `plots` will always match the `step` in the top-level request. 

I am not currently resending metrics from failed requests, just the latest ones. Should I also send metrics from previous steps?

Suor:
> I am not currently resending metrics from failed requests, just the latest ones. Should I also send metrics from previous steps?

No need to resend old metrics

src/dvclive/live.py
```diff
@@ -87,6 +95,23 @@ def __init__(
         else:
             self._cleanup()
 
+        self._latest_studio_step = self.get_step()
+
+        if self._report == "studio":
+            from scmrepo.git import Git
```
dtrifiro, line 101: Do we need to add scmrepo to setup.cfg as an extra?

src/dvclive/studio.py
```diff
@@ -0,0 +1,72 @@
...
+        response = requests.post(
+            getenv(STUDIO_ENDPOINT, "https://studio.iterative.ai/api/live"),
+            json=data,
+            headers={
+                "Content-type": "application/json",
+                "Authorization": f"token {live.studio_token}",
+            },
+            timeout=5,
+        )
+    except Timeout:
```
dtrifiro, line 70: Might be worth to also catch  `requests.exceptions.ConnectionError`

dtrifiro (approved):

src/dvclive/live.py
```diff
@@ -155,7 +180,20 @@ def set_step(self, step: int) -> None:
...
+            logger.debug("Post to Studio `data` event")
+            if not response or response.status_code != 200:
+                logger.warning(
+                    "`post_to_studio` `data` event failed."
+                    "Data will be resent on next call."
+                )
+                if response:
+                    logger.debug(response.content.decode())
+            else:
+                self._latest_studio_step = step
```
dtrifiro, line 194:
```suggestion
        if self._report == "studio":
            response = post_to_studio(self, "data")
            logger.debug("Post to Studio `data` event")
            if response:
                self._latest_studio_step = step
            else:   
                logger.warning(
                    "`post_to_studio` `data` event failed."
                    "Data will be resent on next call."
                )
                if message := response.content.decode():
                    logger.debug(message)
```

dtrifiro: Although the conditional for `logger.debug(message)` might be a bit odd in some cases, might be worth

```python
message = response.content.decode()
logger.debug(f"post_to_studio: {response.status_code=} " + f", {message=}" if message else "")
```

src/dvclive/studio.py
daavoo: Used the base class `RequestException`

src/dvclive/live.py
daavoo: Would probably be the right thing to do, although I don't think it matters because the feature can only work if `dvc` is installed.

src/dvclive/live.py
daavoo: Moved the `debug` messages to inside `post_to_studio` and left only the `warning` in `Live` class, please TAL

src/dvclive/studio.py
```diff
@@ -0,0 +1,82 @@
...
+    for name, plot in plots.items():
+        datapoints = _get_unsent_datapoints(plot, latest_step)
+        plots[name] = _cast_to_numbers(datapoints)
+
+    metrics = {live.summary_path: {"data": metrics}}
+    plots = _to_dvc_format(plots)
+    return metrics, plots
+
+
+def post_to_studio(live, event_type, logger) -> bool:
```
dtrifiro, line 43: How about using enums for event types?

src/dvclive/live.py
```diff
@@ -260,10 +291,14 @@ def make_report(self):
...
                 self.dir, self.summary_path, self.report_path, self._report
             )
-
             if self._report == "html" and env2bool(env.DVCLIVE_OPEN):
                 open_file_in_browser(self.report_path)
 
+    def end(self):
+        if self._report == "studio":
+            if not post_to_studio(self, "done"):
+                logger.warning("`post_to_studio` `done` event failed.")
```
dtrifiro, line 300:
```suggestion
        if self._report == "studio" not post_to_studio(self, "done"):
            logger.warning("`post_to_studio` `done` event failed.")
```

dtrifiro (approved):

Issue #289, opened 2021-08-31
dvclive: Add a note _not_ to use `dvclive.init` when using keras Callback


iesahin:
I think it's better to notify the user that `dvclive.init` is called inside the callback in https://dvc.org/doc/dvclive/user-guide/ml-frameworks/keras 

They also don't need `import dvclive`. 

@daavoo @dberenbaum

Issue #207, opened 2022-01-14
integrations: Add `live` kwarg to callbacks


dberenbaum:
Instead of assuming a new `Live` instance in callbacks, should it allow for `DvcLiveCallback(live=myliveinstance)`? This could better support the workflow mentioned in https://github.com/iterative/dvc.org/issues/2773.

For example:

```
from dvclive import Live
from dvclive.keras import DvcLiveCallback

live = Live("custom_path")
model.fit(callbacks=[DvcLiveCallback(live=live)])
```

daavoo:
> This could better support the workflow mentioned in iterative/dvc.org#2773.

Not sure what workflow do you refer to 😅 

What would be the benefits of that approach? or What are the limitations of the current one?

dberenbaum:
It's not high priority, and it's not intending to replace the current approach.

In iterative/dvc.org#2773, it was unclear whether to instantiate the `Live` instance before using the callback.

For example:

```
live = Live("custom_path")
model.fit(callbacks=[DvcLiveCallback()])
```

The outputs will be in `dvclive` instead of 'custom_path`, which might be unexpected. If they want to get the expected behavior, they will have to rewrite their code like:

```
model.fit(callbacks=[DvcLiveCallback(path="custom_path")])
```

An alternative could be to also allow:

```
live = Live("custom_path")
model.fit(callbacks=[DvcLiveCallback(live=live)])
```

If both `live` and other kwargs like `path` are provided, an error could be thrown. It just gives users another way to fix the problem in case it's not obvious to drop the instantiation.

I will close it for now. If anyone encounters confusion, we can revisit.

PR #302, opened 2022-09-25
keras: add ability to pass live into callback

shcheklein:
Suggestion. If we decide it's good I can try to add it to all other integrations.

It makes this workflow (non obvious, not natural, less scalable, and not documented?): 

```python
logger = DvcLiveCallback(path="evaluation", report=None)
live = logger.dvclive
```

into:

```python
live = Live("evaluation", report=None)

# ... somewhere down the road:

.... DvcLiveCallback(dvclive = live)
```

- We need an ability to pass live logger instance or params, otherwise callback would initialize its own instance and will be producing reports for example, would ignore path, etc, etc
- It's better to pass live to callbacks (vs asking callback to give us logger): proper order of control (code looks natural, easier to test), one `live` instance could be shared across multiple callbacks, etc

### DIFF
--- a/src/dvclive/keras.py
+++ b/src/dvclive/keras.py
@@ -1,4 +1,5 @@
 import os
+from typing import Optional
 
 from tensorflow.keras.callbacks import (  # noqa pylint: disable=import-error, no-name-in-module
     Callback,
@@ -13,12 +14,16 @@ from dvclive.utils import standardize_metric_name
 
 class DvcLiveCallback(Callback):
     def __init__(
-        self, model_file=None, save_weights_only: bool = False, **kwargs
+        self,
+        model_file=None,
+        save_weights_only: bool = False,
+        dvclive: Optional[Live] = None,
+        **kwargs
     ):
         super().__init__()
         self.model_file = model_file
         self.save_weights_only = save_weights_only
-        self.dvclive = Live(**kwargs)
+        self.dvclive = dvclive if dvclive is not None else Live(**kwargs)
 
     def on_train_begin(self, logs=None):  # pylint: disable=unused-argument
         if (
--- a/tests/test_keras.py
+++ b/tests/test_keras.py
@@ -2,6 +2,7 @@ import os
 
 import pytest
 
+from dvclive import Live
 from dvclive.data.scalar import Scalar
 from dvclive.keras import DvcLiveCallback
 from tests.test_main import read_logs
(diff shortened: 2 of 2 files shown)


src/dvclive/keras.py
```diff
@@ -13,12 +14,16 @@
 
 class DvcLiveCallback(Callback):
     def __init__(
-        self, model_file=None, save_weights_only: bool = False, **kwargs
+        self,
+        model_file=None,
+        save_weights_only: bool = False,
+        dvclive: Optional[Live] = None,
```
pared, line 20: Should it be `dvclive` or `live`?

pared (commented):
Added test, one question.

src/dvclive/keras.py
shcheklein: I was using this name since in the __init__ itself it's using `dvclive`

shcheklein: I don't have any strong opinion on this tbh

daavoo:
@shcheklein what's the use case here? 

I remember discussing this a long time ago and I think the conclusion was that it makes more sense to have a single instance (and output path) associated with the train loop. 

The primary motivation was organization-wise (separate inside/outside train loop in folders) and just overall efficiency (having a out of train loop evaluation in a separate stage avoids to recompute training when changing evaluation stuff)

dberenbaum:
@daavoo Maybe you are thinking of https://github.com/iterative/dvclive/issues/207?

shcheklein:
@daavoo use case is when you want to log certain things besides what's being tracked inside one of the callbacks, so you need to have access to the `live` instance.

This is the way it can be solved:

https://github.com/shcheklein/hackathon/blob/universe/train.py#L55-L56

But I think this is wrong to do it that way (wrong control order). I've outlined a few reasons for that above.

shcheklein:
Alternatively maybe `live` instance can be global (similar to Python logger?) and callbacks can rely on that instance?

shcheklein:
@daavoo @dberenbaum yes, #207 and #289 are the same / very related. I hit this bump independently.

dberenbaum:
> Alternatively maybe `live` instance can be global (similar to Python logger?) and callbacks can rely on that instance?

I think it used to be global but we moved away from it in case users want multiple live instances (for example, they train multiple models). IMO optionally passing live into the callback (as this PR does) is the most flexible since you can choose whether you want to keep everything in one instance or not.

daavoo (approved):

daavoo:
> @daavoo use case is when you want to log certain things besides what's being tracked inside one of the callbacks, so you need to have access to the live instance.
> 
> This is the way it can be solved:
> 
> https://github.com/shcheklein/hackathon/blob/universe/train.py#L55-L56
> 
> But I think this is wrong to do it that way (wrong control order). I've outlined a few reasons for that above.

I understand the low-level motivation, my question was about the high-level use case, which in the linked example appears to be something like: "log the best value of each metric so that those are the values shown in the table".

I am ok with the P.R. but my feeling still is that the use cases for this are either workarounds of existing dvclive limitations or cases where would be a better practice to have a separate stage and/or output folder.

shcheklein:
> but my feeling still is that the use cases for this are either workarounds of existing dvclive limitations or cases where would be a better practice to have a separate stage and/or output folder.

Yes, I think I understand your feelings. My intuition though that this kind of extensibility / flexibility will be still needed (it will be very hard to fit / cover everything with callbacks / higher level APIs). Also, tbh, I don't see that much harm exposing this. We can keep learning from the way people using it and migrate scenarios inside.

I'll try to expand to all other frameworks / add docs if needed.

pared:
I think that this approach might be useful if we don't want to log everything on each step:
let's say that we want a callback to log training metrics, but only want to dump something "big" (confusion matrix for example) at the end of the training.

PR #308, opened 2022-09-27
bump mypy to 0.981, enable recursive type aliases

dtrifiro:


dtrifiro:
Not sure what's going on, but this is the error (linting session):
```console
nox > safety check --full-report
Unhandled exception happened: 'catalyst'
```

daavoo (commented):
I believe we can also remove the 3.10.6 pin on CI as the mypy release is supposed to fix the issue

dtrifiro:
@daavoo the safety session is failing because of a catalyst[dev] dependency (sphinx) with a security vulnerability. I see that catalyst is pinned: `catalyst<=21.12`, do you know why?

daavoo:
> @daavoo the safety session is failing because of a catalyst[dev] dependency (sphinx) with a security vulnerability 🤦‍♀️. I see that catalyst is pinned: `catalyst<=21.12`, do you know why?

I believe because of this https://github.com/iterative/dvclive/issues/219

daavoo:
> @daavoo the safety session is failing because of a catalyst[dev] dependency (sphinx) with a security vulnerability 🤦‍♀️. I see that catalyst is pinned: `catalyst<=21.12`, do you know why?

Vulnerability appears to also affect the latest version so the pin was not really the problem (anyway it is good to update)

PR #321, opened 2022-10-03
tests: Simplify huggingface tests.

daavoo:
Use simple dataset and model from `transformers/tests/trainer`.

### DIFF
--- a/src/dvclive/huggingface.py
+++ b/src/dvclive/huggingface.py
@@ -41,5 +41,6 @@ class DvcLiveCallback(TrainerCallback):
         if self.model_file:
             model = kwargs["model"]
             model.save_pretrained(self.model_file)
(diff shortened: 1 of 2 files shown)


pared (commented):
Seems like windows has some problems with new model.

daavoo:
> Seems like windows has some problems with new model.

I have no idea whats going on 😅

skshetry:
`Trainer` is using `time.time()` which has a very low resolution in Windows. So I think `time.time() - start_time` becomes `0`. Try adding a `time.sleep()` to the minimum resolution.
```python
time.sleep(time.get_clock_info('time').resolution)
```

skshetry:
Or, propose a fix in `transformers` to move away from `time.time()`. 😅

Issue #311, opened 2022-09-27
Configure DVC experiment from DVCLive


dberenbaum:
Follow up to #310

> Before creating the experiment ref, DVCLive should write to a `dvc.yaml` file inside the `Live` directory, which should configure parameters, metrics, and plots logged by DVCLive. This could happen either during logging or at the end before creating the experiment ref. Depends on https://github.com/iterative/dvc/issues/8371.
>
> If DVCLive is run inside of a DVC pipeline, can we detect and skip all of this functionality for now? Long term, it would be nice for DVC pipelines to be able to support this DVCLive-configured `dvc.yaml`.

_Originally posted by @dberenbaum in https://github.com/iterative/dvclive/issues/310#issuecomment-1259774535_

dberenbaum:
> If DVCLive is run inside of a DVC pipeline, can we detect and skip all of this functionality for now? Long term, it would be nice for DVC pipelines to be able to support this DVCLive-configured `dvc.yaml`.

Not sure we need to actually skip writing to `dvc.yaml` in pipelines. Probably we need an option to choose whether to skip, since if someone has a custom plots configuration, for example, they won't want to use the auto-generated `dvc.yaml`.

daavoo:
I guess we could have DVCLive auto-generate names for now (related to https://github.com/iterative/vscode-dvc/issues/2457) if it's easier than revisiting how exp names are assigned, but it might be better to track changes across the entire repo instead of only in `dvc.lock`.

_Originally posted by @dberenbaum in https://github.com/iterative/dvc/pull/8408#discussion_r1014642023_

Issue #271, opened 2022-08-10
log_plot(): add custom line plot


gcaria:
During my evaluation stage, I plot a simple line plot where the x values are predefined, and the y values are calculated from truth and predicted values, and are informative of the model performance.

I'd love to see how the line plot changes from experiment to experiment, which means that I'd like to use `log_plot` for doing this.

Would it be possible to implement such a feature? It seems to me that `log_plot` offers advanced features (ROC etc) but not a simple one like this.

dberenbaum:
> During my evaluation stage, I plot a simple line plot where the x values are predefined, and the y values are calculated from truth and predicted values, and are informative of the model performance.

Could you clarify more what kind of plot you would like or give an example?

gcaria:
Although I have a specific line plot in mind, I've tried to be as generic as possible because this is not really relevant for what I'm proposing.

In simple terms, the user would just need to specify two arrays (in a json file I guess), one for the x and one for the y coordinates of the points in a line plot (in my specific case the x values would always be the same, but they could change too, I don't expect this to be requirement).

I haven't used the available options for `log_plot()` because they don't apply to my case, but it seems they all have to perform some kind of computation on the json data to get the x and y values that actually go into the plot. What I'm proposing is a simpler logging option, where the user provides directly x and y (which she computed in whatever way she wants).

daavoo:
We would need to change the current API of `log_plot` a little bit.

We currently use `name` (first arg) to select the plot template to use (https://dvc.org/doc/dvclive/api-reference/live/log_plot#supported-plots):

```python
y_true = [0, 0, 1, 1]
y_score = [0.1, 0.4, 0.35, 0.8]
live.log_plot("calibration", y_true, y_score)
````
`calibration` above indicates both the name of the output plot and the template to use.

If we add support to this, I assume that it would make sense to support arbitrary names.
So, we would need to do 2 things: 
- Introduce support for `linear` template
- Decouple `name` and introduce a new `template` arg (💣 breaking change)

```python
x = [0, 1, 2]
y = [0.1, 0.2, 0.3]
live.log_plot(x, y, name="foo", tempalte="linear",)
```

gcaria:
My two cents on the API, admitting that I haven't seen the DVC code yet, but just to give a suggestion on how to minimize the changes, more importantly the breaking ones:

[...]

PR #322, opened 2022-10-04
live: Revisit output names and structure.

### LINKED ISSUE #246: live: Revisit output names
can we rename it? is it a default name? it's a bit confusing to have `dvclive` dir, `dvclive/scalars` which are actually plots and `dvclive.json` that is actually scalars ... how can we improve this naming? cc @dberenbaum @daavoo wdyt?

_Originally posted by @shcheklein in https://github.com/iterative/vscode-dvc/pull/1618#discussion_r859418877_

daavoo:
Applied https://github.com/iterative/dvclive/issues/246#issuecomment-1258582603

Closes #246

Note this targets branch `1.0`

---

For this script:

```python
import random

from dvclive import Live
from PIL import Image
from PIL import ImageDraw 

live = Live()

for i in range(2):
    live.log("foobar", i + random.random())
    live.log("foo/bar", i + random.random())

    img = Image.new("RGB", (50, 50), (255, 255, 255))
    draw = ImageDraw.Draw(img)
    draw.text((20, 20), f"{i  + random.random():.1f}", (0,0,0))
    live.log_image("img.png", img)

    live.next_step()

live.set_step(None)
live.log_plot("confusion_matrix", [0, 0, 1, 1], [0, 1, 0, 1])
```

This is the output:

```
dvclive
├── metrics.json
├── plots
│   ├── data_series
│   │   └── confusion_matrix.json
│   ├── images
│   │   ├── 0
│   │   │   └── img.png
│   │   └── 1
│   │       └── img.png
│   └── metrics
│       ├── foo
│       │   └── bar.tsv
│       └── foobar.tsv
└── report.html
```

### DIFF
--- a/src/dvclive/data/plot.py
+++ b/src/dvclive/data/plot.py
@@ -6,7 +6,7 @@ from .base import Data
 
 class Plot(Data):
     suffixes = [".json"]
-    subfolder = "plots"
+    subfolder = "data_series"
 
     @property
     def output_path(self) -> Path:
--- a/src/dvclive/data/__init__.py
+++ b/src/dvclive/data/__init__.py
@@ -1,6 +1,6 @@
 from .image import Image
+from .metric import Metric
 from .plot import Calibration, ConfusionMatrix, Det, PrecisionRecall, Roc
-from .scalar import Scalar
 from .utils import NumpyEncoder  # noqa: F401
 
 PLOTS = {
@@ -10,4 +10,4 @@ PLOTS = {
     "precision_recall": PrecisionRecall,
     "roc": Roc,
 }
-DATA_TYPES = (*PLOTS.values(), Scalar, Image)
+DATA_TYPES = (*PLOTS.values(), Metric, Image)
--- a/src/dvclive/data/metric.py
+++ b/src/dvclive/data/metric.py
@@ -10,9 +10,9 @@ from .base import Data
 from .utils import NUMPY_SCALARS
 
 
-class Scalar(Data):
+class Metric(Data):
     suffixes = [".csv", ".tsv"]
-    subfolder = "scalars"
+    subfolder = "metrics"
 
     @staticmethod
     def could_log(val: object) -> bool:
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -10,7 +10,7 @@ from typing import Any, Dict, List, Optional, Union
 from ruamel.yaml.representer import RepresenterError
 
 from . import env
-from .data import DATA_TYPES, PLOTS, Image, NumpyEncoder, Scalar
+from .data import DATA_TYPES, PLOTS, Image, Metric, NumpyEncoder
 from .dvc import make_checkpoint
 from .error import (
     ConfigMismatchError,
@@ -109,10 +109,10 @@ class Live:
     def _cleanup(self):
         for data_type in DATA_TYPES:
             shutil.rmtree(
-                Path(self.dir) / data_type.subfolder, ignore_errors=True
+                Path(self.plots_path) / data_type.subfolder, ignore_errors=True
             )
 
-        for f in (self.summary_path, self.report_path, self.params_path):
+        for f in (self.metrics_path, self.report_path, self.params_path):
             if os.path.exists(f):
                 os.remove(f)
 
@@ -153,12 +153,12 @@ class Live:
         return os.path.join(self.dir, "params.yaml")
 
     @property
-    def exists(self):
-        return os.path.isdir(self.dir)
+    def metrics_path(self):
+        return os.path.join(self.dir, "metrics.json")
 
     @property
-    def summary_path(self):
-        return str(self.dir) + ".json"
+    def plots_path(self):
+        return os.path.join(self.dir, "plots")
 
     def get_step(self) -> int:
         return self._step or 0
@@ -194,13 +194,13 @@ class Live:
         self.set_step(self.get_step() + 1)
 
     def log(self, name: str, val: Union[int, float]):
-        if not Scalar.could_log(val):
+        if not Metric.could_log(val):
             raise InvalidDataTypeError(name, type(val))
 
         if name in self._scalars:
             data = self._scalars[name]
         else:
-            data = Scalar(name, self.dir)
+            data = Metric(name, self.plots_path)
             self._scalars[name] = data
 
         data.dump(val, self._step)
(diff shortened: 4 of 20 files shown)


src/dvclive/data/plot.py
```diff
@@ -6,7 +6,7 @@
 
 class Plot(Data):
     suffixes = [".json"]
-    subfolder = "plots"
+    subfolder = "data_series"
```
daavoo, line 9: I believe this is the only name I don't really like. It doesn't really match the meaning of VSCode extension as their `data series` is our `metrics` and this folder can contain things like confusion matrix, so it doesn't really apply

dberenbaum: `sklearn` seems too implementation-focused, but it's mentioned in the [docs](https://dvc.org/doc/dvclive/api-reference/live/log_plot), and it's more descriptive than `data_series`.

dberenbaum: `json` is another that's too implementation-focused but at least descriptive.

dberenbaum: Or maybe something like `builtin` or `presets`?

daavoo: @shcheklein @jorgeorpinel 
Any suggestions regarding alternatives to the `data_series` name (and/or the rest of the outputs)? 

`data_series` is meant to store pre-defined plots based on [scikit-learn visualizations ](https://scikit-learn.org/stable/visualizations.html).

The current state is:

| Method  | Output |
| ------------- | ------------- |
| [live.log](https://dvc.org/doc/dvclive/api-reference/live/log) | `dvclive/plots/metrics` & `dvclive/metrics.json` |
| [live.log_image](https://dvc.org/doc/dvclive/api-reference/live/log_image)  | `dvclive/plots/images` |
| [live.log_param](https://dvc.org/doc/dvclive/api-reference/live/log_param)  | `dvclive/params.yaml` |
| [live.log_plot](https://dvc.org/doc/dvclive/api-reference/live/log_lot)  | `dvclive/plots/data_series` |

daavoo: Just checked in the VSCode extension that `Data Series` actually contains any template (including confusion matrix) so I guess we are actually being consistent here

shcheklein: yep, to be honest I also don't like it (had the same struggle when I was trying to come with VS Code section names). Some ideas from the top of my head:

- would it make sense to try to make `live.log` take care of `live.log_plot`. A bit of confusion / ugliness come from the mismatch between `data_series` and `_plot` in the name.
- alternative could be to dump plots right under the `dvclive/plots` (+ still have `images` and `metrics` as special subfolders)?

dberenbaum: > * would it make sense to try to make `live.log` take care of `live.log_plot`

Instead of combining the methods, we could rename `live.log_plot` to match whatever we decide for a name here.

Maybe a good convention is to name the methods according to their inputs (rather than thinking about the outputs), since it:
1. Is more self-explanatory to the user what to input.
2. Leaves flexibility to change the outputs without breaking the method naming. 
 
With that in mind, I think `live.log_sklearn` makes sense since the inputs closely follow the underlying sklearn methods. The output could then be `plots/sklearn`.

shcheklein: (btw, by saying "I don't like it" I don't mean that we should not go with this - I'm sharing my thoughts. If we don't come with something better we should proceed with this)

> With that in mind, I think live.log_sklearn makes sense since the inputs closely follow the underlying sklearn methods. The output could then be plots/sklearn.

hmm, that's interesting ... maybe I missed something, but why `sklearn` specifically?

dberenbaum: From https://dvc.org/doc/dvclive/api-reference/live/log_plot and the link at the top (which may need to be updated because it doesn't link to the expected methods), the method arguments come from sklearn. Most sklearn evaluation methods take similar arguments like `metrics.method(y_true, y_score)`. 


If we start to log any other plot types that take different arguments, we are likely to need a separate method. 

For users who are familiar, it's the easiest way to explain the inputs, and sklearn is used for evaluation even when it's not used for modeling (the top result for pytorch roc curve is https://discuss.pytorch.org/t/how-to-plot-roc-curve-using-pytorch-model/122033).

shcheklein: > From https://dvc.org/doc/dvclive/api-reference/live/log_plot and the link at the top ...

Got it, thanks! I forgot (not the first time :) ) that interface there was coupled with the sklearn types of plots. Yes, it makes more sense to rename into `sklearn_plot`. On the other hand, we already do things like this:

https://github.com/iterative/example-get-started/blob/main/src/evaluate.py#L49-L59

should it be taken care of by the some kind of `log_plot("<str>") ?

feels like even if rename the existing one to `log_sklearn_plot` or something, we'll need the second `log_plot` pretty soon (now?)

daavoo: >  the method arguments come from sklearn. Most sklearn evaluation methods take similar arguments like 

Not only that. We use `sklearn` internally to compute the results to be plotted.

> > If we start to log any other plot types that take different arguments, we are likely to need a separate method.

Indeed, we would need to define a different interface for plots beyond the predefined `sklearn` set. 

What `wandb` does is provide [secondary methods](https://docs.wandb.ai/guides/track/log/plots) that are passed as arguments to `wandb.log`:

```python
data = [[x, y] for (x, y) in zip(x_values, y_values)]
table = wandb.Table(data=data, columns = ["x", "y"])
wandb.log({"my_custom_plot_id" : wandb.plot.line(table, "x", "y",
           title="Custom Y vs X Line Plot")})
```

> should it be taken care of by the some kind of `log_plot("") ?

It could already use `log_sklearn_plot("precision_recall", ...)` but afaik (per the comment above) it was kept to showcase an alternative approach of manually generating a plot file.

dberenbaum: Sounds like we can move forward with `log_sklearn_plot`?

> > should it be taken care of by the some kind of `log_plot("") ?
> 
> It could already use `log_sklearn_plot("precision_recall", ...)` but afaik (per the comment above) it was kept to showcase an alternative approach of manually generating a plot file.

Even if we don't need it for the example get started repo, it would probably make sense after https://github.com/iterative/dvclive/issues/311 to introduce a more generic method that logs any dvc plot (`log_custom_plot`?). Should we discuss in #271?

src/dvclive/data/sklearn.py
```diff
@@ -4,7 +4,7 @@
 from .base import Data
 
 
-class Plot(Data):
+class SKLearn(Data):
```
shcheklein, line 7: should we name it SKLeanrPlot (sklearn alone can be misleading?)

dberenbaum (approved):

PR #329, opened 2022-10-17
fast ci with cpu-based wheels

skshetry:


pared (approved):

daavoo:
Hard to say because is not deterministic to compare across different workflow runs, but it doesn't look like there is much difference in speed?

skshetry:
It does seem to save 2-3mins in CI.

Maybe more so locally (torch is alone 1.3Gb compared to 170Mb CPU based wheel, similar difference in size with tensorflow-cpu).

In CI, most of the time is being spent on installation of the wheel rather than download, so if we could cache venv, it’d have been much faster.

Issue #326, opened 2022-10-14
plots: make output structure consistent between plot types


dberenbaum:
Follow up to #322.

The difference in behavior between `log_image` and `log_sklearn_plot` bothers me:

```
├── plots
│   ├── images
│   │   ├── 0
│   │   │   └── img.png
│   │   └── 1
│   │       └── img.png
│   └── sklearn
│       └── confusion_matrix.json
```

Now that they both get saved under `plots`, it seems more odd that they have such different behavior and output structure. `images` create a new folder for each step, while `sklearn` throws an exception when used with multiple steps. #271 will only make this more complicated. 

For 1.0, dvclive should be more consistently opinionated. Since the focus has shifted a bit from serverless live tracking to more of a convenience for dvc logging, I would vote to overwrite plots at each step since this is how tools like dvc plots diff, vs code, and studio expect them (this was my mistake pushing for the nested structure originally). That would mean the output would look like:

```
├── plots
│   ├── images
│   │   └── img.png # flatten and overwrite at each step.
│   └── sklearn
│       └── confusion_matrix.json # enable overwriting at each step.
```

daavoo:
I think it makes sense as default behavior, but doesn't really fix the use case of wanting to view the image updates across epochs (https://github.com/iterative/vscode-dvc/issues/1640).

Nested structure makes sense for that use case and anyhow if users want that structure they can manually generate it (and we can document it) as follows:

```python
live.log_image(img, f"{live.get_step()}/img.png")
```

dberenbaum:
> ```python
> live.log_image(f"{live.get_step()}/img.png", img)
> ```

What about `live.log_image(f"img/{live.get_step()}.png", img)`? That gives me all steps in a single dir, which is easier for me to browse and easier for us to handle in VS Code, Studio, etc.

PR #331, opened 2022-10-18
Introduce public summary. Remove "no step" / "step" logic from plots.

daavoo:
"no step" / "step" logic was initially introduced to support different logging formats between step and not step updates.

For `live.log_image`, "step" mode now overwrites the path instead of creating subfolder by step. Closes https://github.com/iterative/dvclive/issues/326

For `live.log`, the "no step" was meant not to generate the `.tsv` file but only the `.json`.
Added a public property `summary` so "no step" scenarios can work as follows:

```python
live = Live()

live.summary["foo"] = 1
live.make_summary()
```

This is a similar design used by wandb (https://docs.wandb.ai/guides/track/log#summary-metrics), where the latest values are in summary by default but single scalars are supposed to be added via manual `summary` modification instead of `log`


daavoo:
> QQ: so, how will it look like for the `example-get-started` repo, for example? 

If you leave the code as it is right now, everything will be the same with the exception that a `.tsv` file will be additionally created in `dvclive/plots/metrics`. 

> We'll need to modify summary to log "scalars"?

If you don't want to generate a single-value `.tsv` plot file, yes.

> Also, can rename the file?

Output files have already been renamed to `metrics` in #322 
Are you referring to the property `summary`?

tests/plots/test_sklearn.py
```diff
@@ -124,6 +111,19 @@ def test_dump_kwargs(tmp_dir, y_true_y_pred_y_score, mocker):
...
+    live = Live()
+
+    for step in range(2):
+        live.log_sklearn_plot("confusion_matrix", [0, 0], [step, step])
+        live.next_step()
+
+    plot_path = tmp_dir / live.plots_path / SKLearnPlot.subfolder
+    plot_path = plot_path / "confusion_matrix.json"
+
+    assert json.loads(plot_path.read_text())[0]["predicted"] == str(step)
```
pared, line 124:
```suggestion
def test_override_on_step(tmp_dir):
    live = Live()

    live.log_sklearn_plot("confusion_matrix", [0, 0], [0, 0])
    live.next_step()
    live.log_sklearn_plot("confusion_matrix", [0, 0], [1, 1])

    plot_path = tmp_dir / live.plots_path / SKLearnPlot.subfolder
    plot_path = plot_path / "confusion_matrix.json"

    assert json.loads(plot_path.read_text()) == [{"actual": "0", "predicted": "1"}, {"actual": "0", "predicted": "1"}]
```

pared: Seems easier to parse for me, WDYT?

tests/plots/test_image.py
```diff
@@ -0,0 +1,60 @@
...
+
+def test_step_dump(tmp_dir):
+    live = Live()
+    for step in range(2):
+        img = np.ones((10, 10, 3), np.uint8) * step
+        live.log_image("image.png", img)
+        live.next_step()
+
+    img_path = tmp_dir / live.plots_path / LiveImage.subfolder / "image.png"
+    assert np.array(Image.open(img_path))[0][0][0] == step
```
pared, line 46:
```suggestion
def test_override_on_step(tmp_dir):
    live = Live()

    zeros = np.zeros((2, 2, 3), np.uint8)
    live.log_image("image.png", zeros)

    live.next_step()

    ones = np.ones((2, 2, 3), np.uint8)
    live.log_image("image.png", ones)

    img_path = tmp_dir / live.plots_path / LiveImage.subfolder / "image.png"
    assert np.array_equal(np.array(Image.open(img_path)), ones)
```

tests/test_main.py
```diff
@@ -379,3 +380,14 @@ def test_logger(tmp_dir, mocker, monkeypatch):
 
     live = Live(resume=True)
     logger.info.assert_called_with("Resumed from step 0")
+
+
+def test_no_log_summary(tmp_dir):
```
pared, line 385: I am not sure I understand - does this test name mean dvclive should not log `summary.json` but `metrics.json`?

dberenbaum:
Can you easily separate the `log_image` changes from the rest? I think that's easy to merge, but the other part seems a little more divisive .

> This is a similar design used by wandb (https://docs.wandb.ai/guides/track/log#summary-metrics), where the latest values are in summary by default but single scalars are supposed to be added via manual `summary` modification instead of `log`

What about params? Those are non-step values that get logged via a logger method, although I think we discussed setting them available by dict modification.

It seems this PR is only halfway towards what wandb and mlflow do, since they both have `step = 0` by default, whereas this leaves some in-between state where the default is `step = None` and there is no `step` value in `metrics.json`, but dvclive writes out the tsv files with `step = 0`. Would it simplify logic internally to default to `step = 0`?

> it feels we are optimizing for DL scenarios a bit too much?

Assuming a step-based workflow is what both mlflow and wandb do, so I can understand the push to do the same. However, I don't like that dvclive would then be generating single-point tsv plots from all the metrics. Maybe we can leave in some logic for step 0/None to ignore writing to tsv until `set_step` is called/wait until there is a change in a metric before writing it to tsv (and possibly not add step 0 to `metrics.json`, although I'm less worried about this)?

dberenbaum:
> Should `make_summary` be called `write_summary` or just `log_summary`?

Hmm, no strong opinion, but should this method also be either made private or documented?

tests/test_main.py
daavoo: The summary file is currently called `metrics.json` (could be changed though).

The test is for the workflow where you can write to the summary file without calling `live.log` but rather editing the `summary` property directly.

daavoo:
> What about params? Those are non-step values that get logged via a logger method, although I think we discussed setting them available by dict modification.

I think `params` should be handled like `summary` here (and `wandb.config` https://docs.wandb.ai/guides/track/config).

> It seems this PR is only halfway towards what wandb and mlflow do, since they both have step = 0 by default, whereas this leaves some in-between state where the default is step = None and there is no step value in metrics.json, but dvclive writes out the tsv files with step = 0. Would it simplify logic internally to default to step = 0?

Not sure I see the problem. 

`there is no step value in metrics.json` can only happen if the user doesn't update the step value. In this no step scenario I see no issue with not having step in `metrics.json`

> Maybe we can leave in some logic for step 0/None to ignore writing to tsv until set_step is called/wait until there is a change in a metric before writing it to tsv (and possibly not add step 0 to metrics.json, although I'm less worried about this)?

The idea is that, for no-step workflow, we document the following (instead of calling `live.log_metric`):

```
live = Live()

live.summary["foo"] = 1
live.make_summary()
```

This won't generate `.tsv` but only update `metrics.json`. I was just mentioning in ivan answer that even without changing the code the result would be similar, but I mean to say that we should change the code in example-get-started to the snippet above.

> Hmm, no strong opinion, but should this method also be either made private or documented?

Method should be public to document the snippet above

dberenbaum:
> The idea is that, for no-step workflow, we document the following (instead of calling `live.log_metric`):
> 
> ```python
> live = Live()
> 
> live.summary["foo"] = 1
> live.make_summary()
> ```

I don't mind supporting it, but I wouldn't expect that the summary dict is a primary workflow, and I wouldn't introduce it in our example-get-started project. wandb doesn't introduce this in their get started materials, and I think they generally assume users are fine with logging to `step=0` by default. I think we should either keep no-step logic or live with some ugly output in example-get-started (primarily the step-0 tsv files/plots).

> However, I don't like that dvclive would then be generating single-point tsv plots from all the metrics. Maybe we can leave in some logic for step 0/None to ignore writing to tsv until `set_step` is called/wait until there is a change in a metric before writing it to tsv (and possibly not add step 0 to `metrics.json`, although I'm less worried about this)?

> `there is no step value in metrics.json` can only happen if the user doesn't update the step value. In this no step scenario I see no issue with not having step in `metrics.json`

Why is there a step logged to the tsv file? This seems more important than `metrics.json` since it generates new directories, files, and plots that aren't useful.

> Would it simplify logic internally to default to step = 0?

What I meant here is there still seems to be some "no step" logic that we could drop if we default to `self._step = 0`:

https://github.com/iterative/dvclive/blob/e115105495746f73721aa14a76fc431b46f05761/src/dvclive/live.py#L162

https://github.com/iterative/dvclive/blob/e115105495746f73721aa14a76fc431b46f05761/src/dvclive/live.py#L165-L167

https://github.com/iterative/dvclive/blob/e115105495746f73721aa14a76fc431b46f05761/src/dvclive/live.py#L260-L261

tests/test_main.py
pared: I guess we could add here a functionality checking that `foo.tsv` does not exist - to make sure that values added via dict modification don't get logged into `tsv` files. WDYT?

daavoo:
> I don't mind supporting it, but I wouldn't expect that the summary dict is a primary workflow, and I wouldn't introduce it in our example-get-started project. wandb doesn't introduce this in their get started materials, and I think they generally assume users are fine with logging to step=0 by default. I think we should either keep no-step logic or live with some ugly output in example-get-started (primarily the step-0 tsv files/plots).

I don't see it that way. I think the difference is that our example-get-started focuses on "no step" scenario whereas any other experiment tracker focuses on "step" scenarios first (or even only).

The introduction to `wandb.log` (https://docs.wandb.ai/guides/track/log) is clearly oriented towards a "step" scenario and yet the `summary` workflow has a dedicated section (https://docs.wandb.ai/guides/track/log#summary-metrics)

dberenbaum:
Discussed with @daavoo and agreed to:
1. Keep using `live.log()` as the primary workflow for step and no-step scenarios but without any special syntax for no-step scenarios. This means no-step workflows will still generate tsv files and have `step: 0` in the summary. This is no different than mlflow and wandb and doesn't seem to bother people.
2. Introduce `live.summary["foo"] = 1` from this PR as a special case (for example, a step scenario where some metrics are not step-based).

dberenbaum:
Sorry, forgot to approve this one.

PR #338, opened 2022-10-24
live.set_step: Remove `make_checkpoint` and `make_report` calls.

### LINKED ISSUE #103: Decouple epoch metrics logging and checkpoints
Users may want to save metrics every epoch but only save models every few epochs. Can we support this in dvclive?

### LINKED ISSUE #231: `report`: how to make report in no-step scenarios

daavoo:
Call `make_report` manually in the integrations.

Closes #231
Closes #103

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -326,21 +326,6 @@ def test_get_step_control_flow(tmp_dir):
     assert values == [float(x) for x in range(10)]
 
 
-def test_make_checkpoint(tmp_dir, mocker, monkeypatch):
-    make_checkpoint = mocker.patch("dvclive.live.make_checkpoint")
-
-    dvclive = Live()
-    dvclive.log("foo", 1)
-    dvclive.next_step()
-    assert not make_checkpoint.called
-
-    monkeypatch.setenv(env.DVC_CHECKPOINT, True)
-    dvclive = Live()
-    dvclive.log("foo", 1)
-    dvclive.next_step()
-    assert make_checkpoint.called
-
-
 def test_logger(tmp_dir, mocker, monkeypatch):
     logger = mocker.patch("dvclive.live.logger")
     monkeypatch.setenv(env.DVCLIVE_LOGLEVEL, "DEBUG")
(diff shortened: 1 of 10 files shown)


daavoo:
@dberenbaum Merging for now but we can get back and change stuff if something is found during docs discussions

tests/test_main.py
```diff
@@ -326,21 +326,6 @@ def test_get_step_control_flow(tmp_dir):
...
-    dvclive = Live()
-    dvclive.log("foo", 1)
-    dvclive.next_step()
-    assert not make_checkpoint.called
-
-    monkeypatch.setenv(env.DVC_CHECKPOINT, True)
-    dvclive = Live()
-    dvclive.log("foo", 1)
-    dvclive.next_step()
-    assert make_checkpoint.called
```
pared, line 341: I think we should have a test veryfing that `make_checkpoint` actually makes the file. 

Created #341

pared (approved):

dberenbaum:
Why is `make_report` added to all the integrations but `make_checkpoint` is not?

Can we still have `next_step` call these by default? I like keeping it all modular like this, but it's now a lot of methods needed at each step, especially along with https://github.com/iterative/dvclive/issues/238#issuecomment-1293877535, and these aren't methods that users would need in mlflow or other experiment trackers.

As far as I can tell, a basic workflow without an integration would looks something like:

```
for step in ...:
    live.log_metric(...)
    live.make_report()
    live.make_checkpoint() # Might not be needed
    live.make_summary()
    live.next_step()
```

What if we make these part of `next_step` but not `set_step`? `next_step` runs `set_step` along with all the other methods above. This still gives users modularity if needed but makes a default workflow simpler.

PR #339, opened 2022-10-24
log_sklearn_plot: Support custom name.

### LINKED ISSUE #323: log_plot: custom filename and/or title for sklearn plots
See https://github.com/iterative/example-repos-dev/blob/b8a8f0af39b88c735af0b8065c37b609242b53ed/example-get-started/code/src/evaluate.py#L61. I have to keep two separate live instances to be able to log sklearn plots for both train and test data because there's no way to create multiple confusion matrix (or any other type) plots in the same instance.

daavoo:
Rename the existing `name` argument to `kind`. This argument is used to determine the type of plot. Add a new optional `name` argument to define the output file.

Closes #323

### DIFF
--- a/src/dvclive/data/sklearn_plot.py
+++ b/src/dvclive/data/sklearn_plot.py
@@ -10,7 +10,8 @@ class SKLearnPlot(Data):
 
     @property
     def output_path(self) -> Path:
-        _path = Path(f"{self.output_folder / self.name}.json")
+        _name = self.name.replace(".json", "")
+        _path = Path(f"{self.output_folder / _name}.json")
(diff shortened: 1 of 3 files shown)


daavoo:
> One minor request: look for a `.json` extension and strip it so users don't end up with `roc.json.json`.

should be done now

dberenbaum:
Nice, thanks! Is there a docs PR?

daavoo:
> Nice, thanks! Is there a docs PR?

I am grouping all 1.0 in a single branch

dberenbaum:
Can we have separate PRs and merge into a similar dvclive-1.0 branch in dvc.org so we don't have one massive review?

Issue #341, opened 2022-10-27
tests: test making checkpoint


pared:
We do not have a test verifying that our `make_checkpoint` behaves the same as its counterpart in dvc.

PR #342, opened 2022-10-27
integration: dvc: test checkpoint file creation

pared:
Fixes: #341

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -398,3 +398,37 @@ def test_context_manager(tmp_dir):
     assert not log_file.exists()
     report_file = tmp_dir / live.report_file
     assert report_file.exists()
+
+
+@pytest.mark.parametrize("dvc_root", [True, False])
+@pytest.mark.parametrize("set_env", [True, False])
+def test_create_checkpoint_file(
+    tmp_dir, monkeypatch, dvc_root, set_env, mocker
+):
+    if dvc_root:
+        (tmp_dir / ".dvc" / "tmp").mkdir(parents=True)
+    if set_env:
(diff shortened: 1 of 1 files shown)


src/dvclive/dvc.py
```diff
@@ -36,7 +36,10 @@ def make_checkpoint():
     if not root_dir:
         return
 
-    signal_file = os.path.join(root_dir, ".dvc", "tmp", env.DVC_CHECKPOINT)
+    tmp_dir = os.path.join(root_dir, ".dvc", "tmp")
+    os.makedirs(tmp_dir, exist_ok=True)
```
daavoo, line 40: Do we need this? `dvc.api.make_checkpoint` doesn't do it

pared: I think so - on dvc side we can be pretty sure that tmp is in place. I think that in `dvclive` we should not rely on it.
For example:
1. someone clones the repo - there is no `tmp` before first dvc command
2. someone runs training script without `dvc repro` - as cmd defined in dvc.yaml
3. script fails because `.dvc` exists and `.dvc/tmp` does not

Actually that might mean that we need to fix `dvc.api.make_checkpoint` too - seems like this use case is valid there too.

daavoo: > script fails because .dvc exists and .dvc/tmp does not

I think this would be a very edge case, it would require the user to manually set up the `env.DVC_CHECKPOINT`

pared: > I think this would be a very edge case, it requires the user to manually set up the env.DVC_CHECKPOINT

Good point, in case of exp run we will get the tmp.

daavoo (approved):

PR #343, opened 2022-10-27
live: Rename `log` to `log_metric`.

### LINKED ISSUE #336: `log`: make `log_metric` alias?
Not sure why anyone would want a longer method, but `log_metric` is more explicit. Is it worth adding for clarification?

daavoo:
Closes #336

### DIFF
--- a/src/dvclive/catalyst.py
+++ b/src/dvclive/catalyst.py
@@ -18,7 +18,7 @@ class DvcLiveCallback(Callback):
         for loader_key, per_loader_metrics in runner.epoch_metrics.items():
             for key, value in per_loader_metrics.items():
                 key = key.replace("/", "_")
-                self.dvclive.log(f"{loader_key}/{key}", float(value))
+                self.dvclive.log_metric(f"{loader_key}/{key}", float(value))
 
         if self.model_file:
             checkpoint = utils.pack_checkpoint(
--- a/src/dvclive/fastai.py
+++ b/src/dvclive/fastai.py
@@ -18,7 +18,7 @@ class DvcLiveCallback(Callback):
         for key, value in zip(
             self.learn.recorder.metric_names, self.learn.recorder.log
         ):
-            self.dvclive.log(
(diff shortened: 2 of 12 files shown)


daavoo (commented):
@dberenbaum went for replacing instead of alias because of https://github.com/iterative/dvclive/issues/336#issuecomment-1293352694. 
At this point where we have all the other methods named like `log_{x}` so `log` might look like a particular case for logging anything (we could consider actually implementing a `log` for anything but that would be a separate request)

dberenbaum:
I guess now is the time to do it 😄

dberenbaum:
One thought here: since I consider this a pretty severe breaking change, can we leave an unimplemented `log` method that raises an exception telling users it's been changed to `log_metric`?

daavoo:
> One thought here: since I consider this a pretty severe breaking change, can we leave an unimplemented `log` method that raises an exception telling users it's been changed to `log_metric`?

I guess I can just add the alias instead as it would take as much code

dberenbaum:
> I guess I can just add the alias instead as it would take as much code

I think that depends on how much you worry that `log` may cause confusion, or if you think we need to reuse `log` for something in the near future.

daavoo:
> I think that depends on how much you worry that log may cause confusion, 

I don't worry that much, but the thing is that I don't see any benefit of keeping `log` apart from the fact that removing it is a breaking change. Given that 1.0 is full of those, sounds like the time to remove it.

> or if you think we need to reuse log for something in the near future.

A general purpose `log` similar to `wandb.log` would be a breaking change to the API if we keep the current backward compatible `log`.

dberenbaum (approved):
I'm fine with whatever you decide here.

skshetry:
What about `log_metrics`? Is that required?

daavoo:
> What about `log_metrics`? Do we need that too?

https://github.com/iterative/dvclive/issues/350

PR #345, opened 2022-10-28
live: Replace `path` with `dir`/`file`.

daavoo:
`path` is ambiguous. There was also an inconsistency between the terms used: `path` for the argument and `dir` for the property.


dberenbaum:
I haven't done any QA, so trusting you to make sure everything works the same, but no product concerns.

dberenbaum (approved):

daavoo:
> I haven't done any QA, so trusting you to make sure everything works the same, but no product concerns.

I am trusting on CI 😅

PR #352, opened 2022-11-02
log_metric: Make timestamp optional.

### LINKED ISSUE #233: Option to disable timestamps for scalar output
Currently scalar values have timestamps associated with each step that is written to the csv file. It would be nice to be able to disable this for cases where the files are stored within source control as timestamps are always going to be different, even if the step and values are the same.

daavoo:
Default to not include timestamp.
Closes #233


dberenbaum:
Why default to false? I think most people would prefer to have them.

daavoo:
> Why default to false? I think most people would prefer to have them.

We don't display them in any of the UIs. How are most people using them?

dberenbaum (approved):
Okay then. LGTM.

PR #354, opened 2022-11-03
live: Add `__enter__` and `__exit__`.

daavoo:
Wait for #353

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -243,6 +243,8 @@ class Live:
         if self.report_mode == "studio":
             if not post_to_studio(self, "done", logger):
                 logger.warning("`post_to_studio` `done` event failed.")
+        else:
+            self.make_report()
 
     def make_checkpoint(self):
         if env2bool(env.DVC_CHECKPOINT):
@@ -257,3 +259,9 @@ class Live:
     def read_latest(self):
         with open(self.metrics_file, encoding="utf-8") as fobj:
             return json.load(fobj)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.end()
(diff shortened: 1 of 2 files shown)


tests/test_main.py
```diff
@@ -355,3 +359,29 @@ def test_make_summary_without_calling_log(tmp_dir):
...
+        # no `step`
+        "foo": 1.0
+    }
+    log_file = tmp_dir / live.plots_dir / Metric.subfolder / "foo.tsv"
+    assert not log_file.exists()
+
+
+def test_context_manager(tmp_dir):
+    with Live() as live:
+        live.summary["foo"] = 1.0
```
daavoo, line 380: `__exit__` only calls `live.make_summary` and `post_to_studio` for now. 

After https://github.com/iterative/dvc/pull/8408 is available, the context manager should take care of generating the DVC experiment

daavoo: @dberenbaum Should `__exit__` also call `make_report()`? I think so

dberenbaum: Yup, I agree, it should call `make_report()`.

dberenbaum (approved):

src/dvclive/live.py
```diff
@@ -257,3 +259,9 @@ def read_step(self):
     def read_latest(self):
         with open(self.metrics_file, encoding="utf-8") as fobj:
             return json.load(fobj)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.end()
```
dberenbaum, line 267: Minor follow-up: let's either make `end()` public or get rid of it and move the logic here. What do you think @daavoo?

dberenbaum: I slightly prefer getting rid of `end()` since I'm not sure it's worth adding to docs, but no strong preference

PR #356, opened 2022-11-04
Added Optuna callback.

### LINKED ISSUE #118: integrations: Optuna
From [Optuna](https://github.com/optuna/optuna):

> Optuna is an automatic hyperparameter optimization software framework, particularly designed for machine learning. It features an imperative, define-by-run style user API. Thanks to our define-by-run API, the code written with Optuna enjoys high modularity, and the user of Optuna can dynamically construct the search spaces for the hyperparameters.

Existing integrations with other *ML Loggers* are maintained in the repository implemented as callbacks:

https://github.com/optuna/optuna/tree/master/optuna/integration

Related discord thread:

https://discordapp.com/channels/485586884165107732/563406153334128681/866102743235297280

daavoo:
Closes #118

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -48,7 +48,7 @@ tests =
     pytest-mock==3.8.2
     pylint==2.15.0
     pylint-plugin-utils>=0.6
-    mypy==0.981
+    mypy>=1.1.1
     %(image)s
     %(plots)s
     %(markdown)s
@@ -74,6 +74,8 @@ fastai =
 pytorch_lightning =
     pytorch_lightning>=1.9
     torch<1.13
+optuna =
+    optuna
 all =
     %(image)s
     %(mmcv)s
@@ -84,6 +86,7 @@ all =
     %(catalyst)s
     %(fastai)s
     %(pytorch_lightning)s
+    %(optuna)s
     %(plots)s
     %(image)s
     %(plots)s
--- a/src/dvclive/optuna.py
+++ b/src/dvclive/optuna.py
@@ -1,0 +1,46 @@
+from dvclive import Live
+
+
+class DVCLiveCallback:
+    def __init__(self, metric_name="metric", **kwargs) -> None:
+        kwargs["dir"] = kwargs.get("dir", "dvclive-optuna")
+        kwargs.pop("save_dvc_exp", None)
+        self.metric_name = metric_name
+        self.live_kwargs = kwargs
+
+    def __call__(self, study, trial) -> None:
+        with Live(save_dvc_exp=True, **self.live_kwargs) as live:
+            self._log_metrics(trial.values, live)
+            live.log_params(trial.params)
+
+    def _log_metrics(self, values, live):
+        if values is None:
+            return
+
+        if isinstance(self.metric_name, str):
+            if len(values) > 1:
+                # Broadcast default name for multi-objective optimization.
+                names = [
+                    "{}_{}".format(self.metric_name, i) for i in range(len(values))
+                ]
+
+            else:
+                names = [self.metric_name]
+
+        else:
+            if len(self.metric_name) != len(values):
+                raise ValueError(
+                    "Running multi-objective optimization "
+                    "with {} objective values, but {} names specified. "
+                    "Match objective values and names,"
+                    "or use default broadcasting.".format(
+                        len(values), len(self.metric_name)
+                    )
+                )
+
+            else:
+                names = [*self.metric_name]
+
+        metrics = {name: val for name, val in zip(names, values)}
+        for k, v in metrics.items():
+            live.log_metric(k, v)
--- a/tests/test_frameworks/test_optuna.py
+++ b/tests/test_frameworks/test_optuna.py
@@ -1,0 +1,27 @@
+# pylint: disable=protected-access, unused-argument
+import optuna
+
+from dvclive.optuna import DVCLiveCallback
+from dvclive.serialize import load_yaml
+from dvclive.utils import parse_json
+
+
+def objective(trial):
+    x = trial.suggest_float("x", -10, 10)
+    return (x - 2) ** 2
+
+
+def test_optuna_(tmp_dir, mocked_dvc_repo):
+    n_trials = 5
+    metric_name = "custom_name"
+    callback = DVCLiveCallback(metric_name=metric_name)
+    study = optuna.create_study()
+
+    study.optimize(objective, n_trials=n_trials, callbacks=[callback])
+
+    assert mocked_dvc_repo.experiments.save.call_count == n_trials
+
+    metrics = parse_json("dvclive-optuna/metrics.json")
+    assert metric_name in metrics
+    params = load_yaml("dvclive-optuna/params.yaml")
+    assert "x" in params


daavoo:
@tibor-mach Would be great to get your input here. 
`multirun=True` is what we discussed as an alternative to `make_checkpoint`.
This doesn't really work if the optuna script wants to be run as an actual stage inside a `dvc.yaml` but we were targeting here the "no dvc.yaml" scenario.

dberenbaum:
@daavoo What do you think about starting with only `multirun = False` and adding this arg later?

@tibor-mach If you are interested, you could focus on polishing this callback and then showing it off instead of focusing on manual integration of Optuna with checkpoints. What do you think?

tibor-mach:
@daavoo @dberenbaum 
Hi guys, sorry for not replying sooner. I ended up having to move (rather unexpectedly). I will have a look at it tomorrow and let you know my 5 cents.

daavoo:
> @daavoo What do you think about starting with only multirun = False and adding this arg later?

Ok

dberenbaum:
@tibor-mach Hope everything is okay! Your timing is good, because both of us were out all last week 😄 .

tibor-mach:
@dberenbaum I think this makes sense seeing as this is also a good usecase for DVCLive (if you don't want to observe the training as it happens it is less useful to log the individual steps as experiments since for most purposes the summaries kept by the optuna study will be enough anyway...unless you want to keep track of the individual trained models as well but I believe that is a rather unusual scenario as I mentioned in the comment above.

But I think it would be good to ensure that this approach can (eventually) also lead to running this as a part of a dvc pipeline (run with repro). I am not quite sure why this doesn't work just yet, so I'd like to understand that better first.

Oh and everything's alright, thanks :-)

dberenbaum:
> But I think it would be good to ensure that this approach can (eventually) also lead to running this as a part of a dvc pipeline (run with repro). I am not quite sure why this doesn't work just yet, so I'd like to understand that better first.

Yup, makes sense. I hope if we start with only `multirun = False`, it should be possible to make it work as part of a pipeline. What do you think @daavoo?

tibor-mach:
@daavoo Any update on the questions above? If you have other priorities right now just let me know and, I will just assume that the constraints you mention are a given atm, have a look at it and at least give a feedback (or potentially  integrate it into a simple demo if it works well)

daavoo:
> @daavoo Any update on the questions above? If you have other priorities right now just let me know and, I will just assume that the constraints you mention are a given atm, have a look at it and at least give a feedback (or potentially integrate it into a simple demo if it works well)

Apologies @tibor-mach , I completely missed this notification 😞 
I will update you with the latest status

daavoo:
>     1. Why is this not going to work with a dvc.yaml file? I though the issue was with the checkpoints specifically, but perhaps this has the same low-level behaviour w.r.t. the dvc pipelines? I suppose it is fine not to integrate it with dvc.yaml yet, but any solution should keep that door open, I think. At least I would appreciate it.

In the current state of the P.R, the default behavior (`multi_exp=False`) works with `dvc.yaml`. Each individual trial is just recorded as a `step`. However, this doesn't really allow for managing trials with DVC.

With `multi_exp=True`, each trial will create a new DVC experiment. This at least allows to manage trials with `dvc exp` commands, but we we currently skip the creation of new DVC experiments when there is a `dvc.yaml`. 

>     2. The name `multirun` confused me a bit before I had a look at the code. I understand the logic, since in a way you run multiple experiment runs, but at least from a data scientist's perspective (well, mine anyway), the individual steps in hyperparameter optimization cannot really be seen as separate runs. Their order is important, since optuna uses the results from previous trials to sample the hyperparams for new trials...but I guess the names are not quite as important at this stage of development.

I renamed from multirun to multi_exp to better reflect that each trial will be a DVC experiment.

> the individual steps in hyperparameter optimization cannot really be seen as separate runs. 

I guess is a matter of preference 😅 I see no issue in considering each trial as an individual experiment. 

> Their order is important, since optuna uses the results from previous trials to sample the hyperparams for new trials

The order requirement is only True for some but not all of the sampling algorithms. In basic ones like grid search or random sampling there is no order.

>     3. I wonder whether this approach would allow versioning the trained models themselves. This is not all that useful, so it might be ok not to do it, but I can imagine some scenarios where it might make sense to log the trained models too. Specifically, if the training is very time-consuming but the trained models themselves are not prohibitively large, it might be worth saving them for further experimentation (my objective function might still be a half-baked idea and I want to explore more than just the best model/pareto front of best models). But like I said, this is likely a niche scenario.

I think it's probably not that niche, but we don't have a mechanism yet to track the model with DVC from inside the Python script (beyond what you have already tried using checkpoints). If we add this https://github.com/iterative/dvclive/issues/378, each trial could optionally include the model in each DVC experiment created

dberenbaum:
Looks good to me at a quick glance, except we may need to clean up `save_dvc_exp` and `multi_exp`.

> With `multi_exp=True`, each trial will create a new DVC experiment. This at least allows to manage trials with `dvc exp` commands, but we we currently skip the creation of new DVC experiments when there is a `dvc.yaml`.

I think this is already a bit surprising/unusual behavior in `save_dvc_exp`, and it is even moreso here. We need to explain that `save_dvc_exp` will be ignored here in favor of `multi_exp`, although they are basically equivalent? And what is the behavior if you set `multi_exp=True` but have a `dvc.yaml`?

tibor-mach:
> The order requirement is only True for some but not all of the sampling algorithms. In basic ones like grid search or random sampling there is no order.

Fair enough, that's true. I was being too categorical :-) 

Also, thanks for the explanation wtih the `multi_exp`

> And what is the behavior if you set multi_exp=True but have a dvc.yaml?

I'd also like to know that...but I guess I can play with it a bit myself. Tomorrow's Christmas Eve and then the Holidays ... but I think I will find some time to have a look at it and prepare at least a quick blogpost draft between now and the first week in January.

tibor-mach:
So I finally tested things a bit more hands-on. Some thoughts:

I was thinking a bit more about the use-case. The way I see it, the integration with DVCLive seems mostly useful to allow manual interruption of the optimization process based on observation (while logging all the trials). 

So, while the `multi_exp=False` version seems to work as intended (can be run with `dvc repro`), I wonder whether it is really all that useful. The only added benefit over not using the callback is that I get to observe the optimization, but (AFAIK) I cannot really stop it manually without making the repro run unusable (or rather not reproducible, which is the main point of `repro`).

The `multi_exp=True` seems to be a lot more useful...but I am either misunderstanding something about DVCLive or something is not working as intended. I think it goes back to this:

> With multi_exp=True, each trial will create a new DVC experiment. This at least allows to manage trials with dvc exp commands, but we we currently skip the creation of new DVC experiments when there is a dvc.yaml.

Now, if I set `multi_exp=True` and then run everything with `dvc exp run` I get the following warning after each trial

> WARNING:dvclive:Ignoring `_save_dvc_exp` because `dvc exp run` is being used.  

...which probably is the answer for the question @dberenbaum had.

Of course, @daavoo you mention it when you said

> but we we currently skip the creation of new DVC experiments when there is a dvc.yaml.

When I try to run `dvc repro` in this setup I get a similar warning:

> WARNING:dvclive:Ignoring `_save_dvc_exp` because there is an existing `dvc.yaml` file.

By the way, I looked at [where these warnings are defined](https://github.com/iterative/dvclive/blob/7764ef05db5a7b719303dde8ded2239f12137587/src/dvclive/live.py#L125-L140) and I am a bit surprised that with the latter warning you handle it by setting `self._save_dvc_exp = False` which matches the logger warning text but with the former (the one related to `dvc exp run`) you don't do that (you do change values of some other things so perhaps it is just handled differently in which sorry and ignore me on this :-) ).

In any case, I am not quite sure how to actually run it then at all when `multi_exp=True` to achieve the correct behaviour (of actually having a new experiment for each trial). The way it behaves now is that no experiments are created and values from each step are successively overwritten so that at the end, I just end up with metrics from the last step.

If you guys are thinking about deprecating checkpoints then I'd rather avoid those but now it seems to me that a combination of this callback with `multi_exp=False` and checkpoints is what would give the desired result (well, almost, since `dvc repro` doesn't work with checkpoints).

So, I'd probably rather help with polishing this before writing any blog posts. But in order to do that I think I need to understand those two warnings better.

daavoo:
Hi @tibor-mach , sorry for the late response.
---
 
As a high-level point, note that `save_dvc_exp` was introduced to provide a way for people to create dvc experiments (and see metrics/params/plots) without the need of creating a dvc pipeline (`dvc.yaml`). So, given a script using `DVCLive`, run it with `python script.py`, and be able to see plots and content in the `exp show` table.

The Optuna scenario was not the main use but more of an "exploit" of the feature, so there are things that just might not make sense and need to be changed (i.e. the restriction of not creating an experiment in you have a `dvc.yaml`)

---

> In any case, I am not quite sure how to actually run it then at all when multi_exp=True to achieve the correct behaviour (of actually having a new experiment for each trial).

As described above, if you have a script using optuna, it would need to be executed from the CLI directly (i.e. `python script_with_optuna.py`) instead of running with `dvc exp run` / `dvc repro`.

> you do change values of some other things so perhaps it is just handled differently in which sorry and ignore me on this 

There are 3 separate features with conditional behavior: sending metrics to iterative studio, creating a dvc experiment, and creating a hidden `dvc.yaml`. 
Depending on the project setup and the context of execution (`dvc exp run` vs `dvc repro` vs `python train.py`) different things need to be done and we went with the most restrictive default mode (shut down experiment creation) in the release so existing dvc projects are not altered in unexpected ways.

> If you guys are thinking about deprecating checkpoints then I'd rather avoid those but now it seems to me that a combination of this callback with multi_exp=False and checkpoints is what would give the desired result (well, almost, since dvc repro doesn't work with checkpoints).

I think it would be nice if we could come up with a better of achieving what you are doing with checkpoints but without the current checkpoint implementation, if that makes sense.

For what you are using checkpoints in the script, most of the actual logic under the hood would not be needed at all. 
Maybe we need to consider if we can make checkpoints more lightweight and make the current `exp save` create a `checkpoint` if it's called inside a `dvc exp run` execution cc @dberenbaum

daavoo:
We have made some changes in https://github.com/iterative/dvclive/releases/tag/2.0.0 and https://github.com/iterative/dvclive/releases/tag/2.0.1 that made many of the discussions above obsolete.

The only case where it won't create experiments for each trial would be if you run `dvc exp run`. Any other of the previous scenarios (`dvc repro`, `python train.py` with a DVC pipeline present) should now create experiments for each trial.

@dberenbaum I have updated the P.R. to default to `multi_exp=True` I see no reason for the opposite. I think we could merge as is

dberenbaum (approved):

dberenbaum:
@daavoo Sorry for the long delay here. Is it in good shape to merge? There's a request to try it out in [discord](https://discord.com/channels/485586884165107732/563406153334128681/1085855147424501800).

dberenbaum:
@tibor-mach Sorry this took so long. Thanks for all your hard work on this and your patience. Are you using it? Would you still have interest in publishing something about it?

dberenbaum:
Sounds great @tibor-mach! No rush. Let us know and we will be happy to work with you on it and promote it.

tibor-mach:
@dberenbaum I was rather ill most of the previous week and only now I'm getting better. Because of that, I haven't actually started writing any posts yet but today I gave it some thought. First, I am trying to come up with a really compelling use-case to demonstrate this functionality with. 

I believe there are (at least) two use-cases (see below). I actually prefer the second use-case. It seems more "DVC-ish" in that its focus is on traceability and GitOps which is where Iterative tools are the strongest (IMO) and so there I can also see a nice synergy with your other tools. The first use-case just makes use of live metric tracking via DVCLive (which can be done by other tools, possibly even Optuna on its own) and doesn't really benefit from or require reproducibility at all.

Also, sorry for the wall of text, but let me know what you think about these early drafts (cc @jendefig):

# 1. Human learning the meta-hyper-parameters (oh dear)

Let's say we have a setup with the following conditions:

1. There exists a large hyperparameter space for the model and the model takes non-trivial time to train (at least with most hyperparameter choices). 
2. the model/setting is unfamiliar and I have little intuition for what objective function value can be expected or should be considered good enough.

If the conditions above are not met, then logging intermediate trials is not very interesting:

- If my hyperparameter space is small and/or my model is quick to train I can just brute force something (even a complete grid search). I am not very concerned about the intermediate trials in this case since I can be sure I get more or less the best solution anyway. Besides the trials will all be logged in the final optuna study anyway, so I can explore them later.
- If I already have a clear idea of the objective function value which is attainable (and in acceptable time/at an acceptable cost), then I can simply [set up an early stopping rule for the study](https://github.com/optuna/optuna/issues/3685#issuecomment-1159001321) as mentioned above. In fact getting this understanding is more or less the goal of this first use-case. 

So I want to optimize my model but I am not sure how long it will take to get any good values of the objective function. Indeed, I am not even sure which values can be considered good enough (which is to say values which are obtainable at all and where the costs of reaching them are outweighed by the benefits of the somewhat better performing model).

That means I need to do a bit of exploring. One way to do that is to run my hyperparameter optimization with an exceedingly high number of trials. I am not actually expecting to run all those trials, but rather I would like to observe the hyperparameter tuning as it progresses, taking advantage of DVCLive to visualise progress. 

This allows me to get a better feeling of my setup and what can and cannot be expected. Once I conclude that any further gains in the objective function are unlikely, I simply interrupt the optimization. This will obviously not be reproducible, but it will allow me to set up a reproducible and fully automated training/optimization pipeline with a limited number of trials and/or a threshold for the objective function such that the study stops once this threshold is met. Also, I will still end up with a study which can be resumed and which will contain detailed info about all the trials that happened ( and I can use [optuna's visualisation features](https://optuna.readthedocs.io/en/stable/tutorial/10_key_features/005_visualization.html#visualization) to explore those trials further).


**Note:** I still need to check that this is indeed a good way to do this. Optuna now has its [optuna dashboard](https://github.com/optuna/optuna-dashboard), I haven't yet tried it and I am not sure if you can observe the trials there as they appear. If so, this is actually a good alternative. The advantage of Optuna dashboard is a nicer UI tailored specifically for exploration of optuna studies. The advantage of DVCLive is its simplicity (mainly the fact you don't need to set up an RDB first).

# 2. Pareto front AB Testing

Let's now imagine that the relationship of my objective function to the business use-case is complex and the objective function is multi-dimensional. For example, let's say I am building a recommender engine where I want to optimize both engagement time and click-through-rate but in fact I want to improve long-term customer retention which I don't see in my training data.

When I am not completely sure which values of my objective function are actually the best values, one thing I might do is some AB testing using trained models produced by some of the trials on the Pareto front of the objective function. Those are basically trials such that there are no trials which are strictly better than them in all dimensions of the objective function.

Since this can mean potentially quite a complex deployment of multiple models, I probably don't want to do that manually. Traceability and reproducibility is key here, since I want to make sure I actually can apply the lessons from the AB testing to further model development and a good way to ensure traceability is to use a GitOps approach to machine learning. DVC can track all trials as experiments. From those, I can pick (a subset of) the trials on the Pareto front , turn them into standalone branches and use automation tools such as GitHub actions and machine learning CD tools like MLEM to deploy these experiment branches for AB testing.

Now, when I use a multi-dimensional objective function, Optuna keeps track of the best trials which are accessible using the `.best_trials`  method of the optuna `study` object. What DVC adds here is the ability to split the hopeful trials into their own experiment branches, which makes traceability a lot easier and can help to keep a flat structure where there is a one-to-one correspondence between a specific commit in a git repository and a specific model in production.

dberenbaum:
Thanks @tibor-mach! My thoughts on each use case:

1. Human learning the meta-hyper-parameters

I think this is likely to be the most widespread usage since it's simplest. I know of at least one other user (@d-miketa) who uses both DVCLive and the Optuna dashboard together. As you mention, I assume the Optuna dashboard is better when needing some visualizations specific to Optuna and hyperparameter optimization. I think the reason to include DVCLive in that case is to connect the results back to your project and everything else you are tracking in Git and DVC. You could save the hyperparameters, metrics, and model artifacts for each trial so that you maintain a full history of every trial in your project.

If you are interested in exploring this use case, that's great! Otherwise, we will likely try to put together this use case anyway.

2. Pareto front AB Testing

I think it's more of a niche use case that wouldn't have as broad appeal but would really showcase the power of our tools in an advanced end-to-end scenario. I think it could also touch on model registry functionality that we are also developing.

TBH it feels like developing this use case would be pretty involved, so not sure if you feel like you have the time to work on it, but this is all up to you so happy to try to help you with it if you're interested.

tibor-mach:
@dberenbaum Ok, I actually ended up looking at it today already. I had some issues with logging artifacts.

I think there are two ways this can be approached. One is not to use the callback and simply build it all by yourself inside the objetive function using the DVCLive API. I think this is certainly necessary at some point if you want to customize things a lot.

However, I ended up modyfing the callback to enable this. There are some things about it I don't like which I mention in the [draft PR](https://github.com/iterative/dvclive/pull/548) 

There are a few more issues which are mentioned [here](https://github.com/tibor-mach/dvclive-optuna#issues). I am not sure how to fix the first one and while the second one could be approached exactly the same way as the artifact logging above, it would be quite clunky and at that point I would probably just decide against using the callback and use DVCLive inside the objective function directly.

@dberenbaum let me know what you think about this update. If I were to write about this I would basically suggest that for simple logging of metrics and (maybe) artifacts the users should use the callback but switch to actually writing custom scripts with DVCLive if they are attempting something more complex.

One issue I'd definitely like resoved is the one regarding logging the entire study as a part of dvc repro (see the issues I linked to). I am not sure if this has any easy fixes though.

dberenbaum:
@tibor-mach In https://github.com/tibor-mach/dvclive-optuna#issues, you mention:

> Running the dvc pipeline in order to log the Optuna study at the end of training clashes with logging of trials into separate experiments since the study object is only saved at the end of the hyperparameter training.

Great point. Can we save the study as we go as one of the outputs of each trial? That way, they would be more reproducible/resumable.

If there's no RDB backend configured, we could default to setting up a sqlite one and tracking that with DVC so it can be saved, resumed, and shared as needed.

tibor-mach:
@dberenbaum I suppose we could save the study the way you suggest but it is not very elegant - the study keeps adding trials as the hyperparameter search progresses so basically you end up saving a lot of strictly increasing subsets of the final study just to avoid the issue I mention here. The study already keeps track of the progress.

To be fair though, I suppose the advantage is that you can then resume the study from any trial (potentially with updated metaparameters) which might be useful sometimes...so let's see it as a feature for now. I guess I will proceed with putting together an example with the setup we discussed here. I will mention the optuna callback but then I will show how one can build something more custom (and complex) with the dvclive api.

I've been quite busy lately, but I think I should be able to make some progress next week.

dberenbaum:
Sounds good @tibor-mach. If you prefer to go directly to using the underlying ML framework instead of the Optuna callback, that's also fine. If the callback isn't adding any value, we can suggest people use the ML framework integrations instead and provide examples as needed.

tibor-mach:
@dberenbaum I am still thinking about what is the better approach (with or without the callback).

The thing is that if we want to progressively version the study object together with the trials, then it makes sense to use the callback (since it takes the study as a parameter) and in fact I couldn't figure out an easy way to do it otherwise. This has the following specific solutions, neither of which I like very much:

1. The study is not logged by dvc at all. That is not a good solution as the study captures a lot ofuseful information about how the hyperparameter optimization went (and can be used to restart the search from a specific point). The only advantage is that in this case the callback can be replace with a few lines of the dvclive api code inside the objective function.
2. The Optuna callback can be expanded with hardcoded user attribute names. Users can then log those, but they need to manually create the trial/study user attributes as a part of the objective function definition AND those user attributes have to match the names hardcoded in the callback...or the names could be made into optional parameters of the callback. Either way, this feels very inflexible and not modular at all...we'd be cramming things into the callback with very little flexibility.
3. Users will be expected to write the callback themselves with Iterative limiting themselves to providing a BaseOptunaCallback of sorts which logs a few carefully selected things that do not require any user-defined attributes. I think that the one written by @daavoo basically meets the criteria of such a BaseOptunaCallback (I'd just add study logging to it, maybe with an option to turn it off). Users could then write their own callback classes which inherit from the BaseOptunaCallback. This would require the users to write new logging methods in their CustomOptunaCallback and to add user attributes to the trial object in the objective function (or even to the study object if they decide to log something specific not accessible by the standard optuna `Study` attributes).

So...I dislike the third option the least. It is far from plug and play and many data scientists cannot be expected to be familiar with OOP (even though there is not a lot of it here). Still, with some examples provided, it does seem me like the least bad option (at least of those I could think of), so I think I'd go ahead with that. What's your opinion on this @dberenbaum  @daavoo ?

tibor-mach:
I looked at this a bit more and I guess that in order to log the study the only way to avoid having to write callbacks manually (as opposed to just using the DVCLive API) is to set up an RDB and save the study there (as you suggest @dberenbaum ).

It is a bit annoying since everything else can be done just fine without having to set up the database, but with the study it is either this or putting everything into a callback.

I'll have a look at using the RDB for this and then get back to you...

tibor-mach:
@dberenbaum I think [this](https://github.com/tibor-mach/dvclive-optuna/tree/custom_callback) is really the best way to go about it. It does require writing a custom callback and specifying user attributes, one only really needs to specify the paths for the serialized models and studies. 

This could even be replaced by sensible defaults like "<dvclive_dir>/model.pkl" which would eliminate the need to explicitly provide the paths. It felt like too much stuff happening under the hood though, so I decided against that. I also used inheritance from the DVCLiveCallback which handles metrics and parameters. Since most common artifacts to log are going to be the models and studies and since this really is not that much more extra code, I wonder if it might not make sense to actually include this in the original callback class after all...I set things up so that by default neither the artifact nor the model is logged so the users don't really have to worry about any user attributes if they don't want those.

Optionally, the user also can specify the names for the models and studies (might be useful for example as a way to track trial numbers) in user attributes, but they can just use the path stem as DVCLIve does in log_artifact already

By the way there is an issue with snake_case names of artifacts, for some reason the generated dvc.yaml does not accept them as valid names...not sure if this is a bug, but probably it is.

I also tried doing this without the callback [here](https://github.com/tibor-mach/dvclive-optuna/releases/tag/without_callback), but I am not very happy about that. The logged study is always going to be one step behind because it is logged from inside the objective function in this case and any workarounds are much uglier than actually using the callback and user attributes (I noticed that Optuna actually uses user attributes for trials and studies in some of their examples of custom callbacks, so it seems to be intended).

Anyway, let me know what you think. Personally, I would use this (the callback version) to write a short blogpost (I am using multi-objective optimization because logging models and studies at each trial is more useful in that case as multiple trials might yield interesting models and there is more likely a need for tweaking the study - it can be resumed from a specific trial with a different hyperparameter space for example). As I mentioned I think it might even potentially make sense to include these changes in the original callback, but that is up to you.

dberenbaum:
Thanks @tibor-mach! You mention above that the study captures a lot of useful information. Could we log other useful info about the study, like `study.best_params`, `study.best_value`, and the number of trials so far? Is there anything else that might be useful?

Do you find the callback saving the study more useful than doing it yourself without the callback? Optuna seems to push for saving to an RDB backend, and I'm not sure if we want to push a different approach, although saving to a file is nice in that you can easily revert, resume, and share with others without having to setup a database server. 

For model files, I'm even less sure about saving them from within the callback. Generally, we have moved away from trying to save the model files since it seems like it usually adds friction more than it helps, and it's often better for users to save the models using the framework's preferred method.

tibor-mach:
@dberenbaum

1. I think a nice feature that optuna provides is the [visualisation](https://optuna.readthedocs.io/en/stable/reference/visualization/index.html) module. All functions in that module take a study as input. I guess one could simply log all of the plots it generates directly with DVCLive, but sometimes you don't want all of them and simply logging the study is more compact. So this is why I think that logging the study itself is very useful
2. Optuna does seem to prefer using an RDB backend, I'd say it is mostly because of [Optuna dashboard](https://optuna-dashboard.readthedocs.io/en/latest/getting-started.html). But they [do not seem to actively discourage users from saving studies into files](https://github.com/optuna/optuna/issues/2678) and as you say, saving it as a file is a lot more simple and lightweight. I'd say both approaches are valid in different circumstances. Generally, I'd probably use the file approach in a small team, the RDB one in a larger project with a lot of contributors.
3. The main problem with not using the callback is that you cannot really log the study at the point where the current trial is finished and evaluated if you implement logging with DVCLive from inside the objective function. When you call optuna `study.optimize()`, it successively calls the objective function to evaluate the current trial. The trial is logged to the study before the objective function call but its status is RUNNING and its result is not logged yet. Once the objective function returns the metric(s) the status is changed to FINISHED and the metrics are logged. Then any callbacks you specified are called ([this](https://github.com/optuna/optuna/blob/30fabb90fafbb499536e5999f4caec2ffe0c2963/optuna/study/_optimize.py#L148-L175) is where it is implemented). ... so if you want the trial experiment to contain a study which includes the outcome of the latest trial and has it correctly logged as FINISHED, you need to log only after the objective function call and the only way to do that without forking optuna is through the callback.
4. As for saving model files - to me this is actually the main reason why I might use DVCLive with Optuna. Otherwise all the trials are stored quite neatly inside the study so if all you have left of them are their metadata, then I don't see much value of distributing them to DVC experiments. I'd rather use the optuna visualisations or even the optuna dashboard to explore them and perhaps still use DVC to log the final study and the optimization metaparameters...that way, the entire optimization process is reproducible, which the individual trial-experiments are not (they are only reproducible in the sense of running the entire optimization again). What can be (situationally) useful though is the option to quickly and easily apply any specific experiment to the workspace and play around with the model fitted in that specific trial-experiment. If individual trials are cheap and fast to train, this is not an issue, but if one trial takes a few hours to fit the model then it is nice to have them saved. It can also be nice to have the study object versioned at any trial so you can easily restart from there (although that is a but more niche as a use-case, I'd say). If all we use DVCLive and DVC experiments for is logging metadata, then I would argue that the optuna study already does that better, especially if you use the RDB backend.

d-miketa:
I’ll just add that the RDB backend also lets Optuna manage distributed studies, which is a significant advantage.

dberenbaum:
> If the callback isn't adding any value, we can suggest people use the ML framework integrations instead and provide examples as needed.

@tibor-mach Thanks for the insight. It feels like this may be the way to go for now given your thoughts. There is some value to the custom callback, but for the amount of effort it takes to understand it and write your own custom callback, it doesn't feel worth it if you are mostly interested in saving the model from each trial. You can save/resume with a sqlite database or other RDB, so unless you want to revert to an earlier point in the study (not sure I see a lot of use cases for that), it doesn't feel like the callback adds much value.

@d-miketa What is your take on saving metadata about the study (best trial params/metrics, study name, etc.). Is it valuable to you, or having a record of the individual trials is enough?

tibor-mach:
@dberenbaum I agree that reverting studies is a very niche use-case and doesn't make sense with most samplers you'd use in Optuna (which are Bayesian). The reason I even started with that was that I tried to find a way to run `dvc repro` and get both logged experiments and the final study as a dvc pipeline output, making the whole optimization reproducible. But perhaps reproducibility is not the main thing here now as this is still a use-case very much in a prototyping phase...in which case, I'd probably still use the callback made by @daavoo , because if I just log the artifacts without any description of the trial results, I will not have an easy way to find out which trials are interesting. Logging trial number would be ideal (and a parameter of the trial object, so this would be easily updated and would not need any extra work from the users.

@d-miketa Distributed studies are also kind of an interesting point from the perspective of DVC and reproducibility in general. When studies are not run sequentially they will be inherently non-reproducible (unless you use a very simple sampler which does not depend on previous trials, like grid search for example). I guess this is again fine in a prototyping stage whre you want to explore a large hyperparameter space fairly quickly but in a more stable production phase of development I would personally prefer switching to sequential optimization (and perhaps a smaller hyperparameter space) to be able to benefit from full reproducibility.

d-miketa:
@dberenbaum Hm… DVC can track a good deal of that. But the Optuna dashboard’s useful and valuable and that AFAIK requires a study RDBMS.

PR #364, opened 2022-11-15
frameworks: Call `live.end` at the end of training.

daavoo:
Ensure logic inside `live.end` is only called once.

This is mainly for Studio to get a "completed" event when using the frameworks


src/dvclive/live.py
```diff
@@ -241,12 +241,14 @@ def make_report(self):
                 open_file_in_browser(self.report_file)
 
     def end(self):
```
dberenbaum, line 243: Minor: Could you use the `@functools.cache` decorator on `def end(self):` instead of adding `self._ended`?

daavoo: Depends on https://github.com/iterative/dvclive/pull/364#issuecomment-1322159888

dberenbaum:
None of this is a blocker since I don't think anything breaks the basic live metrics workflow, but it would be good to have it to "unbreak" more complex scenarios.

> If I decouple the `post_to_studio(self, "done", logger)` from the `live._ended`, I think the snippet would behave as expected.

It's a little ugly but more likely to work the way users want (log summary/report always).

> I would expect the live metrics functionality to be only working inside the callback.

Can we also have something like this to enable the context manager to override other calls to `end`?

> ```diff
>     self.make_summary()
>     if self.report_mode == "studio":
>         if not self._ended:
>             if not post_to_studio(self, "done", logger):
>                 logger.warning("`post_to_studio` `done` event failed.")
>-           self._ended = True
>+           if not self._keep_alive: # set to True in __enter__
>+               self._ended = True
>     else:
>         self.make_report()
> ```

dberenbaum (approved):

PR #366, opened 2022-11-22
Create `DVC` experiment on `live.end`.

### LINKED ISSUE #310: Save results as DVC experiment
Depends on https://github.com/iterative/dvc/issues/6746.

If DVC is installed, DVCLive should be able to create an experiment ref. This should be possible by either:
1. Making `Live` a context manager invoked by `with dvclive.Live() as live:`.
2. Adding a `Live.end` method.

For the first iteration, having one or the other of these approaches is fine, but ultimately we might want both.

### LINKED ISSUE #311: Configure DVC experiment from DVCLive

daavoo:
Closes #310 
Closes #311 

Requires https://github.com/iterative/dvc/pull/8599 and https://github.com/iterative/dvc/pull/8529


src/dvclive/utils.py
```diff
@@ -127,3 +128,22 @@ def matplotlib_installed() -> bool:
...
...tamed', 'tamer', 'tangy', 'taped', 'tardy', 'tarot', 'tarry', 'tasty', 'tatty', 'taunt', 'tawie', 'teary', 'techy', 'telic', 'tenor', 'tense', 'tenth', 'tenty', 'tepid', 'terse', 'testy', 'third', 'tidal', 'tight', 'tiled', 'timid', 'tinct', 'tined', 'tippy', 'tipsy', 'tonal', 'toned', 'tonic', 'toric', 'total', 'tough', 'toxic', 'trade', 'treed', 'treen', 'trial', 'truer', 'tubal', 'tubby', 'tumid', 'tuned', 'tutti', 'twill', 'typal', 'typed', 'typic', 'umber', 'unapt', 'unbid', 'uncut', 'undue', 'undug', 'unfed', 'unfit', 'union', 'unlet', 'unmet', 'unwed', 'unwet', 'upper', 'upset', 'urban', 'utile', 'uveal', 'vagal', 'valid', 'vapid', 'varus', 'vatic', 'veiny', 'vital', 'vivid', 'vocal', 'vogie', 'volar', 'vying', 'wacky', 'wally', 'waney', 'warty', 'washy', 'waspy', 'waste', 'waugh', 'waxen', 'webby', 'wedgy', 'weeny', 'weepy', 'weest', 'weird', 'welsh', 'wersh', 'whist', 'white', 'whity', 'whole', 'wider', 'wight', 'winey', 'wired', 'wised', 'wiser', 'withy', 'wonky', 'woods', 'woozy', 'world', 'wormy', 'worse', 'worst', 'woven', 'wrath', 'wrier', 'wrong', 'wroth', 'xeric', 'yarer', 'yolky', 'young', 'yucky', 'yummy', 'zesty', 'zingy', 'zinky', 'zippy', 'zonal')  # noqa: E501
```
dberenbaum, line 145: ADJETIVES -> ADJECTIVES

Is this needed or a nice to have based on other naming discussions?

src/dvclive/live.py
```diff
@@ -247,6 +248,12 @@ def end(self):
                 logger.warning("`post_to_studio` `done` event failed.")
         else:
             self.make_report()
+        if dvc_installed():
+            from dvc.repo import Repo
+
+            from dvclive.utils import random_name
+
+            Repo().experiments.save(name=random_name())
```
dberenbaum, line 256: When running this for the first time, either DVCLive or DVC needs to add the DVCLive outputs to Git, right?

daavoo: Yes. I am giving it a try on DVC side first

src/dvclive/utils.py
dberenbaum: `Repo().experiments.save()` doesn't require `name`, correct? I'm wondering why we need to generate a name in DVCLive. I think we have discussed some reasons it could be useful, but I'm missing why it's needed in this PR.

daavoo: We discussed it on DVC `exp save` P.R:

https://github.com/iterative/dvc/pull/8408#discussion_r1014642023

dberenbaum: 🤦 👌

dberenbaum: @shcheklein We can come back and change this, so let's not block the PR, but you might be interested that we are moving towards auto-generating names in DVCLive.

shcheklein: it makes sense! we need to be sure that we show git hash nearby everywhere (or even first) so that ppl can copy paste and do `git` / `dvc` commands on it

it doesn't change / affect the need in VS Code to generate names in a similar way when ppl run `dvc exp run`

also, qq - does it prevent / check for duplicates (same name already exists?) ... it might be useful to include some random hash in the name in our case as well, so that we 100% sure that it's globally unique, wdyt? (if we are doing this already, please disregard the comment :) )

daavoo: > it makes sense! we need to be sure that we show git hash nearby everywhere (or even first) so that ppl can copy paste and do git / dvc commands on it

The commit hash is still there.

> does it prevent / check for duplicates (same name already exists?) ... it might be useful to include some random hash in the name in our case as well, so that we 100% sure that it's globally unique, wdyt? (if we are doing this already, please disregard the comment :) )

Just retries when DVC fails with `ExperimentExists` (this is an easy fast check). The probability of collision is <1e-6 anyway.

I feel that, realistically, there is no need to include hash in the generated name. The number of possible combinations in a Repo (> 1e6) is way over what would break any DVC exp command

src/dvclive/live.py
```diff
@@ -41,95 +41,134 @@ def __init__(
...
+    def _init_dvc(self):
+        self._dvc_repo = get_dvc_repo()
+        if os.getenv(env.DVC_EXP_BASELINE_REV, None):
+            # `dvc exp` execution
+            self._baseline_rev = os.getenv(env.DVC_EXP_BASELINE_REV, "")
+            self._exp_name = os.getenv(env.DVC_EXP_NAME, "")
+            self._inside_dvc_exp = True
+        else:
+            # `Python Only` execution
+            # TODO: How to handle `dvc repro` execution?
```
daavoo, line 102: @dberenbaum what do you think we should do if DVCLive is used inside a dvc pipeline that has been executed with `dvc repro`?

I think we should skip the experiment creation.

tests/test_dvc.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def test_make_dvcyaml(tmp_dir):
+    live = Live()
+    make_dvcyaml(live)
+
+    assert load_yaml(live.dvc_file) == {
+        "metrics": ["metrics.json"],
+        "params": ["params.yaml"],
+        "plots": ["plots"],
```
daavoo, line 25: Follow up: configure sklearn plots

src/dvclive/live.py
dberenbaum: Yes, let's skip it. We may still want to call `make_dvcyaml`, but we can skip that also for now if it simplifies things.

src/dvclive/dvc.py
```diff
@@ -46,3 +53,58 @@ def make_checkpoint():
...
+        try:
+            dvc_repo.experiments._validate_new_ref(exp_ref)
+        except ExperimentExistsError:
+            continue
+
+        return name
+
+
+def make_dvcyaml(live):
+    if not os.path.exists(live.dvc_file):
```
dberenbaum, line 102: Will `dvc.yaml` get deleted each time a new live instance is created? Maybe we need some option to configure whether to overwrite it. In some cases, I may log new stuff and want to overwrite, but in others I might have added some custom config that I want to keep.

src/dvclive/live.py
```diff
@@ -41,95 +41,135 @@ def __init__(
...
+            if self._dvc_repo is not None:
+                self._baseline_rev = self._dvc_repo.scm.get_rev()
+                self._exp_name = random_exp_name(
+                    self._dvc_repo, self._baseline_rev
+                )
+                make_dvcyaml(self)
```
dberenbaum, line 108: I'm not as sure that we should skip `make_dvcyaml`, but we can follow up on this separate from this PR.

src/dvclive/dvc.py
daavoo: > Will dvc.yaml get deleted each time a new live instance is created? 

It should not be deleted and it should preserve the manual modifications from users.

> In some cases, I may log new stuff and want to overwrite, but in others I might have added some custom config that I want to keep.

What would be overwritten? The current version is extremely simple as it doesn't configure sklearn plots (added a follow-up https://github.com/iterative/dvclive/issues/371)

src/dvclive/live.py
dberenbaum: Another follow up: pass an experiment name to use.

dberenbaum: And another: consider whether to include the dvclive user script/notebook in the tracked files.

daavoo: > And another: consider whether to include the dvclive user script/notebook in the tracked files.

Do you mean to include it in the `include_tracked` list passed to experiments save or other thing?

dberenbaum:
Demo from @daavoo:

https://user-images.githubusercontent.com/2308172/204875763-cf0a1a6d-7146-4de5-8e04-4da89e668803.mp4

src/dvclive/dvc.py
dberenbaum: Sounds good. Not a blocker.

> What would be overwritten? 

As you say, it will include sklearn plots config, which seems like the most complicated scenario, so let's discuss there.

src/dvclive/live.py
dberenbaum: Yes, whether to include it in `include_tracked`. I'm not sure it's needed, so it was more of a discussion point than a request. There's something clean about the default being to only save the live dir.

src/dvclive/dvc.py
skshetry: It should probably just write to the root of the repo, not to its own file by default. It is the dvc way, and this is against the principle of having one file to rule them all.

dberenbaum: 🤔 But the dvclive way is to create a self-contained folder 😄 . IMHO dvclive should not write to the user's root `dvc.yaml` file, because it it makes it hard to share only the dvclive output, and it mixes machine-generated config with user-generated config. Also, if there is some conflict between the existing config and what dvclive generates, I would rather have dvclive run successfully and be able to later debug the conflicts between the files.

src/dvclive/lightning.py
```diff
@@ -17,11 +17,16 @@ def __init__(
         dir: Optional[str] = None,  # noqa pylint: disable=redefined-builtin
         resume: bool = False,
         report: Optional[str] = "auto",
+        save_dvc_exp: bool = False,
```
daavoo, line 20: @dberenbaum named the option like this. wdyt?

dberenbaum: I don't know if it makes sense for the `dvc.yaml` saving part and whether those should be coupled together, but I think it's fine for now.

daavoo: I though your concerns also applied to both the git ref and the `dvc.yaml` creation

Issue #374, opened 2022-11-30
metrics: consolidate into one file


dberenbaum:
`live/plots/metrics` ends up containing too many individual files. For example, using the lightning logger, I ended up with this tsv file:

```
step	epoch
0	0
1	0
2	0
3	0
4	0
5	0
...
```

By contrast, lightning has a built-in csv logger that combines all the metrics into one csv file. Can dvclive either use that logger or write its own outputs to one csv/tsv file?

PR #380, opened 2022-12-01
frameworks: Don't log `epoch`.

daavoo:


dberenbaum:
I think I'm missing the context here @daavoo. Why not log epoch?

daavoo:
> I think I'm missing the context here @daavoo. Why not log epoch?

I felt like it is just redundant with the `step`. 
What is the use case of the `epoch` linear plot which is a constant `y=x`?

dberenbaum:
Often the step corresponds to something other than a full epoch (like one batch iteration). I would still want the epoch recorded separately in that case, although I wouldn't want to plot it. I would rather keep recording it and address plotting it elsewhere. If we consolidate the tsv output like in #374, we will have to specify which columns to plot, and we could ignore it there. What do you think?

daavoo:
> Often the step corresponds to something other than a full epoch (like one batch iteration)

From the integrations we provide, I think is only true for PyTorch Lightning.

> I would rather keep recording it and address plotting it elsewhere. If we consolidate the tsv output like in #374, we will have to specify which columns to plot, and we could ignore it there. What do you think?

I still don't undestand how the epoch signal can be used in practice nowadays but removing is not critical to me, so we can wait.

dberenbaum:
🤔 It's possible to log at the batch step level in other frameworks (https://www.tensorflow.org/guide/keras/custom_callback, https://huggingface.co/docs/transformers/v4.25.1/en/main_classes/callback#transformers.TrainerCallback.on_step_end), but this seems too granular, so I agree it's unhelpful.

We could leave it for lightning and drop it for others.

Should/can we rename step to epoch for those other frameworks since technically they tend to use step to refer to batch iterations and epoch to refer to full dataset iterations?

PR #383, opened 2022-12-02
lightning: record actual training step

dberenbaum:
Lightning was recording one step per logging iteration, but by default that logs every 50 steps (https://pytorch-lightning.readthedocs.io/en/stable/extensions/logging.html#logging-frequency). This leaves you with 3 different values between epoch, lightning step, and dvclive step.

### DIFF
--- a/tests/test_lightning.py
+++ b/tests/test_lightning.py
@@ -145,6 +145,7 @@ def test_lightning_kwargs(tmp_dir):
     assert os.path.exists("dir")
     assert os.path.exists("dir/report.md")
 
+
 def test_lightning_steps(tmp_dir):
     model = LitXOR()
     # Handle kwargs passed to Live.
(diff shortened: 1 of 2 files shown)


daavoo (approved):

tests/test_lightning.py
```diff
@@ -143,3 +143,28 @@ def test_lightning_kwargs(tmp_dir):
...
+    assert latest["step"] == 8
+    assert latest["epoch"] == 1
+
+    scalars = os.path.join(
+        dvclive_logger.experiment.plots_dir, Metric.subfolder
+    )
+    epoch_loss = history[os.path.join(scalars, "train", "epoch", "loss.tsv")]
+    step_loss = history[os.path.join(scalars, "train", "step", "loss.tsv")]
+    assert len(epoch_loss) == 2
+    assert len(step_loss) == 4
```
daavoo, line 170: I have no idea what are the differences between these 2. 
I assume `epoch/metric` contains the mean value of the whole epoch 🤷

daavoo (approved):

tests/test_lightning.py
daavoo: Found https://pytorch-lightning.readthedocs.io/en/stable/extensions/logging.html#automatic-logging

Issue #381, opened 2022-12-01
make_dvcyaml: always save dvc.yaml


dberenbaum:
After playing around with `save_dvc_exp` a bit, I think it should always call `make_dvcyaml` even when run with `dvc exp run` or `dvc repro`. It makes the transition to using these smoother since you can keep the existing `dvc.yaml` to configure the metrics and plots.

Edit: I think because my `dvclive/dvc.yaml` file was untracked, it wasn't getting saved once I turned off `save_dvc_exp`. Regardless, I think we should overwrite it on each run to ensure it exists and is up to date.

dberenbaum:
Some notes:

* If there is a `dvclive/dvc.yaml` but no `dvclive/params.yaml`, `dvc.api.params_show()` will complain. Can dvclive add/drop the params section in `dvclive/dvc.yaml` depending on whether `dvclive/params.yaml` exists?
* If I already have a root `dvc.yaml` with stage metrics and plots outputs and generate `dvclive/dvc.yaml`, it will duplicate plots but won't break anything.

If we can make params work, it might make sense to write out `dvclive/dvc.yaml` even when `save_dvc_exp=False`.

dberenbaum:
Another scenario to consider: if I run `dvc exp run` with checkpoints but without metrics/plots configured, it would be useful to call `make_dvcyaml` at each step to get live metrics/plots.

dberenbaum:
I plan to play around with this.

dberenbaum:
@daavoo Added two possibilities in #384 and #385. I'm happy to go with either option for now.

Both always write out `dvc.yaml`. I don't see much downside here. Even if dvc isn't installed, it means I could go back and see metrics, plots, and params if I later install it and add it to my repo.

Both also write out a new `dvc.yaml` at least for each live instance, since it's likely the format will change over time, and this ensures the updates get propagated to the file.

dberenbaum:
Reopening because this behavior was deemed too aggressive in some cases. For now, we can revert this behavior.

Comment from @shcheklein:

> we are generating evaluation:dvc.yaml even when there is a dvc.yaml already in the root. This causes strange effects (rendering of extra plots, some of the one single value (for metrics, since we dump them into plots as well). And at the end - everything is good. I think it’s too noisy / not expected

Let's please keep thinking about how this should behave. I'd still like to be able to generate this when using pipelines so that I don't have to manually configure the plots, metrics, and params myself.

dberenbaum:
[...]

Issue #371, opened 2022-11-30
dvc: Configure sklearn plots in generated `dvc.yaml`.


daavoo:
Follow up: configure sklearn plots

_Originally posted by @daavoo in https://github.com/iterative/dvclive/pull/366#discussion_r1036220642_

mattseddon:
👍🏻 

I just committed the auto-generated `dvclive/dvc.yaml` against the VS Code demo project ([here](https://github.com/iterative/vscode-dvc-demo/pull/27/files)) and ended up with this:

<img width="671" alt="image" src="https://user-images.githubusercontent.com/37993418/206330660-f1542b70-c398-4475-a87c-ac90a04754e7.png">


Not sure why the file was generated in an existing project. Perhaps because I had never run an experiment against that clone?

PR #385, opened 2022-12-02
write out dvcyaml at each step with relevant outputs

dberenbaum:
Slightly more complicated approach to #381.

Summary: write `dvc.yaml` at each call of `live.next_step` or `live.end` with only sections for outputs that exist (for example, skip `params` section in `dvc.yaml` if no params were logged).

Notes on this approach:
1. Everything works so far even when transitioning to pipelines.
2. Could be more flexible when trying to implement things like #371. Not sure how long it will make sense to keep `dvc.yaml` static.

### DIFF
--- a/tests/test_dvc.py
+++ b/tests/test_dvc.py
@@ -21,6 +21,20 @@ def test_make_dvcyaml(tmp_dir):
     live = Live()
     make_dvcyaml(live)
 
+    assert load_yaml(live.dvc_file) == {}
+
+    live = Live()
+    make_dvcyaml(live)
+    live.log_param("foo", 1)
+    live.next_step()
+
+    assert load_yaml(live.dvc_file) == {
+        "params": ["params.yaml"],
+    }
+
+    live.log_metric("bar", 2)
+    live.end()
+
     assert load_yaml(live.dvc_file) == {
         "metrics": ["metrics.json"],
         "params": ["params.yaml"],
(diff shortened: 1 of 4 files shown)


tests/test_dvc.py
```diff
@@ -21,6 +21,20 @@ def test_make_dvcyaml(tmp_dir):
     live = Live()
     make_dvcyaml(live)
 
+    assert load_yaml(live.dvc_file) == {}
+
+    live = Live()
+    make_dvcyaml(live)
```
daavoo, line 27: Is there a reason for this call?

daavoo (approved):
I prefer this implementation over #384 . It is not much complicated, in comparison, and both points you mention are benefits.

Can't really find downsides beyond the hypothetical use case where the user wants to manually modify the `dvc.yaml` (and preserve the modifications). 

We could always introduce an option to toggle `make_dvcyaml` if someone asks for it.

dberenbaum:
> Can't really find downsides beyond the hypothetical use case where the user wants to manually modify the `dvc.yaml` (and preserve the modifications).

The more I use it and think about it, the more I think that we are better off suggesting that user's create a new plot in the root `dvc.yaml` file and never manually touch the dvclive outputs.

tests/test_dvc.py
dberenbaum, line 27:
```suggestion
    live = Live()
```

dberenbaum:
@skshetry PTAL here. We are writing to `dvclive/dvc.yaml` at each step of training. That means that top-level plots, metrics, and params could be updated multiple times during execution of single stage. Does it make sense? Do you see any concerns?

PR #388, opened 2022-12-05
lightning: log params

dberenbaum:
Adds lightning param logging, which requires user setup like:

```
class MyLightningModule(LightningModule):
    def __init__(self, myparam):
        self.save_hyperparameters()
```

This is the lightning way 🤷 (see https://forums.pytorchlightning.ai/t/how-to-save-hparams-when-not-provided-as-argument-apparently-assigning-to-hparams-is-not-recomended/339/3)


src/dvclive/lightning.py
```diff
@@ -38,7 +38,7 @@ def name(self):
 
     @rank_zero_only
     def log_hyperparams(self, params, *args, **kwargs):
-        pass
+        self.experiment.log_params(params)
```
daavoo, line 41: Apparently there are some utility functions that the other loggers use:

https://github.com/Lightning-AI/lightning/blob/e7917492cd0361c3d8c675f3f6e1e2e54e831cc8/src/pytorch_lightning/loggers/wandb.py#L421-L426

We don't need `_flatten_dict` but I guess the other 2 won't hurt

dberenbaum: Done. Didn't add anything to tests since I wasn't sure what it would add beyond testing pytorch-lightning utilities, but LMK if you think tests are needed.

daavoo (approved):

PR #391, opened 2022-12-07
studio: Remove `STUDIO_REPO_URL` requirement.

### LINKED ISSUE #368: Remove `STUDIO_REPO_URL` requirement.
Try to infer it from the Git config (remote URL of the current tracking branch)

daavoo:
Closes #368

### DIFF
--- a/src/dvclive/studio.py
+++ b/src/dvclive/studio.py
@@ -1,8 +1,11 @@
 # pylint: disable=protected-access
+import logging
 from os import getenv
 
 from dvclive.env import STUDIO_ENDPOINT
 from dvclive.utils import parse_metrics
+
+logger = logging.getLogger(__name__)
 
 
 def _get_unsent_datapoints(plot, latest_step):
@@ -41,7 +44,7 @@ def _get_updates(live):
     return metrics, plots
 
 
-def post_to_studio(live, event_type, logger) -> bool:
+def post_to_studio(live, event_type) -> bool:
     import requests
     from requests.exceptions import RequestException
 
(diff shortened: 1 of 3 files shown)


src/dvclive/dvc.py
```diff
@@ -59,14 +59,12 @@ def make_checkpoint():
 def get_dvc_repo():
     # noqa pylint: disable=unused-import
     try:
-        import dvc  # noqa: F401
+        from dvc.exceptions import NotDvcRepoError
+        from dvc.repo import Repo
+        from dvc.scm import SCMError
```
daavoo, line 64: Unrelated, sorry for including it

src/dvclive/studio.py
```diff
@@ -81,3 +82,11 @@ def post_to_studio(live, event_type, logger) -> bool:
...
 
     return response.status_code == 200
+
+
+def get_studio_repo_url(git_repo):
+    # https://github.com/iterative/studio/blob/main/backend/repos/models/repo.py#L67
+    remote_name = git_repo.active_branch.tracking_branch().remote_name
+    remote_url = git_repo.remotes[remote_name].url
+    parsed = urlparse(remote_url.replace(".git", ""))
+    return f"{parsed.netloc}:{parsed.path.lstrip('/')}"
```
mvshmakov, line 92: Please have look at [this](https://github.com/iterative/studio/blob/develop/backend/repos/urltools.py) file. We have [these](https://github.com/iterative/studio/blob/develop/backend/repos/urltools.py#L13-L17) types of remotes + `custom-gitlab` (can be handled with `STUDIO_REPO_URL` for now). We strip down `.com` from them.

Overall suggestion, let's validate remotes against whitelisted URLs (main GH/GL/BB SaaS deployments). Otherwise, we should fail and ask the user to explicitly provide `STUDIO_REPO_URL`.

@nimdraugsael let us know if it makes sense.

mvshmakov (changes requested):

src/dvclive/studio.py
daavoo: > Please take a look at [this](https://github.com/iterative/studio/blob/develop/backend/repos/urltools.py) file. We have [these](https://github.com/iterative/studio/blob/develop/backend/repos/urltools.py#L13-L17) types of remotes + custom-gitlab (can be handled with STUDIO_REPO_URL for now). We strip down .com from them.

Skimmed through the file and missed some parts.
Added tests for the 3 providers and ssh.

> let's validate remotes against allowlisted URLs (main GH/GL/BB SaaS deployments). Otherwise, we should fail and ask the user to explicitly provide STUDIO_REPO_URL. If we don't do this, then gitlab.freedesktop.org's hosted repo will become gitlab.freedesktop.org:iterative/dvclive-example, which is invalid from Studio BE PoV.

Not sure I follow why we should duplicate that sort of validation on the DVCLive side. There are other cases where we rely on Studio backend to handle the invalid request and return the appropriate response (invalid token, non existing repo, etc).
We already handle those responses from studio and log them in debug mode.

mvshmakov: > Not sure I follow why we should duplicate that sort of validation on the DVCLive side

I believe currently we'll receive ["401 Permission denied"](https://github.com/iterative/studio/blob/develop/backend/live/views.py#L57) from the Studio backend on an invalid URL, which does not explain what is the problem and how to mitigate it. BE does not validate such URLs anyhow, except [checking that it is a string](https://github.com/iterative/studio/blob/develop/backend/live/views.py#L32).

Validating these things on the DVCLive side makes us fail fast and say which env variable the user should modify to make it work. Making Studio BE know that DVCLive uses the `STUDIO_REPO_URL` env variable and where are the docs for it to build a correct error message does not seem to be the best solution IMO. We may have another consumer of Studio's live ingestion endpoint and for them, the message `Please modify STUDIO_REPO_URL as ${docsLink}` may not make a lot of sense as `STUDIO_REPO_URL` exists only in the context of DVCLive.

daavoo: Added some validation and warnings. When docs are ready I will update to include the link in the message

daavoo: Still feel like Studio validation could give more info beyond Permission denied. If I understand correctly, passing a non existing `repo_url` will return the same error as passing a valid `repo_url` with an invalid token.

mvshmakov: I agree with you, it should be done on both sides. I've created https://github.com/iterative/studio/issues/4736 to iterate on it.

mvshmakov: > When docs are ready I will update to include the link in the message

Do we need an issue for it to not forget/add it to our [shared checklist](https://github.com/iterative/studio/issues/4666)?

mvshmakov (approved):
Nice work, It makes more sense to me now, thanks for addressing concerns!

daavoo:
> @daavoo Not a blocker, but this is a little noisy right now, raising a warning whenever dvclive is initialized. What is the setup intended to look like to enable studio reporting? Should studio be considered a report type?

What warning are you seeing? 
I think the one you are seeing is not coming to this P.R. but I have also identified that it is being shown unnecessarily

dberenbaum:
> What warning are you seeing?

WARNING:dvclive.studio:You can try manually setting the Studio repo url using the environment variable `STUDIO_REPO_URL`.

PR #392, opened 2022-12-09
Create signal file whenever DVCLive is running a non-DVC experiment (VS Code)

mattseddon:

----

There was a behaviour change shipped in the VS Code extension a few days ago where we no longer auto-select experiments for plotting by default. We (I) changed the behaviour to only auto-select experiments that have been run inside of the current session. Otherwise, we leave the selection up to the user (and/or the previous session). In order to do this I changed the extension's internal mechanics to rely on the `executor` & `status` fields inside of the `exp show` data. These fields are not populated when DVCLive is running outside of the experiment context and using `exp save` to generate the final experiment data (I've looked at the code and can see this would be a difficult change). 

All of that leads to the following behaviour for people being onboarded through the new route of instrumenting their code with DVCLive:

https://user-images.githubusercontent.com/37993418/206624828-7bc29465-f97c-4738-9229-4b6662969d97.mov

I had a long conversation with @dberenbaum this morning (original [here](https://iterativeai.slack.com/archives/C01APS0FHDM/p1670530599513299)) about this subject and we talked about patching the behaviour. I looked into the code (DVC, DVCLive and the extension) and decided that writing a signal file (🤮) was probably the sanest thing to do. Here is what the behaviour will look like after this is shipped and I've made a small update in the extension:

https://user-images.githubusercontent.com/37993418/206626354-663bc4e7-022e-49a4-a2fd-acdf401612c9.mov

LMK what you think.

Note: My Python skills are rusty and I'm not precious about making any required changes to the code. I also had a hard time getting the project set up to run locally (ended up deleting dependencies and tests in order to lint/test what I've changed). Presumably, the difficulty was associated with me being on M1.

### DIFF
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -17,6 +17,15 @@ def _dvc_dir(dirname):
     return os.path.join(dirname, ".dvc")
 
 
+def _dvc_exps_run_dir(dirname: str) -> str:
+    return os.path.join(dirname, ".dvc", "tmp", "exps", "run")
+
+
+def _dvclive_only_signal_file(root_dir: str) -> str:
+    dvc_exps_run_dir = _dvc_exps_run_dir(root_dir)
+    return os.path.join(dvc_exps_run_dir, "DVCLIVE_ONLY")
+
+
 def _find_dvc_root(root=None):
     if not root:
         root = os.getcwd()
@@ -36,8 +45,18 @@ def _find_dvc_root(root=None):
     return None
 
 
-def make_checkpoint():
+def _write_file(file: str):
     import builtins
+
+    with builtins.open(file, "w", encoding="utf-8") as fobj:
+        # NOTE: force flushing/writing empty file to disk, otherwise when
+        # run in certain contexts (pytest) file may not actually be written
+        fobj.write("")
+        fobj.flush()
+        os.fsync(fobj.fileno())
+
+
+def make_checkpoint():
     from time import sleep
 
     root_dir = _find_dvc_root()
@@ -46,12 +65,8 @@ def make_checkpoint():
 
     signal_file = os.path.join(root_dir, ".dvc", "tmp", env.DVC_CHECKPOINT)
 
-    with builtins.open(signal_file, "w", encoding="utf-8") as fobj:
-        # NOTE: force flushing/writing empty file to disk, otherwise when
-        # run in certain contexts (pytest) file may not actually be written
-        fobj.write("")
-        fobj.flush()
-        os.fsync(fobj.fileno())
+    _write_file(signal_file)
+
     while os.path.exists(signal_file):
         sleep(_CHECKPOINT_SLEEP)
 
@@ -108,3 +123,33 @@ def make_dvcyaml(live):
     if live._metrics or live._plots or live._images:
         dvcyaml["plots"] = [os.path.relpath(live.plots_dir, live.dir)]
     dump_yaml(dvcyaml, live.dvc_file)
+
+
+def mark_dvclive_only_started():
+    """
+    Signal DVC VS Code extension that
+    an experiment is running in the workspace.
+    """
+    root_dir = _find_dvc_root()
+    if not root_dir:
+        return
+
+    exp_run_dir = _dvc_exps_run_dir(root_dir)
+    os.makedirs(exp_run_dir, exist_ok=True)
+
+    signal_file = _dvclive_only_signal_file(root_dir)
+
+    _write_file(signal_file)
+
+
+def mark_dvclive_only_ended():
+    root_dir = _find_dvc_root()
+    if not root_dir:
+        return
+
+    signal_file = _dvclive_only_signal_file(root_dir)
+
+    if not os.path.exists(signal_file):
+        return
+
+    os.remove(signal_file)
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -46,13 +46,13 @@ def _find_dvc_root(root=None):
     return None
 
 
-def _write_file(file: str):
+def _write_file(file: str, contents=""):
     import builtins
 
     with builtins.open(file, "w", encoding="utf-8") as fobj:
         # NOTE: force flushing/writing empty file to disk, otherwise when
         # run in certain contexts (pytest) file may not actually be written
-        fobj.write("")
+        fobj.write(str(contents))
         fobj.flush()
         os.fsync(fobj.fileno())
 
@@ -156,7 +156,7 @@ def mark_dvclive_only_started():
 
     signal_file = _dvclive_only_signal_file(root_dir)
 
-    _write_file(signal_file)
+    _write_file(signal_file, os.getpid())
 
 
 def mark_dvclive_only_ended():
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -8,7 +8,14 @@ from typing import Any, Dict, List, Optional, Set, Union
 from ruamel.yaml.representer import RepresenterError
 
 from . import env
-from .dvc import get_dvc_repo, make_checkpoint, make_dvcyaml, random_exp_name
+from .dvc import (
+    get_dvc_repo,
+    make_checkpoint,
+    make_dvcyaml,
+    mark_dvclive_only_ended,
+    mark_dvclive_only_started,
+    random_exp_name,
+)
 from .error import (
     InvalidDataTypeError,
     InvalidParameterTypeError,
@@ -115,6 +122,7 @@ class Live:
                     self._exp_name = random_exp_name(
                         self._dvc_repo, self._baseline_rev
                     )
+                    mark_dvclive_only_started()
 
     def _init_studio(self):
         if not self._dvc_repo:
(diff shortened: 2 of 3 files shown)


src/dvclive/dvc.py
```diff
@@ -108,3 +123,33 @@ def make_dvcyaml(live):
     if live._metrics or live._plots or live._images:
         dvcyaml["plots"] = [os.path.relpath(live.plots_dir, live.dir)]
     dump_yaml(dvcyaml, live.dvc_file)
+
+
+def mark_dvclive_only_started():
+    """
+    Signal DVC VS Code extension that
+    an experiment is running in the workspace.
+    """
```
mattseddon, line 132: Probably an unnecessary comment, LMK if it can be improved or deleted.

src/dvclive/dvc.py
```diff
@@ -108,3 +123,33 @@ def make_dvcyaml(live):
     if live._metrics or live._plots or live._images:
         dvcyaml["plots"] = [os.path.relpath(live.plots_dir, live.dir)]
     dump_yaml(dvcyaml, live.dvc_file)
+
+
+def mark_dvclive_only_started():
```
mattseddon, line 128: [Q] Are there better names for these functions `mark_dvclive_only_started` & `mark_dvclive_only_ended`?

shcheklein:
To clarify - it's not (only) about auto selection - it's about detecting the "running" state in the workspace? What does `dvc exp show` return in this case, does it detect that an experiment is running?

shcheklein:
Would it be easier to watch `dvclive` by default and provide an option to setup its location if people change it?

mattseddon:
> To clarify - it's not (only) about auto selection - it's about detecting the "running" state in the workspace? What does `dvc exp show` return in this case, does it detect that an experiment is running?

`dvc exp show` does not detect an experiment is running.

daavoo (approved):
Implementation-wise, nothing against the P.R. I am ok with merging it

I have a question about alternatives, though. 

Given that this only targets the `DVCLive Only` scenario:

Q) Does VScode parse `dvc.yaml` files for anything? 
If yes, would it make sense to detect that there are no stages and just skip this new optimization for auto-selecting experiments?

> it's about detecting the "running" state in the workspace? 

What does the "running" state mean? 
Without this P.R. I am already seeing the workspace row being updated and If I select workspace, I see the plot updates. What additional "running" signal is needed?

src/dvclive/dvc.py
daavoo: Probably 😅 but It is ok for me to merge as it is

dberenbaum:
@daavoo Let's consider this a release blocker for dvclive because we need some solution to the plots auto-selection in vs code.

@mattseddon Is this better than reverting the recent vs code changes until we come up with something better? That's still not clear to me, but I will trust whatever you and @daavoo agree here because a quick fix is most important right now.

shcheklein:
> I'm not sure it's needed or even expected to have a spinner when running some dvclive code outside a dvc pipeline.

if we don't know that something is running, how do we expect to keep updating plots though? I think I'm missing something ...

> I think this also still leaves us without a solution for auto-selecting plots with non-dvclive dvc exp save

auto-selecting a new experiment when it appears in the table can be done I think, it can annoy people if they already have workspace selected though - they will see a duplicate. So, it's not clear in this case if it's valuable since they could have seen plots already, right?

----

Implementation - the biggest concern is that this file won't disappear if the process dies and users will have to deal with it? I think we have in DVC file-based locks that are being unlocked if process dies. I would take a look there.

dberenbaum:
> if we don't know that something is running, how do we expect to keep updating plots though? I think I'm missing something ...

I feel the same way for different reasons 😅. There's something I'm missing here. The plots in the workspace are already being updated.

https://user-images.githubusercontent.com/2308172/206819223-7832ce2b-f75c-4f47-acc7-8bc1d576b4ad.mov

I guess @mattseddon can answer how it works (file watchers?). So I don't follow why this signal file is needed?

> auto-selecting a new experiment when it appears in the table can be done I think, it can annoy people if they already have workspace selected though - they will see a duplicate. So, it's not clear in this case if it's valuable since they could have seen plots already, right?

The problem is that if I run a 2nd experiment, the 1st one disappears because it was never auto-selected:

https://user-images.githubusercontent.com/2308172/206819516-90abe521-304b-47da-8bc2-61be3b006d5a.mov

Compare that to the behavior before 0.5.24 of the extension:


https://user-images.githubusercontent.com/2308172/206819816-06cceb69-08e0-49a4-b974-2141fc8b8b69.mov

shcheklein:
> I guess @mattseddon can answer how it works (file watchers?). So I don't follow why this signal file is needed?

You had to select (Add an experiment) button to make it work. I think the idea here is that it should be automatic. We can't expect workspace to be always selected, at least not at the moment. 

> Compare that to the behavior before 0.5.24 of the extension:

do you have exp save in that case? actually in both cases? what do you see in the table?

dberenbaum:
> You had to select (Add an experiment) button to make it work. I think the idea here is that it should be automatic. We can't expect workspace to be always selected, at least not at the moment.

I was doing it to focus on the live updating rather than the auto selection. Take a look at the blog post video for an example where workspace wasn't manually selected (using pre-0.5.24):

https://user-images.githubusercontent.com/2308172/206856997-f989cfce-31e2-47a3-9052-e86953c64714.mov

> do you have exp save in that case? actually in both cases? what do you see in the table?

They are both made using `DVCLiveLogger(save_dvc_exp=True)` (from the same repo linked above). Again, check the blog post video above to see how the table looks (I don't notice anything odd or any changes between versions).

shcheklein:
> I was doing it to focus on the live updating rather than the auto selection. Take a look at the blog post video for an example where workspace wasn't manually selected (using pre-0.5.24):

But even on that video, workspace is pre-selected, no?

What we want and don't want from the VS Code perspective (and probably dvc exp show as well):

- whenever I run an experiment (via DVCLive or run button or `dvc exp run`) I would love to see in the table that something is running (spinner for the workspace). We can auto-select workspace I think.
- it means that we can see plots being update for this run.
- when it's done I want to see this experiment in the table (if it's saved)
- ideally, we should select it and unselect workspace and preserve the same color

mattseddon:
There are a few comments here. I've tried to address the important parts. Please LMK if I have missed anything important.

> Q) Could VScode skip this new optimization for auto-selecting experiments?
If there are no deps/outs in exp show output, It means we are in DVCLive only and we Could Skip.

Trying to avoid making any assumptions based on the data. The main reason for this is that assumption(s) can be broken with a simple change to the dependency. Relying on the proposed assumption and skipping would also mean introducing a separate logic branch for "DVCLive only".

> What does the "running" state mean?
> Without this P.R. I am already seeing the workspace row being updated and If I select workspace, I see the plot updates. What additional "running" signal is needed?

The workspace row is being updated because the underlying data is being changed. The data change fires file watchers that we have in place. However, this doesn't mean much to the extension under the current implementation. The data updates could be due to applying an experiment, running checkout after changing branches, or anything like that. If we want the resultant experiment to be selected (which I think we do) then we have to signal to the extension that has been created by running something in the workspace.

> I think this also still leaves us without a solution for auto-selecting plots with non-dvclive dvc exp save, which isn't as important but still would be nice not to break.

I think we can live without this.

> @mattseddon Is this better than reverting the recent vs code changes until we come up with something better? That's still not clear to me, but I will trust whatever you and @daavoo agree here because a quick fix is most important right now.

I think this is the correct change to make for now. We are not doing anything irreversible but we do need to make an incremental change to move forwards.

> I guess @mattseddon can answer how it works (file watchers?). So I don't follow why this signal file is needed?
> The problem is that if I run a 2nd experiment, the 1st one disappears because it was never auto-selected:

Yes, the workspace record is being updated by file watchers. The gap here is selecting the workspace for plotting if it is unselected and an experiment starts in the workspace. To cover the second point we need to move the workspace selection to the experiment that has been created so it is not lost if there is another run/a new experiment created. As stated previously I would like to avoid relying on further assumptions put on top of the data that we receive from `exp show`. 

IMO the old implementation/UX of auto-selecting all experiments for plotting was really bad. The reasons for the change were as follows:

- Generally, duplicates were selected (workspace + experiment)
- The implementation hurt performance (both fetching the revisions and then holding all the data in memory).
- Made plots illegible when changing between branches (all experiments selected).
- Different default behaviour when compared to plots diff (default is to select workspace + most recent commit - not workspace + recent commit + all experiment in the workspace).
- Not having anything selected on first use helps with discoverability (If I click this, the experiment gets a color/what does that do).
- Had to make a decision on which experiments to select when there were > 7.

---

> Implementation - the biggest concern is that this file won't disappear if the process dies and users will have to deal with it? I think we have in DVC file-based locks that are being unlocked if process dies. I would take a look there.

We can save the PID in the file and check for that PID when accessing the file (in both DVCLive and VS Code). This is what DVC does.

Hope that answers everything please LMK if we need to clarify anything.

mattseddon:
> We can save the PID in the file and check for that PID when accessing the file (in both DVCLive and VS Code). This is what DVC does.

I'll make this update on Monday (tomorrow) which will hopefully mean I have unblocked the release.

daavoo:
> > We can save the PID in the file and check for that PID when accessing the file (in both DVCLive and VS Code). This is what DVC does.
> 
> ~I'll make this update on Monday (tomorrow)~ I've made this change hopefully I have unblocked the release 🙏🏻. Please re-review and LMK if you need any changes @daavoo / @dberenbaum

As a follow-up, we could use the signal file to make VSCode aware of the name of the experiment created by DVCLive while it's still being run, in case VSCode wants to do something with that in the UI

PR #394, opened 2022-12-09
log_sklearn_plot: fix .json replacement

dberenbaum:
Currently breaks when using `.json` in the `name` because `live._plots` keeps the suffix.


daavoo (approved):
Added a test case and reordered tests a bit

PR #399, opened 2022-12-10
Use `get_random_exp_name` from dvc.

### LINKED ISSUE #396: random_exp_name: update to use/match dvc
`random_exp_name` was broken by https://github.com/iterative/dvc/pull/8659 because it removes `_validate_new_ref`.

daavoo:
Closes #396

Requires DVC release


src/dvclive/live.py
```diff
@@ -110,10 +110,12 @@ def _init_dvc(self):
                     # `dvc repro` execution: skip
                     self._save_dvc_exp = False
                 else:
+                    from dvc.repo.experiments.utils import get_random_exp_name
```
dberenbaum, line 113: I guess it's safe to import here because `self._dvc_repo is not None`?

src/dvclive/live.py
```diff
@@ -122,7 +122,7 @@ def _init_dvc(self):
                     self._save_dvc_exp = False
                 else:
                     # `DVCLive Only` execution
-                    self._exp_name = random_exp_name(
+                    self._exp_name = get_random_exp_name(
                         self._dvc_repo, self._baseline_rev
```
dberenbaum, line 126:
```suggestion
                        self._dvc_repo.scm, self._baseline_rev
```

Or else change the args for `get_random_exp_name`?

daavoo: 🙏 Got messed up when resolving conflicts with main

dberenbaum (approved):
What do you think of adding back some basic tests, like that it returns a string and that it will return a different name each time?

daavoo:
> What do you think of adding back some basic tests, like that it returns a string and that it will return a different name each time?

It is tested in https://github.com/iterative/dvc/blob/main/tests/func/experiments/test_utils.py#L4 , since this is a exact copy-paste, felt like no reason to replicate the same

PR #401, opened 2022-12-12
studio: Fix passing `rev` in `dvc repro` execution.

daavoo:
`dvc repro` execution was not correctly getting `baseline_rev`, breaking Live metrics remote scenario


mvshmakov (approved):
Nice, thanks for a hot-fix!

PR #404, opened 2022-12-12
Set cache: 'pip'

daavoo:


skshetry:
Github actions provide 10 GB of cache storage. It seems that a single workflow requires a 9.44 GB  cache (and we might soon add support for 3.11), so that would not be enough.

PR #407, opened 2022-12-13
studio: Use `ls-remote --get-url` as default.

daavoo:
In CI scenarios the previous logic was not reliable. Feels reasonable to just use this simpler logic as default for all scenarios.


mvshmakov (approved):
Thanks, let's see how it'll work.

daavoo:
> Thanks, let's see how it'll work.

It worked for me in example repo and Studio prod

PR #413, opened 2022-12-19
Use `dvc-studio-client`.

### LINKED ISSUE #397: remove dvc lib requirement
> - it should be done w/o DVC api. I don't like that we keep depending on it. It's really heavy, it's a very strong assumption that it's installed this way ...

_Originally posted by @shcheklein in https://github.com/iterative/dvclive/issues/381#issuecomment-1344611831_
      
Dvclive now depends on the dvc api for a few things like:
* saving experiments -- can we call the cli?
* checking for dvc repro execution -- is there another way to detect this?
* anything else?

daavoo:
Need to merge https://github.com/iterative/dvc/pull/8688 so DVC includes `dvc-studio-client` (pending to resolve #397 )


src/dvclive/live.py
```diff
@@ -338,11 +346,24 @@ def make_report(self):
...
     def end(self):
         self.make_summary(update_step=False)
-        if self._studio_url and self._studio_token:
-            if "done" not in self._studio_events_to_skip:
-                if not post_to_studio(self, "done"):
-                    logger.warning("`post_to_studio` `done` event failed.")
-                self._studio_events_to_skip.add("done")
+        if "done" not in self._studio_events_to_skip:
+            response = False
+            if post_live_metrics is not None:
```
nimdraugsael, line 351: if dvc_studio_client is not installed then it will fail silently. Maybe add a logger.warning("post_to_studio` `done` event failed - dvc_studio_client package is not installed"?

daavoo: I added it to the `start` event (you could not really reach the `done` event without the package installed).

Issue #317, opened 2022-10-01
`log_param`: check integrations and update them log params


shcheklein:
E.g. in the Lighting logger there is call:

```
    @rank_zero_only
    def log_hyperparams(self, params, *args, **kwargs):
        pass
```

should we utilize it to automatically track things?

daavoo:
> should we utilize it to automatically track things?

Yes, we should use `log_params` there

daavoo:
Examples of logging parameters in other frameworks:

- [fast.ai](fast.ai)
- [huggingface](https://github.com/huggingface/transformers/blob/bc21aaca789f1a366c05e8b5e111632944886393/src/transformers/integrations.py#L1189)

dberenbaum:
I would suggest we prioritize at the same time as #305 since it's part of the same story (auto logging everything from the framework callbacks).

PR #416, opened 2022-12-19
fastai: Log some params.

daavoo:
https://github.com/iterative/dvclive/issues/317

### DIFF
--- a/src/dvclive/fastai.py
+++ b/src/dvclive/fastai.py
@@ -30,7 +30,7 @@ class DVCLiveCallback(Callback):
         model_file: Optional[str] = None,
         with_opt: bool = False,
         live: Optional[Live] = None,
-        **kwargs
+        **kwargs,
     ):
         super().__init__()
         self.model_file = model_file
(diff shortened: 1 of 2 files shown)


src/dvclive/fastai.py
```diff
@@ -12,14 +12,29 @@ def __init__(
...
+    def before_fit(self):
+        if hasattr(self, "lr_finder") or hasattr(self, "gather_preds"):
+            return
+        params = {
+            "model": type(self.learn.model).__qualname__,
+            "batch_size": getattr(self.dls, "bs", None),
+            "batch_per_epoch": len(getattr(self.dls, "train", [])),
+            "frozen": bool(getattr(self.opt, "frozen_idx", -1)),
+            "frozen_idx": getattr(self.opt, "frozen_idx", -1),
+            "transforms": f"{getattr(self.dls, 'tfms', None)}",
```
daavoo, line 31: @alex000kim Asked for your review since I know you use fastai. 

Do you think these params are relevant for being "autologged"? Some missing?

alex000kim: I'd add `n_epoch` and `base_lr`.
In general, you can log optimizer's hyperparams as done in wandb and comet integrations:
- https://github.com/fastai/fastai/blob/master/fastai/callback/wandb.py#L117
- https://github.com/fastai/fastai/blob/master/fastai/callback/comet.py#L61

PR #432, opened 2023-01-26
report: Add `notebook` mode.

### LINKED ISSUE #309: `report`: render inline in ipython
Render the report inline within a notebook instead of in a separate HTML/mardown document. Should be prioritized after we have a basic DVCLive-only workflow for parameters/metrics/plots.

daavoo:
Renders existing HTML report inside an IFrame and updates it on each next_step.

Closes #309

---

https://user-images.githubusercontent.com/12677733/217227396-5ea5f552-bd9c-460a-8ebb-66892a969413.mp4

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -56,6 +56,7 @@ tests =
     %(plots)s
     %(dvc)s
     %(markdown)s
+    ipython
 dev =
     %(tests)s
     %(all)s
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -37,6 +37,8 @@ class DVCLiveLogger(Logger):
             self._live_init["dir"] = dir
         self._experiment = experiment
         self._version = run_name
+        # Force Live instantiation
+        self.experiment  # noqa pylint: disable=pointless-statement
 
     @property
     def name(self):
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -18,10 +18,16 @@ from .dvc import (
 )
 from .error import InvalidDataTypeError, InvalidParameterTypeError, InvalidPlotTypeError
 from .plots import PLOT_TYPES, SKLEARN_PLOTS, Image, Metric, NumpyEncoder
-from .report import make_report
+from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
 from .studio import get_studio_updates
-from .utils import env2bool, matplotlib_installed, nested_update, open_file_in_browser
+from .utils import (
+    env2bool,
+    inside_notebook,
+    matplotlib_installed,
+    nested_update,
+    open_file_in_browser,
+)
 
 try:
     from dvc_studio_client.env import STUDIO_TOKEN
@@ -62,6 +68,7 @@ class Live:
         os.makedirs(self.dir, exist_ok=True)
 
         self._report_mode: Optional[str] = report
+        self._report_notebook = None
         self._init_report()
 
         if self._resume:
@@ -176,8 +183,20 @@ class Live:
                 self._report_mode = "md"
             else:
                 self._report_mode = "html"
-        elif self._report_mode not in {None, "html", "md"}:
-            raise ValueError("`report` can only be `None`, `auto`, `html` or `md`")
+        elif self._report_mode == "notebook":
+            if inside_notebook():
+                from IPython.display import HTML, display
+
+                self._report_mode = "notebook"
+                self._report_notebook = display(
+                    HTML(BLANK_NOTEBOOK_REPORT), display_id=True
+                )
+            else:
+                self._report_mode = "html"
+        elif self._report_mode not in {None, "html", "notebook", "md"}:
+            raise ValueError(
+                "`report` can only be `None`, `auto`, `html`, `notebook` or `md`"
+            )
         logger.debug(f"{self._report_mode=}")
 
     @property
@@ -202,8 +221,12 @@ class Live:
 
     @property
     def report_file(self) -> Optional[str]:
-        if self._report_mode in ("html", "md"):
-            return os.path.join(self.dir, f"report.{self._report_mode}")
+        if self._report_mode in ("html", "md", "notebook"):
+            if self._report_mode == "notebook":
+                suffix = "html"
+            else:
+                suffix = self._report_mode
+            return os.path.join(self.dir, f"report.{suffix}")
         return None
 
     @property
--- a/src/dvclive/report.py
+++ b/src/dvclive/report.py
@@ -20,6 +20,13 @@ if TYPE_CHECKING:
 # noqa pylint: disable=protected-access
 
 
+BLANK_NOTEBOOK_REPORT = """
+<div style="width: 100%;height: 700px;text-align: center">
+DVCLive Report
+</div>
+"""
+
+
 def get_scalar_renderers(metrics_path):
     renderers = []
     for suffix in Metric.suffixes:
@@ -123,6 +130,12 @@ def make_report(live: "Live"):
 
     if live._report_mode == "html":
         render_html(renderers, live.report_file, refresh_seconds=5)
+    elif live._report_mode == "notebook":
+        from IPython.display import IFrame
+
+        render_html(renderers, live.report_file)
+        if live._report_notebook is not None:
+            live._report_notebook.update(IFrame(live.report_file, "100%", 700))
     elif live._report_mode == "md":
         render_markdown(renderers, live.report_file)
     else:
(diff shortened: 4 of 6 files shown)


daavoo (commented):
I need to add tests and manually try the different frameworks but I think it is ready to play around @dberenbaum 

It kind of highlights the need for https://github.com/iterative/dvc-render/issues/21
Also unsure about the values to hardcode the size of the IFrame

shcheklein:
@daavoo cool! could you please do a brief screen recording to see how it looks? (no voice of course, kinda live screenshot :) )

dberenbaum:
@daavoo Does it work for you locally in [https://github.com/iterative/dvc-get-started-cv with](https://github.com/iterative/dvc-get-started-cv/tree/1-notebook-dvclive)? At first I get what looks like maybe an empty report:

Screenshot 2023-01-27 at 1 37 22 PM

Later I get a 404:

Screenshot 2023-01-27 at 1 39 00 PM

I was having trouble rendering when I played around with it last month, so it's possible it's an issue on my end.

daavoo:
> @daavoo cool! could you please do a brief screen recording to see how it looks? (no voice of course, kinda live screenshot :) )

Added a recording to P.R. description.

daavoo:
> Does it work for you locally in [https://github.com/iterative/dvc-get-started-cv with](https://github.com/iterative/dvc-get-started-cv/tree/1-notebook-dvclive)?

It does work for me.

> I was having trouble rendering when I played around with it last month, so it's possible it's an issue on my end.

Could you check for errors in the browser JS console?

Where are you calling `jupyter notebook` from?

daavoo:
> @daavoo Does it work for you locally in [https://github.com/iterative/dvc-get-started-cv with](https://github.com/iterative/dvc-get-started-cv/tree/1-notebook-dvclive)? At first I get what looks like maybe an empty report:

That is just a placeholder `IPython.display` I added on `Live.init` to ensure that the report is always on top and doesn't mess up with the other logs/plugins.

 I can remove `DVCLive Report` in favor of an empty string.
 
> Later I get a 404:

The issue is the `os.chdir` performed at the beginning of the notebook.
I will update the notebook in https://github.com/iterative/example-repos-dev/pull/161

shcheklein:
> Added a recording to P.R. description.

🙏 

Q: do we want this to be enabled by default? Feels a bit too intrusive (and it'll overlap with VS Code, Studio monitoring, etc). I think it would be great to log options when you run an experiment, .e.g:

```
If you want to see report you can do one this:

- embed into Notebooks ...
- VS Code ...
- Studio ....
```

wdyt?

dberenbaum:
Thanks @daavoo! Seems to work for me on other repos. Here's a more realistic demo (using https://github.com/iterative/lstm_seq2seq with tweaked params to run faster) to see how it interacts with other notebook outputs:

https://user-images.githubusercontent.com/2308172/215843310-5b12a7f9-0667-4c7a-98ef-5c449757b9d4.mov

At the end, I saw some repeated "blinking." Could `make_report` be called many times at the end for some reason?

https://user-images.githubusercontent.com/2308172/215843620-30a110c3-c1c7-4028-a4cf-5b5236a57b0e.mov

daavoo:
> to see how it interacts with other notebook outputs:

um. looks like the instantiation of the callback (and thus `Live` object) is being delayed there. 
That is unfortunate, as the placeholder IPython.display doesn't do the trick.
Could you try manually instantiating `Live` object outside and passing it to the callback?

daavoo:
> At the end, I saw some repeated "blinking." Could make_report be called many times at the end for some reason?

Will take a look using the repo

dberenbaum:
> Q: do we want this to be enabled by default? Feels a bit too intrusive (and it'll overlap with VS Code, Studio monitoring, etc). I think it would be great to log options when you run an experiment, .e.g:

🤔 @shcheklein Can you explain more what you would expect all the options to say? Could you turn the message off? Is it specific to this notebook mode, or should it apply to any mode? What do you see as the default onboarding workflow for both VS Code and non-VS Code users to see live metrics?

daavoo:
> Q: do we want this to be enabled by default? Feels a bit too intrusive (and it'll overlap with VS Code, Studio monitoring, etc). 

This is enabled in `auto` mode. For those overlaps, we could chose to not use the `notebook` display if there are signals that we are overlapping.

For Studio monitoring should be doable because DVCLive has the info available. For VSCode I am not sure if there is an easy way for DVCLive to detect that the DVC extension is enabled

dberenbaum:
@daavoo Can we clean up the table? If we want to keep any kind of HTML output, it seems like this would go a long way towards making it feel more polished.

dberenbaum:
> um. looks like the instantiation of the callback (and thus `Live` object) is being delayed there. That is unfortunate, as the placeholder IPython.display doesn't do the trick until `Live` gets instantiated.

It's delayed because it is set up to only log every 50 steps.

> To verify, could you try manually instantiating `Live` object outside and passing it to the callback?

Sure. It does generate the placeholder immediately. Not sure if that looks any better?

https://user-images.githubusercontent.com/2308172/215854435-de93dd18-e3d3-42e8-8520-efb18353ab35.mov

daavoo:
> It's delayed because it is set up to only log every 50 steps.

I mean that in some frameworks the `Live` object is instantiated when the callback is instantiated but lightning uses that `experiment` property that appears to be delayed until needed.
I think we could change this in our logger, though .

> Sure. It does generate the placeholder immediately. Not sure if that looks any better?

Yeah I know it doesn't matter right now 😅 

However, the point is that the placeholder could be actually a not so ugly HTML of equal size to the future IFrame so the space is clearly reserved in the output, and it doesn't interfere with progressbars or whatever output the framework might generate

shcheklein:
> 🤔 @shcheklein Can you explain more what you would expect all the options to say? 

List of things you can do to see actual live updates:

- If you are in nobebook ... adding an option to `live` object (e.g. "render=True")...
- use Studio (a link to docs)
- install VS Code extension (link to the marketplace)

> Could you turn the message off?

Yes.

> Is it specific to this notebook mode, or should it apply to any mode?

Any mode

> What do you see as the default onboarding workflow for both VS Code and non-VS Code users to see live metrics?

Same as today + a message that would help navigating.

------

To me rendering an HTML like this inside notebook by default feels a bit too intrusive. It overlaps potentially with VS Code in some cases (ppl can get confused). Even if we go that path we should at least also show how to disable this in the HTML itself.

daavoo:
> To me rendering an HTML like this inside notebook by default feels a bit too intrusive.

Is your concern about how it is implemented (`IFrame`), about the space that it takes, or both?

> It overlaps potentially with VS Code in some cases (ppl can get confused).

If I understand correctly, the scenario overlapping would be someone with the VSCode editor and the DVC extension installed, running the notebook inside a Git+DVC Repo (`dvc init` has been run), right?

I might be wrong but I see that scenario as far less likely than any others like Google Colab, Kaggle Notebooks, Jupyterlab servers, or just people running the notebook without meeting the above conditions (running local notebooks outside a Git Repo is also common)

I think for the overlapping scenario, the best would be to figure out how to get a signal that the VSCode and DVC extension are up and running (dvc extension could expose env var maybe?).

shcheklein:
> Is your concern about how it is implemented (IFrame), about the space that it takes, or both?

Neither :).

It just doesn't feel right (by default). Fels a bit too much of a side effect from running the code by default. Are there any other examples that come to your mind that would by default do similar things?

E.g. W&B could have potentially embedded an IFrame into the notebook by default as well.

My expectation when I run my code is to see the regular output that is usually expected and produced by the framework itself. Our additional output should be interfering with it much I think.

dberenbaum:
> E.g. W&B could have potentially embedded an IFrame into the notebook by default as well.

For context, here's what W&B does (note that I can't get the example repo to work): https://docs.wandb.ai/guides/track/jupyter

Anyway, let's discuss auto behavior as a follow-up and not block the rest of the PR? We can merge without making it part of auto since that will be the less intrusive option while we discuss it.

shcheklein:
So, for W&B you need to:

```
... start a new cell with %%wandb to see live graphs in the notebook. 
```

> Anyway, let's discuss auto behavior as a follow-up and not block the rest of the PR? We can merge without making it part of auto since that will be the less intrusive option while we discuss it.

sounds good to me!

daavoo:
Updated the P.R.

@shcheklein Behavior is only enabled if `report=notebook`. 

@dberenbaum Updated the init placeholder to take the same space as the subsequent IFrame so the space gets "reserved" on init and it doesn't mess up with existing logs because of the resizing that was happening before.

Let's discuss improvements to appearance as a follow-up in https://github.com/iterative/dvc-render/issues/21 unless we consider it a blocker to merge this.

dberenbaum (approved):

Issue #367, opened 2022-11-25
Revisit `live.end` for complex scenarios


daavoo:
None of this is a blocker since I don't think anything breaks the basic live metrics workflow, but it would be good to have it to "unbreak" more complex scenarios.

> If I decouple the `post_to_studio(self, "done", logger)` from the `live._ended`, I think the snippet would behave as expected.

It's a little ugly but more likely to work the way users want (log summary/report always).

> I would expect the live metrics functionality to be only working inside the callback.

Can we also have something like this to enable the context manager to override other calls to `end`?

> ```diff
>     self.make_summary()
>     if self.report_mode == "studio":
>         if not self._ended:
>             if not post_to_studio(self, "done", logger):
>                 logger.warning("`post_to_studio` `done` event failed.")
>-           self._ended = True
>+           if not self._keep_alive: # set to True in __enter__ and False in __exit__
>+               self._ended = True
>     else:
>         self.make_report()
> ```

_Originally posted by @dberenbaum in https://github.com/iterative/dvclive/issues/364#issuecomment-1322487643_

PR #436, opened 2023-01-30
live: Skip `live.end` calls inside context manager.

daavoo:
Add `live._inside_with` boolean to handle it.

The idea is to include additional data in an experiment when using frameworks:

```python
from dvclive import Live
from dvclive.keras import DVCLiveCallback

with Live(save_dvc_exp=True) as live:
    live.log_param("foo", 2)
    model.fit(
        x, y,
        # Don't call live.end
        callbacks=[DVCLiveCallback(live=live)])
    # So it is possible to include additional data in the experiment
    live.summary["out-of-loop-metric"] = 1
```

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -55,6 +55,7 @@ class Live:
         self._images: Dict[str, Any] = {}
         self._params: Dict[str, Any] = {}
         self._plots: Dict[str, Any] = {}
+        self._inside_with = False
 
         os.makedirs(self.dir, exist_ok=True)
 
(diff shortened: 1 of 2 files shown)


daavoo (commented):
@dberenbaum I think we already discussed this somewhere but can't remember. Just found it convenient to have for https://github.com/iterative/example-repos-dev/pull/161

dberenbaum (approved):
Looks like it's in #367? I guess we need a docs update to say that `end` gets ignored inside the context manager?

PR #439, opened 2023-01-31
add fabric dep to lightning

dberenbaum:
Follow up to https://github.com/iterative/dvclive/pull/430


daavoo (commented):
@dberenbaum is this needed? Do they (pytorch_lightning) indicate it somewhere?
The imports work for me after only installing `pip install pytorch_lightning>=1.9`

dberenbaum:
Sorry, maybe it was an env issue for me.

Issue #435, opened 2023-01-28
Missing data in dvclive to be displayed by "dvc exp show"


nv-pipo:
Following the documentation in:
- https://dvc.org/doc/user-guide/experiment-management/checkpoints

The output of "dvc exp show" and "dvc metrics diff" that I can produce differs from the official documentation: The columns of metrics and steps are missing.

Output with latest dvc:
<img width="561" alt="output produced" src="https://user-images.githubusercontent.com/7769945/215275602-0abc89f6-f6af-49aa-8b55-d0a39144ab50.png">

Expected:
<img width="690" alt="expected" src="https://user-images.githubusercontent.com/7769945/215275609-14fcebdd-b328-4ac5-9a88-9f70c3fa1cc1.png">

Am I missing any flag in "dvc exp show"/"dvc metrics diff"? Should I report this to DVC repo instead? it seems to me that it is more a dvclive issue, hence the report here.

dberenbaum:
Hi @nv-pipo! Could you give some more details. Do you have a `dvc.yaml` pipeline? If so, what does it look like?

nv-pipo:
Hi @dberenbaum,
Thanks for following up!

When saving checkpoints via ```dvc exp run``` the experiment parameters are not tracked. Note that using the "Python API" (```python train.py``` and setting ```save_dvc_exp=True```) the parameters are correctly tracked.

To replicate:
~~~python
# python.py
from random import random

from dvclive import Live

NUM_EPOCHS = 2

with Live(save_dvc_exp=True) as live:
    live.log_param("epochs", NUM_EPOCHS)

    for epoch in range(NUM_EPOCHS):

        live.log_metric("metric_name1", random())
        live.next_step()
~~~
~~~yaml
# dvc.yaml
stages:
  train:
    cmd: python train.py
# live:
#   dvclive:
#     summary: true
#     html: true
~~~
Then run:
~~~shell
git init
git add .
git commit -am 'Initial'
dvc init
dvc exp run
dvc exp show
~~~
The experiment (checkpoint) is tracked but the parameters and metrics are not saved/show by ```dvc exp show```. Output of last 2 commands:
<img width="872" alt="Screenshot 2023-01-31 at 20 41 49" src="https://user-images.githubusercontent.com/7769945/215865742-a248b796-2c73-48a5-a9a3-c828c3433257.png">


However this works without issues when using the "Python API":
~~~shell
rm -f dvc.yaml
python train.py
dvc exp show
~~~
[...]

PR #440, opened 2023-01-31
don't remove dvc.yaml file

dberenbaum:
Until we address https://github.com/iterative/dvclive/issues/381 (needs a major version change), we can stop removing `dvc.yaml` during init. That way, for people who transition to using pipelines, they still keep their `dvc.yaml` from the no-pipeline runs. Should at least partially address https://github.com/iterative/dvclive/issues/435.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -96,7 +96,6 @@ class Live:
             self.metrics_file,
             self.report_file,
             self.params_file,
-            self.dvc_file,
         ):
             if f and os.path.exists(f):
                 os.remove(f)


dberenbaum:
Let's include in 2.0

src/dvclive/live.py
```diff
@@ -96,7 +96,6 @@ def _init_cleanup(self):
             self.metrics_file,
             self.report_file,
             self.params_file,
-            self.dvc_file,
```
daavoo, line 99: Though this was covered by tests 😅

daavoo (approved):

PR #443, opened 2023-02-02
add explicit make_dvcyaml arg

### LINKED ISSUE #381: make_dvcyaml: always save dvc.yaml

dberenbaum:
Fixes #381. This is a breaking change, so it will require a 2.0 release.

### DIFF
--- a/tests/test_studio.py
+++ b/tests/test_studio.py
@@ -54,7 +54,7 @@ def test_post_to_studio(tmp_dir, mocker, monkeypatch):
             "step": 0,
             "metrics": {live.metrics_file: {"data": {"step": 0, "foo": 1}}},
             "params": {live.params_file: {"fooparam": 1}},
-            "plots": {scalar_path: {"data": [{"step": 0, "foo": 1.0}]}},
+            "plots": {f"{live.dvc_file}::{scalar_path}": {"data": [{"step": 0, "foo": 1.0}]}},
             "client": "dvclive",
         },
         headers={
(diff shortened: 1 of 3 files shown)


dberenbaum:
Merging to 2.0 branch to be safe.

daavoo (approved):
This cause some UI problems in certain scenarios (duplication of plots). 

I and @dberenbaum discussed that it's more important to merge this first to fix scenarios like https://github.com/iterative/dvclive/issues/435 and will look into the UI duplication in follow-ups

tests/test_studio.py
```diff
@@ -77,7 +79,9 @@ def test_post_to_studio(tmp_dir, mocker, monkeypatch):
             "step": 1,
             "metrics": {live.metrics_file: {"data": {"step": 1, "foo": 2}}},
             "params": {live.params_file: {"fooparam": 1}},
-            "plots": {scalar_path: {"data": [{"step": 1, "foo": 2.0}]}},
+            "plots": {
+                f"{live.dvc_file}::{scalar_path}": {"data": [{"step": 1, "foo": 2.0}]}
```
daavoo, line 83: It looks like this ended breaking Live plot updates in Studio. I am trying to isolate and understand why it does work for some plots but not others

PR #444, opened 2023-02-02
Shorten names of fields in plots tsv files

dberenbaum:
This is another breaking change, so it would be good to merge it along with #443, although it doesn't need to block if it's not quick to get merged.

This shortens repetitive plots fields. For example, instead of `results/plots/metrics/train/epoch/acc.tsv: train/epoch/acc`, it becomes `results/plots/metrics/train/epoch/acc.tsv: acc`.


shcheklein (approved):

dberenbaum:
Merging to 2.0 branch to be safe.

PR #445, opened 2023-02-02
Drop DataAlreadyLogged error

### LINKED ISSUE #431: Overwrite manually logged images
Hey guys, thanks for your great work!

I'm logging a plot using `live.log_image(filename, fig_pil)` as part of my notebook. Now, we usually need a couple of tries to get the plot that we want. It would be great if dvclive could overwrite an already logged image file instead of throwing an `DataAlreadyLoggedError`. Otherwise, managing this becomes a huge headache when you are working with multiple plots or images.

For example, as additional argument: 
```Python
log_image(name, val, overwrite=True)
```

Thanks for considering!

dberenbaum:
Fixes #431

### DIFF
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -225,6 +225,14 @@ def test_resume_env_var(tmp_dir, monkeypatch):
     assert Live()._resume
 
 
+@pytest.mark.parametrize("metric", ["m1", os.path.join("train", "m1")])
+def test_allow_step_override(tmp_dir, metric):
+    dvclive = Live("logs")
+
+    dvclive.log_metric(metric, 1.0)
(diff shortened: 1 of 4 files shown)


daavoo (approved):

tests/test_main.py
```diff
@@ -229,19 +225,6 @@ def test_resume_env_var(tmp_dir, monkeypatch):
     assert Live()._resume
 
 
-@pytest.mark.parametrize("metric", ["m1", os.path.join("train", "m1")])
```
daavoo, line 232: minor: Would be nice to have an explicit test overriding a value

dberenbaum: Added, thanks!

PR #446, opened 2023-02-02
2.0

dberenbaum:


daavoo (commented):
@dberenbaum cleanup test is failing because of merging both #384 and #440 .

daavoo (approved):
I did not really review in depth the short metric names in `.tsv` but I think it's ok

daavoo:
@dberenbaum Would be great if you can add some short descriptions for the relevant changes in the release notes

dberenbaum:
> I did not really review in depth the short metric names in `.tsv` but I think it's ok

I tried to test it out a bit in VS Code/Studio. Hard to tell with live metrics because I think I'm running into unrelated issues, but nothing looks broken yet.

dberenbaum:
Thanks @daavoo! I will try to log back on in a few hours to do the release.

PR #447, opened 2023-02-02
Make dvcyaml explicit

dberenbaum:
See https://github.com/iterative/dvc.org/pull/4286/files#r1095007570

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -229,7 +229,10 @@ class Live:
             self._step = 0
 
         self.make_summary()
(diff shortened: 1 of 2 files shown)


shcheklein (approved):
Should we change `make report` to be the same? ;)

dberenbaum:
> Should we change `make report` to be the same? ;)

If it's okay, let's make it a separate issue (I feel like it's been discussed before but doubt we have an issue for it). It's a bit more complicated (not a boolean flag, lots more logic inside it).

PR #448, opened 2023-02-03
`save_dvc_exp` works with existing pipelines/repro

dberenbaum:
@daavoo @shcheklein I messed up the release 🤦 . I was so focused on the `dvc.yaml` changes that I forgot we also need to update `save_dvc_exp` to work when there's an existing pipeline.

There's an open question of whether it should work with `dvc repro`. In the interest of being simple and explicit, I went with saving an experiment even during `dvc repro`. In other words, the only time saving an experiment gets skipped is inside `dvc exp run`.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -119,7 +119,7 @@ class Live:
                 )
         elif self._save_dvc_exp:
             if self._dvc_repo is not None:
-                # `DVCLive Only` execution
+                # `DVCLive Only` or `dvc repro` execution
                 self._exp_name = get_random_exp_name(
                     self._dvc_repo.scm, self._baseline_rev
                 )
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -119,20 +119,11 @@ class Live:
                 )
         elif self._save_dvc_exp:
             if self._dvc_repo is not None:
-                if any(
-                    not stage.is_data_source for stage in self._dvc_repo.index.stages
-                ):
-                    logger.warning(
-                        "Ignoring `_save_dvc_exp` because there is an existing"
-                        " `dvc.yaml` file."
-                    )
-                    self._save_dvc_exp = False
-                else:
-                    # `DVCLive Only` execution
-                    self._exp_name = get_random_exp_name(
-                        self._dvc_repo.scm, self._baseline_rev
-                    )
-                    mark_dvclive_only_started()
+                # `DVCLive Only` execution
+                self._exp_name = get_random_exp_name(
+                    self._dvc_repo.scm, self._baseline_rev
+                )
+                mark_dvclive_only_started()
             else:
                 logger.warning(
                     "Can't save experiment without a DVC Repo."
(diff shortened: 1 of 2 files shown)


shcheklein:
@dberenbaum should it be merged into 2.0? (I see previous commits)

daavoo:
> @dberenbaum should it be merged into 2.0? (I see previous commits)

I think just need to rebase `main` on this branch

daavoo:
> > @daavoo @shcheklein I messed up the release 🤦 . I was so focused on the dvc.yaml changes that I forgot we also need to update save_dvc_exp to work when there's an existing pipeline.
> 
> I think no big deal, we can make a new release on top or if we care that much about semver there are many options:
> 
> https://snarky.ca/what-to-do-when-you-botch-a-release-on-pypi/

I don't have permission to yank the release on pypi, would need to be done by @shcheklein

src/dvclive/live.py
```diff
@@ -119,20 +119,11 @@ def _init_dvc(self):
...
-                        " `dvc.yaml` file."
-                    )
-                    self._save_dvc_exp = False
-                else:
-                    # `DVCLive Only` execution
-                    self._exp_name = get_random_exp_name(
-                        self._dvc_repo.scm, self._baseline_rev
-                    )
-                    mark_dvclive_only_started()
+                # `DVCLive Only` execution
```
daavoo, line 122:
```suggestion
                # `DVCLive Only` or `dvc repro` execution
```

daavoo (approved):
> There's an open question of whether it should work with dvc repro. In the interest of being simple and explicit, I went with saving an experiment even during dvc repro. In other words, the only time saving an experiment gets skipped is inside dvc exp run.

Given that the option still defaults to `False`, I think no big deal.

My only "worry" is that I would prefer to deprecate `dvc repro` as a corner case of `dvc exp run` (trying to fix all issues on current `dvc exp run` so that could be the case).

Making `save_dvc_exp` work on `dvc repro` kind of goes against that direction, but I am not strictly against it.

dberenbaum:
> @dberenbaum should it be merged into 2.0? (I see previous commits)

`2.0` was already merged, deleted, and released as `2.0.0`.

For a fix, I was thinking of doing it like this:

* Merge this into `main`.
* Release `2.0.1` with updated release notes.
* Add a `2.0` tag to `2.0.1` (there's no `2.0` tag yet, only `2.0.0`).
* Possibly update the release notes for `2.0.0` to mention not to use it?

Let me know if you think it's worth pulling the `2.0.0` release @shcheklein.

> My only "worry" is that I would prefer to deprecate `dvc repro` as a corner case of `dvc exp run` (trying to fix all issues on current `dvc exp run` so that could be the case).
> 
> Making `save_dvc_exp` work on `dvc repro` kind of goes against that direction, but I am not strictly against it.

👍 I don't think we will fully deprecate it, but I agree with it being a corner case, low-level command. My thinking here was that we shouldn't bother with special logic to handle that corner case.

With the current PR, I don't think we need to document any special behavior for `save_dvc_exp`. It will always save an experiment. The fact that DVCLive doesn't generate the experiment during `exp run` is an implementation detail -- an experiment still gets created.

PR #450, opened 2023-02-03
metrics: preserve the name stored in the `.tsv`.

daavoo:
Closes #449

PR #451, opened 2023-02-03
Add report titles

dberenbaum:
After #450, the report lost too much context by only using the short names from the tsv files. There was no way to tell which plot was `train/loss` vs `test/loss`:

Screenshot 2023-02-03 at 8 37 21 AM

It looked fine in other DVC plots because they include the path:

Screenshot 2023-02-03 at 8 36 04 AM

This PR adds titles to help separate plots with shortened names:

Screenshot 2023-02-03 at 8 50 05 AM

I wasn't sure how adding a `title` property would work with Studio live metrics, but it doesn't seem to have an impact (which is fine since they already have the file paths in the title, and more importantly it doesn't break anything):

Screenshot 2023-02-03 at 9 02 37 AM

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -119,20 +119,11 @@ class Live:
                 )
         elif self._save_dvc_exp:
             if self._dvc_repo is not None:
-                if any(
-                    not stage.is_data_source for stage in self._dvc_repo.index.stages
-                ):
-                    logger.warning(
-                        "Ignoring `_save_dvc_exp` because there is an existing"
-                        " `dvc.yaml` file."
-                    )
-                    self._save_dvc_exp = False
-                else:
-                    # `DVCLive Only` execution
-                    self._exp_name = get_random_exp_name(
-                        self._dvc_repo.scm, self._baseline_rev
-                    )
-                    mark_dvclive_only_started()
+                # `DVCLive Only` or `dvc repro` execution
+                self._exp_name = get_random_exp_name(
+                    self._dvc_repo.scm, self._baseline_rev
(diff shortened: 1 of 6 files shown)


daavoo (commented):
Nice catch.

For some reason, I though the renderer class in `dvc-render` was automatically setting the `title` if not present, but looks like the logic was left on DVC side:

https://github.com/iterative/dvc/blob/5d2701341b6318af5695e45a356f0acf605fe041/dvc/render/match.py#L103-L104

daavoo (approved):

daavoo:
>   wasn't sure how adding a title property would work with Studio live metrics,

The changes in this P.R. don't affect it as you are only modifying `report.py` and

dberenbaum:
> > wasn't sure how adding a title property would work with Studio live metrics,
> 
> The changes in this P.R. don't affect it as you are only modifying `report.py`

I was confused because I saw you made changes to `report.py` in #450. Prior to that, the report would have looked like:

Screenshot 2023-02-03 at 8 36 27 AM

Anyway, the title looks like a nice improvement to the UI and more closely matches the DVC plots output, so let's go with this.

daavoo:
> I was confused because I saw you made changes to `report.py` in #450. Prior to that, the report would have looked like:

Yeah sorry, those changes were not strictly related to fixing the studio updates and I thought that the titles were being inferred in `dvc-render`.
The changes in `report.py` were meant to ensure consistency in the data points with `dvc plots show` (https://github.com/iterative/dvclive/pull/450#discussion_r1095790377)

PR #455, opened 2023-02-05
Introduce  `dvcyaml: bool = True`  kwarg for DVCLiveLogger

francisquintallauzon:
This PR introduces a the `dvcyaml: bool = True,` kwarg for the DVCLiveLogger for Pytorch Lightning following the addition of the same kwarg to the Live logger class in [PR 443](https://github.com/iterative/dvclive/pull/443).  Tests were updated accordingly 

I did not update documentation because it is already out of date.  

This PR also adds `py` as a dependency since it is no longer an explicit dependency of pytest. 

* [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. https://github.com/iterative/dvc.org/issues/4294

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,6 +52,7 @@ tests =
     pylint==2.15.0
     pylint-plugin-utils>=0.6
     mypy==0.981
+    py==1.11.0
     %(image)s
     %(plots)s
     %(dvc)s
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,7 +52,6 @@ tests =
     pylint==2.15.0
     pylint-plugin-utils>=0.6
     mypy==0.981
-    py==1.11.0
     %(image)s
     %(plots)s
     %(dvc)s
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -52,6 +52,7 @@ tests =
     pylint==2.15.0
     pylint-plugin-utils>=0.6
     mypy==0.981
+    py==1.11.0
```
daavoo, line 55: Hi @francisquintallauzon , sorry I am not fully understanding why this change is needed. How are you running the tests?

daavoo (approved):
Thank you very much @francisquintallauzon !

I just have a question about the change to the deps, which is not really related to the rest of the P.R.

setup.cfg
daavoo, line 55:
```suggestion
```

francisquintallauzon: @daavoo, tests were not running with the current config on my MacOS without this addition to setup.cfg.  I did not try to reproduce the issue with the Github Action's environment though and it might just be happening on my computer. 

image

francisquintallauzon:
> Thank you very much @francisquintallauzon !
> 
> I just have a question about the change to the deps, which is not really related to the rest of the P.R.

Happy to help even if it's a tiny contribution!

PR #462, opened 2023-02-13
Early draft of log_artifact

daavoo:


src/dvclive/live.py
```diff
@@ -299,6 +301,19 @@ def log_param(
         """Saves the given parameter value to yaml"""
         self.log_params({name: val})
 
+    def log_artifact(
+        self,
+        path: str,
+    ):
+        """Saves the given parameter value to yaml"""
+        if self._dvc_repo is not None and not self._inside_dvc_exp:
```
daavoo, line 309: We are going to encounter lock issues here if we are inside `dvc repro` execution. 
Probably need to add a new property `self._inside_dvc_repro` and a way to infer this.

dberenbaum:
@daavoo Can you discuss with @karajan1001? He was going to take this issue.

daavoo:
> @daavoo Can you discuss with @karajan1001? He was going to take this issue.

Yes, we discussed and I open the P.R. to share what I already had

src/dvclive/live.py
```diff
@@ -299,6 +301,19 @@ def log_param(
...
+            dvc_file = self._dvc_repo.add(path)[0].addressing
```
dberenbaum, line 310: Should we check if it already exists as an output in either `dvc.yaml` or `.dvc`?

src/dvclive/live.py
```diff
@@ -299,6 +301,19 @@ def log_param(
...
+            if self._save_dvc_exp:
```
dberenbaum, line 311: Any reason not to do this regardless of `self._save_dvc_exp`?

src/dvclive/live.py
karajan1001: I think we can wrap the exception here.

daavoo:
Closing in favor of https://github.com/iterative/dvclive/pull/464

PR #467, opened 2023-02-23
prevent calling live.end if exception raised

dberenbaum:
This will prevent `live.end` from being called in the event of an exception.

This doesn't solve the problem of showing different completion statuses. For now, we only have a binary choice of whether an experiment is running or completed. This PR makes it so that we will continue to show an experiment in running status when interrupted by an exception. This isn't great, but I think it's more expected than showing it as successfully completed.

Edit: I also think it's probably good practice to abort the rest of `live.end` in the case of unhandled exceptions.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import shutil
+import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


dberenbaum:
This was very sloppy. If you think this is an improvement, we can fix it, but let's decide on whether it's a worthwhile improvement first.

PR #471, opened 2023-02-27
studio: Don't skip data event if `_inside_dvc_exp`.

daavoo:
Workaround #397 for live experiments with `dvc exp run`.

Closes https://github.com/iterative/dvclive/issues/473

Allows to send `data` updates when running inside `dvc exp run` and DVC is not available as a Python library.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -146,7 +146,7 @@ class Live:
                 self._studio_events_to_skip.add("done")
                 return
 
-        if not self._dvc_repo:
+        if not (self._dvc_repo or self._inside_dvc_exp):
             logger.debug("`studio` report can't be used without a DVC Repo.")
             self._studio_events_to_skip.add("start")
             self._studio_events_to_skip.add("data")
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -5,10 +5,13 @@ import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
+from dvc_studio_client.env import STUDIO_TOKEN
+from dvc_studio_client.post_live_metrics import post_live_metrics
 from ruamel.yaml.representer import RepresenterError
 
 from . import env
 from .dvc import (
+    dvc_api_available,
     get_dvc_repo,
     get_random_exp_name,
     make_checkpoint,
@@ -28,13 +31,6 @@ from .utils import (
     nested_update,
     open_file_in_browser,
 )
-
-try:
-    from dvc_studio_client.env import STUDIO_TOKEN
-    from dvc_studio_client.post_live_metrics import post_live_metrics
-except ImportError:
-    post_live_metrics = None
-    STUDIO_TOKEN = None
 
 logging.basicConfig()
 logger = logging.getLogger("dvclive")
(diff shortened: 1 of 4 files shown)


dberenbaum:
Does it work for `repro` (which is still much more common in CML)?

> DVC is not available as a Python library

Can we better clarify in the code or tests that this is what it's for? It took me a bit to understand how `self._dvc_repo` was being used here, and it seems poorly named for this usage. Maybe we could have distinct methods/attributes to determine whether there is a DVC repo and whether DVC is installed as a lib?

daavoo:
> Yes, we should do that. Addressing it should also improve the warnings as the current warning is misleading when DVC is not available as a lib.

Will just do it in this P.R.

daavoo:
> Does it work for repro (which is still much more common in CML)?

To add on this, I just feel like we should not support live experiments with `dvc repro`

dberenbaum:
> To add on this, I just feel like we should not support live experiments with `dvc repro`

Let's discuss in #474 😄

dberenbaum:
@daavoo Is it ready for another review?

src/dvclive/live.py
```diff
@@ -5,10 +5,13 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
+from dvc_studio_client.env import STUDIO_TOKEN
+from dvc_studio_client.post_live_metrics import post_live_metrics
 from ruamel.yaml.representer import RepresenterError
 
 from . import env
 from .dvc import (
+    dvc_installed,
```
shcheklein, line 14: minor: name is a bit misleading ... `dvc_lib_available`?

daavoo:
> @daavoo Is it ready for another review?

@dberenbaum yes

src/dvclive/live.py
```diff
@@ -29,13 +32,6 @@
     open_file_in_browser,
 )
 
-try:
-    from dvc_studio_client.env import STUDIO_TOKEN
-    from dvc_studio_client.post_live_metrics import post_live_metrics
-except ImportError:
-    post_live_metrics = None
-    STUDIO_TOKEN = None
-
```
daavoo, line 38: Unrelated but `dvc_studio_client` has been in `install_requires` of dvclive for a while

dberenbaum:
A couple questions:

1. This looks like it prioritizes warning about installing DVC lib over having a DVC repo. Maybe it's better to always check both and provide both warnings so people don't solve one error only to run into the next?
2. It doesn't need to block this PR, but it feels like with the increasing integration with DVC, we need to come to some resolution on #397 and consider in what scenarios DVCLive makes sense without DVC or a DVC repo. It may be better to require these generally rather than check for them for enabling each individual feature.

daavoo:
>     1. This looks like it prioritizes warning about installing DVC lib over having a DVC repo. Maybe it's better to always check both and provide both warnings so people don't solve one error only to run into the next?

Done.
 
>     2. It doesn't need to block this PR, but it feels like with the increasing integration with DVC, we need to come to some resolution on [remove dvc lib requirement #397](https://github.com/iterative/dvclive/issues/397) and consider in what scenarios DVCLive makes sense without DVC or a DVC repo. It may be better to require these generally rather than check for them for enabling each individual feature.

Let's discuss in the issue

daavoo:
Closing in favor of #481

PR #478, opened 2023-03-01
Add step to plots config

dberenbaum:
Plots in `dvclive/plots/metrics` do not use the `step` value stored in the tsv but instead infer a step value that's a simple counter of the row numbers. This PR configures those plots to use the `step` value from the tsv as the `x` axis.

Example using the lightning callback, which logs every 50 train steps by default:

Before this PR:

visualization(37)

After this PR:

visualization(36)


daavoo (approved):
Nice one

PR #481, opened 2023-03-02
Require DVC >2.45.1

daavoo:
To unbreak to live-experiments, see https://github.com/iterative/studio/issues/5259 for context

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,13 +28,9 @@ package_dir=
     =src
 packages = find:
 install_requires=
-    dvc_render[table]>=0.1.2
-    ruamel.yaml>=0.17.11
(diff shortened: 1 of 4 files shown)


dberenbaum:
Do we first need to drop dvclive from the dvc deps to avoid a circular dependency?

dberenbaum (approved):

daavoo (commented):
Pending of a new DVC release

dberenbaum:
Should we also update the min DVC version required here? I guess it should require 2.45.2?

daavoo:
> Should we also update the min DVC version required here? I guess it should require 2.45.2?

Just updated the requirement to `dvc>2.45.1`

PR #482, opened 2023-03-02
lightning: call next step per epoch

### LINKED ISSUE #477: lightning: Find a better way to handle `next_step`.
>  It was the DeviceStatsMonitor callback from Lightning! It tried to log every single validation step and really slowed the whole thing down.

We still need to figure out a better way of deciding when to call `next_step` in the lightning integration.

In other frameworks, the callbacks for integrations have a clear interface with events like `on_epoch_end`, but lightning Logger is a little messy because you only have access to `log_metrics`, and this can be called by other callbacks and at different paces.

dberenbaum:
Fixes #477. Calls `next_step` only for epoch-level metrics, not step-level metrics. It may get called twice each epoch (once for train and once for val), but it's much less frequent than current behavior.

### DIFF
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -81,15 +81,16 @@ class DVCLiveLogger(Logger):
         ), "experiment tried to log from global_rank != 0"
 
         self.experiment.step = step
-        epoch = False
+        step = False
         for metric_name, metric_val in metrics.items():
             if is_tensor(metric_val):
                 metric_val = metric_val.cpu().detach().item()
-            if metric_name.endswith("_epoch"):
-                epoch = True
+            if metric_name.endswith("_step"):
+                step = True
             metric_name = standardize_metric_name(metric_name, __name__)
             self.experiment.log_metric(name=metric_name, val=metric_val)
-        if epoch:
+        # Only call next_step for epoch-level metrics
+        if not step:
             self.experiment.next_step()
 
     @rank_zero_only
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -81,12 +81,16 @@ class DVCLiveLogger(Logger):
         ), "experiment tried to log from global_rank != 0"
 
         self.experiment.step = step
+        epoch = False
         for metric_name, metric_val in metrics.items():
             if is_tensor(metric_val):
                 metric_val = metric_val.cpu().detach().item()
+            if metric_name.endswith("_epoch"):
+                epoch = True
             metric_name = standardize_metric_name(metric_name, __name__)
             self.experiment.log_metric(name=metric_name, val=metric_val)
-        self.experiment.next_step()
+        if epoch:
+            self.experiment.next_step()
 
     @rank_zero_only
     def finalize(self, status: str) -> None:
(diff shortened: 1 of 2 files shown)


src/dvclive/lightning.py
```diff
@@ -81,12 +81,16 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
         ), "experiment tried to log from global_rank != 0"
 
         self.experiment.step = step
+        epoch = False
         for metric_name, metric_val in metrics.items():
             if is_tensor(metric_val):
                 metric_val = metric_val.cpu().detach().item()
+            if metric_name.endswith("_epoch"):
+                epoch = True
```
daavoo, line 89: Take this with a grain of salt (because the more I look into [logger_connector](https://github.com/Lightning-AI/lightning/blob/master/src/lightning/pytorch/trainer/connectors/logger_connector/logger_connector.py), the more confused I am 😅 ), but I think this should be:

```suggestion
            if metric_name == "epoch":
```

It appears that the inclusion of `_step`/`_epoch` suffix depends on how people set up their `log` calls, so we might fail to set `epoch=True`.

However, lightning logs the `epoch` value as a metric, and it seems to align with the frequency specified with `log_every_n_steps`.

src/dvclive/lightning.py
```diff
@@ -81,12 +81,16 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
...
             metric_name = standardize_metric_name(metric_name, __name__)
             self.experiment.log_metric(name=metric_name, val=metric_val)
-        self.experiment.next_step()
+        if epoch:
+            self.experiment.next_step()
```
daavoo, line 93: Because we already set `step` above:

```suggestion
            self.experiment.make_summary()

            if self.experiment._dvcyaml:
                self.experiment.make_dvcyaml()

            self.experiment.make_report()
```

src/dvclive/lightning.py
daavoo: Ok I think what we want is to only calle next_step if `self.experiment.step != step` (this need to be checked at the top)

dberenbaum: I think instead what we want is to exclude anything that ends with `_step`. WDYT?

daavoo: I mentioned the `step` logic because another fun thing is that other things can call `log_metrics` separately but with the same step value example, the [LearningRateMonitor](https://pytorch-lightning.readthedocs.io/en/stable/api/pytorch_lightning.callbacks.LearningRateMonitor.html)

dberenbaum: > `_step`/`_epoch` prefixes are only included if both `on_step=True, on_epoch=True` has been passed to `self.log`.

Yeah, but I think that's fine because it's only a problem in that scenario. If I have `on_step=False`, the logging is already reasonable. If I have `on_epoch=False`, I think we have no choice but to log everything (I doubt it's very common). If we only skip on `metric_name.endswith("_step")`, it's pretty safe while still reducing the noise for `on_step=True, on_epoch=True` scenarios.

daavoo (commented):
Problem is that this doesn't take into account external callers of `log_metrics` and doesn't respect the user's `log_every_n_steps`.

After consulting with the pillow, I think #483 is what we want

PR #483, opened 2023-03-03
lightning: Determine when to call next_step based on `inspect.stack`.

### LINKED ISSUE #477: lightning: Find a better way to handle `next_step`.

daavoo:
Only call next_step for

- `update_train_step_metrics`
- `update_train_epoch_metrics`
- `log_eval_end_metrics`

Prevents calling next_step when external callbacks call `log_metrics` or during the multiple `update_eval_step_metrics`.

Closes #477

### DIFF
--- a/tests/test_frameworks/test_lightning.py
+++ b/tests/test_frameworks/test_lightning.py
@@ -1,6 +1,7 @@
 import os
 
 import torch
+from dvc_studio_client.env import STUDIO_ENDPOINT, STUDIO_REPO_URL, STUDIO_TOKEN
 from pytorch_lightning import LightningModule
 from pytorch_lightning.trainer import Trainer
 from torch import nn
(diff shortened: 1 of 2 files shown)


tests/test_frameworks/test_lightning.py
```diff
@@ -126,7 +126,7 @@ def test_lightning_default_dir(tmp_dir):
         logger=dvclive_logger,
         max_epochs=2,
         enable_checkpointing=False,
-        log_every_n_steps=1,
+        log_every_n_steps=2,
```
dberenbaum, line 129: Any reason for this change?

dberenbaum (approved):

tests/test_frameworks/test_lightning.py
daavoo: Nop

PR #490, opened 2023-03-07
lightning: prevent conflict with internal `step` property.

daavoo:
If a metric didn't contain `split` but only `freq` suffix, `standardize_metric_name` was causing a conflict in the summary. For example: `grad_step` was being stored as `step/grade`.

### DIFF
--- a/src/dvclive/utils.py
+++ b/src/dvclive/utils.py
@@ -91,8 +91,10 @@ def standardize_metric_name(metric_name: str, framework: str) -> str:
         split, freq, rest = None, None, None
         if parts[0] in ["train", "val", "test"]:
             split = parts.pop(0)
-        if parts[-1] in ["step", "epoch"]:
-            freq = parts.pop()
+            # Only set freq if split was also found.
+            # Otherwise we end up conflicting with out internal `step` property.
+            if parts[-1] in ["step", "epoch"]:
+                freq = parts.pop()
         rest = "_".join(parts)
         parts = [part for part in (split, freq, rest) if part]
         metric_name = "/".join(parts)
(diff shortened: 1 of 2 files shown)


daavoo:
This highlights the underlying issue of treating our internal property `step` as a metric, but that is a whole different story

dberenbaum (approved):

dberenbaum:
Actually, let me take another look here 🙏

dberenbaum:
Okay, you can merge, but I think we need to follow up on this.

The root of the problem is here, right?

https://github.com/iterative/dvclive/blob/767649532dfb79e5f69684df2392ce9fa5c3596f/src/dvclive/utils.py#L34-L35

We try to override `{"step": 0}` with `{"step": {"grad": ...}}` but there is no key inside `step`. It seems like we should be checking that `d.get(k, {})` returns a mapping. If not, we can either override or fail more gracefully.

This could happen without treating `step` as a metric. For example, even without `step`, you could get the same error for `epoch` here if you include epoch-level metrics.

dberenbaum:
Also: should we invert `{split}/{freq}/{metric}` to be `{metric}/{freq}/{split}`? This would keep all variations on a specific metric grouped together, making it easier to compare them. This would have to be a 3.0 change, but it would also reduce these conflicts.

daavoo:
> It seems like we should be checking that d.get(k, {}) returns a mapping. If not, we can either override or fail more gracefully.

Correct. We just never encountered the situation before of trying to convert an already logged metric to a dictionary

> (edit: why aren't we just merging dictionaries here? what added value does `nested_update` provide?)

How would you merge a nested value? See description of https://github.com/iterative/dvclive/pull/197

daavoo:
> This would keep all variations on a specific metric grouped together, making it easier to compare them.

I think it's arguable that there are use cases where you want the current behavior and others where you prefer what you describe. No real good answer.

dberenbaum:
> I think it's arguable that there are use cases where the current behavior would be preferable and others where not. No real good answer.

A specific metric will always have values on the same scale, and in my experience it's more natural to have these all next to each other rather than have potentially dissimilar metrics grouped together for an entire split. Would you generally prefer to have them grouped as they are now?

PR #497, opened 2023-03-13
log stage template

### LINKED ISSUE #468: generate pipeline
#464/#378 is only useful outside of pipelines, but adding a `.dvc` file makes the transition to pipelines harder. Instead, we could add a `dvc.yaml` pipeline with the model as an output. If inside a DVC repo, DVCLive can generate a `dvc.yaml` pipeline stage in the root of the project like:

```
stages:
  dvclive:
    cmd: python train.py
    deps:
      - train.py
    outs:
      - dvclive/metrics.json:
          cache: false
      - dvclive/plots:
          cache: false
      - model.pt
```

On `next_step`/`end`, DVCLive can do `dvc commit dvclive` to save the changes to `dvc.lock`. We may not always be able to infer the `cmd` or `deps`, but we could use a dummy `cmd` and drop `deps` in that case. It still provides a useful skeleton for transitioning to pipelines instead of making it harder to transition.

dberenbaum:
Closes #468.

Logs output during `Live.end()` like:

```
INFO:dvclive:To run with DVC, add this to /Users/dave/Code/lstm_seq2seq/dvc.yaml:
stages:
  dvclive:
    cmd: <python my_code_file.py my_args>
    deps:
    - <my_code_file.py>
    outs:
    - dvclive/params.yaml:
        cache: false
    - dvclive/metrics.json:
        cache: false
    - dvclive/plots:
        cache: false
    - model
```

### DIFF
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -2,11 +2,12 @@
 import logging
 import os
 import random
+from io import StringIO
 from pathlib import Path
 
 from dvclive import env
 from dvclive.plots import Image, Metric
-from dvclive.serialize import dump_yaml
+from dvclive.serialize import dump_yaml, get_yaml
 
 logging.basicConfig()
 logger = logging.getLogger("dvclive")
@@ -156,3 +157,28 @@ def get_random_exp_name(scm, baseline_rev):
         exp_ref = ExpRefInfo(baseline_sha=baseline_rev, name=name)
         if not scm.get_ref(str(exp_ref)):
             return name
+
+
+def get_dvc_stage_template(live):
+    stage = {
+        "cmd": "<python my_code_file.py my_args>",
+        "deps": ["<my_code_file.py>"],
+        "outs": [],
+    }
+    if live._params:
+        params_path = os.path.relpath(live.params_file, live.dir)
+        stage["outs"].append({params_path: {"cache": False}})
+    if live._metrics:
+        metrics_path = os.path.relpath(live.metrics_file, live.dir)
+        stage["outs"].append({metrics_path: {"cache": False}})
+    if live._metrics or live._images or live._plots:
+        plots_path = Path(live.plots_dir).as_posix()
+        stage["outs"].append({plots_path: {"cache": False}})
+    stage["outs"] += list(live._outs)
+    dvcyaml_dict = {"stages": {"dvclive": stage}}
+
+    output = StringIO()
+    get_yaml().dump(dvcyaml_dict, output)
+    dvcyaml = output.getvalue()
+    output.close()
+    return dvcyaml
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -166,11 +166,9 @@ def get_dvc_stage_template(live):
         "outs": [],
     }
     if live._params:
-        params_path = os.path.relpath(live.params_file, live.dir)
-        stage["outs"].append({params_path: {"cache": False}})
+        stage["outs"].append({live.params_file: {"cache": False}})
     if live._metrics:
-        metrics_path = os.path.relpath(live.metrics_file, live.dir)
-        stage["outs"].append({metrics_path: {"cache": False}})
+        stage["outs"].append({live.metrics_file: {"cache": False}})
     if live._metrics or live._images or live._plots:
         plots_path = Path(live.plots_dir).as_posix()
         stage["outs"].append({plots_path: {"cache": False}})
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -165,12 +165,15 @@ def get_dvc_stage_template(live):
         "deps": ["<my_code_file.py>"],
         "outs": [],
     }
+    rel_path = os.path.relpath(os.getcwd(), live._dvc_repo.root_dir)
     if live._params:
-        stage["outs"].append({live.params_file: {"cache": False}})
+        params_path = (Path(rel_path) / Path(live.params_file)).as_posix()
+        stage["outs"].append({params_path: {"cache": False}})
     if live._metrics:
-        stage["outs"].append({live.metrics_file: {"cache": False}})
+        metrics_path = (Path(rel_path) / Path(live.metrics_file)).as_posix()
+        stage["outs"].append({metrics_path: {"cache": False}})
     if live._metrics or live._images or live._plots:
-        plots_path = Path(live.plots_dir).as_posix()
+        plots_path = (Path(rel_path) / Path(live.plots_dir)).as_posix()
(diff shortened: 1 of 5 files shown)


dberenbaum:
TODO: tests

src/dvclive/live.py
```diff
@@ -403,6 +404,16 @@ def end(self):
             finally:
                 mark_dvclive_only_ended()
 
+        if self._dvc_repo and not self._inside_dvc_exp:
+            from dvc.exceptions import DvcException
+
+            try:
+                path = os.path.join(self._dvc_repo.root_dir, "dvc.yaml")
```
daavoo, line 411: The paths in the yaml should also be relative to `self._dvc_repo.root_dir`

src/dvclive/dvc.py
```diff
@@ -156,3 +157,29 @@ def get_random_exp_name(scm, baseline_rev):
         exp_ref = ExpRefInfo(baseline_sha=baseline_rev, name=name)
         if not scm.get_ref(str(exp_ref)):
             return name
+
+
+def get_dvc_stage_template(live):
+    stage = {
+        "cmd": "<python my_code_file.py my_args>",
```
daavoo, line 164: WDYT of just showing `python train.py` without `<>` syntax or anything.

tests/test_dvc.py
```diff
@@ -185,3 +188,106 @@ def test_exp_save_dvcexception_is_ignored(tmp_dir, mocker):
...
+                "cmd": "<python my_code_file.py my_args>",
+                "deps": ["<my_code_file.py>"],
+                "outs": [],
+            }
+        }
+    }
+
+
+def test_get_dvc_stage_template_params(tmp_dir, mocked_dvc_repo):
+    mocked_dvc_repo.root_dir = tmp_dir
```
daavoo, line 209: Is this needed?

src/dvclive/dvc.py
```diff
@@ -156,3 +157,29 @@ def get_random_exp_name(scm, baseline_rev):
...
+        "deps": ["<my_code_file.py>"],
+        "outs": [],
+    }
+    rel_path = os.path.relpath(os.getcwd(), live._dvc_repo.root_dir)
```
daavoo, line 168: Can we add a test for this? I.e. using `chdir` inside a subdirectory in the test

src/dvclive/dvc.py
dberenbaum: 🤔 I worry it will confuse users (although anything we put here might); ideally I'd like to make clear that it's up to the user to fill in the `cmd`. Why do you prefer `python train.py` and do you have any other ideas for how to clarify that it needs to be modified?

tests/test_dvc.py
dberenbaum: Without I get an assertion error like:

```
E         Full diff:
E           {
E            'stages': {'dvclive': {'cmd': '<python my_code_file.py my_args>',
E                                   'deps': ['<my_code_file.py>'],
E         -                         'outs': [{'dvclive/params.yaml': {'cache': False}}]}},
E         +                         'outs': [{'../../../dvclive/params.yaml': {'cache': False}}]}},
E         ?                                    +++++++++
E           }
```

daavoo: Ok, that looks bad 😅 I need to fix the fixtures

src/dvclive/dvc.py
daavoo: > Why do you prefer python train.py and do you have any other ideas for how to clarify that it needs to be modified?

I don't strongly prefer it and don't really know what would be less confusing 😅 
I was just thinking that it would be easier for me to rename the file than to update the entire cmd and remove the `<>`

Leaving it up to you.

daavoo (approved):

tests/test_dvc.py
dberenbaum: I can put something like `_dvc_repo.root_dir = "."` into https://github.com/iterative/dvclive/blob/main/tests/conftest.py#L20 if you prefer.

src/dvclive/dvc.py
dberenbaum: Added a test. PTAL.

tests/test_dvc.py
```diff
@@ -185,3 +190,135 @@ def test_exp_save_dvcexception_is_ignored(tmp_dir, mocker):
...
+            }
+        }
+    }
+
+
+def test_get_dvc_stage_template_chdir(tmp_dir, mocked_dvc_repo):
+    mocked_dvc_repo.root_dir = tmp_dir
+    d = tmp_dir / "sub" / "dir"
+    d.mkdir(parents=True)
+    os.chdir(d)
```
daavoo, line 302:
```suggestion
def test_get_dvc_stage_template_chdir(tmp_dir, mocked_dvc_repo, monkeypatch):
    mocked_dvc_repo.root_dir = tmp_dir
    d = tmp_dir / "sub" / "dir"
    d.mkdir(parents=True)
    monkeypatch.chdir(d)
```

src/dvclive/live.py
```diff
@@ -403,6 +404,16 @@ def end(self):
...
+                yaml = get_dvc_stage_template(self)
+                logger.info(f"To run with DVC, add this to {path}:\n{yaml}")
```
daavoo, line 413: Maybe we should also mention the need for `dvc remove {out}`, as we are listing the `live._outs`

tests/test_dvc.py
daavoo: The monkeypatch one will be undone automatically when the test finish

tests/test_dvc.py
daavoo: I think that maybe `mocked_dvc_repo` should not be a fixture but a function you call passing the `tmp_dir` fixture you have on each test

dberenbaum: Adding to the fixture like I mention above does work. Was your concern about that or more about something else like style or correctness?

src/dvclive/live.py
daavoo: Although DVC would already hint that whith the new error message. NVM

src/dvclive/dvc.py
dberenbaum: Do you think the brackets:
a. Make it clear this needs to be modified
b. Will be left in by users so that they end up running `<python train.py>`

src/dvclive/live.py
dberenbaum: 🤔 It's an interesting idea. I worry about adding more comments like this though, so maybe let's wait and see how many people complain? I tried to improve the errors in https://github.com/iterative/dvc/pull/9138 so that it's at least more obvious how to fix the error when you encounter it.

Issue #499, opened 2023-03-14
frameworks: Remove `model_file`


daavoo:
I think that it would be better to remove the `model_file` argument in favor of using the native callback for saving models of each framework.

DVCLive doesn't need to take care of serializing the model or repricating logic already existing in the framework.

Implementing https://github.com/iterative/dvclive/issues/465#issuecomment-1440655000 would make the DVCLive callback track the model file with DVC which is all we should be doing.

PR #500, opened 2023-03-14
frameworks: Use `log_artifact`.

daavoo:
Tested with setup in https://github.com/iterative/example-get-started-experiments/

```py
from dvclive import Live
from fastai.callback.tracker import SaveModelCallback
train_arch = 'resnet18'

for base_lr in [0.1, 0.01, 0.001]:
    with Live(dir=os.path.join('results', 'train'), save_dvc_exp=True) as live:
        learn = unet_learner(
            data_loader, arch=getattr(models, train_arch), metrics=DiceMulti, 
            path="results")
        fine_tune_args = {
            'epochs': 2,
            'base_lr': base_lr
            }
        live.log_params(fine_tune_args)
        learn.fine_tune(
            **fine_tune_args,
            cbs=[DVCLiveCallback(live=live), SaveModelCallback()])
```

And then retrieving the best model with:

```
!dvc exp show --csv > results.csv
```

```py
import dvc.api
import pandas as pd

results = pd.read_csv("results.csv")
best_exp = results.Experiment[results.dice_multi.argmax()]

with dvc.api.open(
    "results/models/model.pth", rev=best_exp, mode="rb"
) as f:
    learn = learn.load(f)
```

### DIFF
--- a/src/dvclive/fastai.py
+++ b/src/dvclive/fastai.py
@@ -67,7 +67,8 @@ class DVCLiveCallback(Callback):
         # fast.ai calls after_epoch but we don't want to increase the step.
         if logged_metrics:
             if self.model_file:
-                self.learn.save(self.model_file, with_opt=self.with_opt)
+                file = self.learn.save(self.model_file, with_opt=self.with_opt)
+                self.live.log_artifact(str(file))
             self.live.next_step()
 
     def after_fit(self):
@@ -76,4 +77,7 @@ class DVCLiveCallback(Callback):
         if _inside_fine_tune() and not self.freeze_stage_ended:
             self.freeze_stage_ended = True
         else:
+            if hasattr(self, "save_model"):
+                if self.save_model.last_saved_path:
+                    self.live.log_artifact(str(self.save_model.last_saved_path))
             self.live.end()
--- a/src/dvclive/huggingface.py
+++ b/src/dvclive/huggingface.py
@@ -6,6 +6,7 @@ from transformers import (
     TrainerState,
     TrainingArguments,
 )
+from transformers.trainer import Trainer
 
 from dvclive import Live
 from dvclive.utils import standardize_metric_name
@@ -42,6 +43,7 @@ class DVCLiveCallback(TrainerCallback):
             tokenizer = kwargs.get("tokenizer")
             if tokenizer:
                 tokenizer.save_pretrained(self.model_file)
+            self.live.log_artifact(self.model_file)
 
     def on_train_end(
         self,
@@ -50,4 +52,10 @@ class DVCLiveCallback(TrainerCallback):
         control: TrainerControl,
         **kwargs
     ):
+        if args.load_best_model_at_end:
+            trainer = Trainer(
+                args=args, model=kwargs.get("model"), tokenizer=kwargs.get("tokenizer")
+            )
+            trainer.save_model()
+            self.live.log_artifact(args.output_dir)
         self.live.end()
--- a/src/dvclive/keras.py
+++ b/src/dvclive/keras.py
@@ -51,6 +51,7 @@ class DVCLiveCallback(Callback):
                 self.model.save_weights(self.model_file)
             else:
                 self.model.save(self.model_file)
+            self.live.log_artifact(self.model_file)
         self.live.next_step()
 
     def on_train_end(
--- a/tests/test_frameworks/test_fastai.py
+++ b/tests/test_frameworks/test_fastai.py
@@ -1,6 +1,7 @@
 import os
 
 import pytest
+from fastai.callback.tracker import SaveModelCallback
 from fastai.tabular.all import (
     Categorify,
     Normalize,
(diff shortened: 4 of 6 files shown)


src/dvclive/fastai.py
```diff
@@ -67,7 +67,8 @@ def after_epoch(self):
         # fast.ai calls after_epoch but we don't want to increase the step.
         if logged_metrics:
             if self.model_file:
-                self.learn.save(self.model_file, with_opt=self.with_opt)
+                file = self.learn.save(self.model_file, with_opt=self.with_opt)
+                self.live.log_artifact(str(file))
```
daavoo, line 71: @dberenbaum Opened this one to discuss before implementing all frameworks.

I think we should go ahead and deprecate `model_file` (https://github.com/iterative/dvclive/issues/499) but for now I just mad both approaches work (`model_file` and external callback).

src/dvclive/fastai.py
```diff
@@ -76,4 +77,7 @@ def after_fit(self):
         if _inside_fine_tune() and not self.freeze_stage_ended:
             self.freeze_stage_ended = True
         else:
+            if hasattr(self, "save_model"):
+                if self.save_model.last_saved_path:
+                    self.live.log_artifact(str(self.save_model.last_saved_path))
```
dberenbaum, line 82: How does fastai organize the saved models? Will the user end up with a new .dvc file for each call?

daavoo: Checking into the code it looks like the only problem would be if the user sets `every_epoch=True` (defaults to `False`). We could skip `log_artifact` if `every_epoch=True`

dberenbaum: Looking at the docs in https://docs.fast.ai/callback.tracker.html#savemodelcallback, should we be logging `learn.path/learn.model_dir` instead of a specific file?

daavoo: > Looking at the docs in https://docs.fast.ai/callback.tracker.html#savemodelcallback, should we be logging learn.path/learn.model_dir instead of a specific file? 

In cases where there is a single model, the `model_dir` would only contain `last_saved_path`. 

In the other case (`every_epoch=True`) I think logging the entire dir makes the downstream usage worse.

dberenbaum: 🤔 It makes #140 more complex since we would have to figure out which file to load, right? Either way, hopefully it shouldn't matter to the user if we can make the integration work well. I'm surprised how little there is out there for resuming in situations like this.

> We could skip log_artifact if every_epoch=True

Do you mean calling `log_artifact` only at the end of the experiment? Or not calling it at all?

daavoo: > Do you mean calling log_artifact only at the end of the experiment? Or not calling it at all?

Not calling it at all, document it, and let the user decide what to pass to `log_artifact`.

I was playing around adapting https://github.com/iterative/example-get-started-experiments/ and I think I would actually want to call [`learn.export`](https://docs.fast.ai/learner.html#learner.export) after training and use `log_artifact` on its output.

You can use [`load_learner`](https://docs.fast.ai/learner.html#load_learner) on downstream usage but the files saved by `SaveModelCallback` require you to manually instantiate the learner and call `learn.load`, which is inconvenient.

daavoo: > I thought this was the preferred way for loading models in pytorch?

My understanding is that it is the best way to save the model in case one of the potential usages is resuming the training. 

---

Anyhow, the point is that I believe calling `log_artifact` with `every_epoch=True` is not really useful. If someone chooses to use the option, they can still use `log_artifact` manually.

I think by default we should only call `log_artifact` in the scenarios where we are sure is a single model file.

dberenbaum: After discussing with @daavoo, I think we should move forward with the current implementation that's in this PR for simplicity.

src/dvclive/lightning.py
```diff
@@ -113,6 +115,15 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
                 self.experiment._latest_studio_step -= 1
             self.experiment.next_step()
 
+    def _log_checkpoint(self, checkpoint_callback):
+        self.experiment.log_artifact(checkpoint_callback.dirpath)
```
daavoo, line 119: I couldn't really find a way to log the single best model. Lightning by default uses filename patterns so we would be logging different files unless the user hardcodes a fixed pattern.

At this point, I am not sure if we would be better just documenting the context manager pattern and let the user decide what to log outside the callback.

dberenbaum:
@daavoo Let's hold off on releasing until I have a chance to review 🙏

daavoo:
> @daavoo Let's hold off on releasing until I have a chance to review 🙏

😅

dberenbaum:
Okay, in that case, let's not document them yet so we can revert if needed? I'll try to take a look today.

src/dvclive/huggingface.py
```diff
@@ -50,4 +52,10 @@ def on_train_end(
         control: TrainerControl,
         **kwargs
     ):
+        if args.load_best_model_at_end:
+            trainer = Trainer(
+                args=args, model=kwargs.get("model"), tokenizer=kwargs.get("tokenizer")
+            )
+            trainer.save_model()
+            self.live.log_artifact(args.output_dir)
```
dberenbaum, line 60: This was the only other "auto-logging" that you added, right?

Issue #456, opened 2023-02-06
dvc.yaml ignored if entire dvclive folder is tracked as a stage output


dberenbaum:
Reproduction script below. `exp show` shows nothing in the workspace once `dvclive` is added as a stage output.

```bash
$ git clone git@github.com:iterative/dvclive-exp-tracking.git

$ cd dvclive-exp-tracking

$ python train.py

$ dvc exp show
 ────────────────────────────────────────────────────────────────────────────────────
  Experiment                 Created        train_loss   epoch   step   encoder_size
 ────────────────────────────────────────────────────────────────────────────────────
  workspace                  -                0.026054       4    499   64
  main                       Feb 01, 2023            -       -      -   -
  └── 4f01b48 [leafy-wage]   01:10 PM         0.026054       4    499   64
 ────────────────────────────────────────────────────────────────────────────────────

$ dvc stage add -n train -d train.py -O dvclive python train.py

$ dvc exp show
 ──────────────────────────────────────────────────────────────────────────────────────>
  Experiment                 Created        train_loss   epoch   step   encoder_size   >
 ──────────────────────────────────────────────────────────────────────────────────────>
  workspace                  -                       -       -      -   -              >
  main                       Feb 01, 2023            -       -      -   -              >
  └── 4f01b48 [leafy-wage]   01:10 PM         0.026054       4    499   64             >
 ──────────────────────────────────────────────────────────────────────────────────────>
```

This happens because `dvc.yaml` is within the output, which DVC doesn't seem to like. You can avoid it by writing the stage like:

```bash
$ dvc stage add -f -n train -d train.py -O dvclive/metrics.json -o dvclive/plots python train.py
```

However, it's probably not safe to rely on users doing this. Some ways we can remedy it:
1. Write to root `dvc.yaml` (see discussion in #381).
2. Fix on DVC side to look for `dvc.yaml` inside an output (not sure this makes sense since it could include pipeline stages, which would be odd to have inside an output).
3. Other ideas?

daavoo:
> Other ideas?

Don't mention `dvclive` folder at all in the root `dvc.yaml`? What would be the downside?

dberenbaum:
That's true. It's not likely to be a dependency of another stage. Regardless, not sure if documentation is enough here or if it's worth taking some action to prevent possible confusion for users who try it.

shcheklein:
[...]

PR #506, opened 2023-03-24
remove dvclive outputs from dvc stage outs

dberenbaum:
Drop dvclive outs from printed stage template.

Related:
* https://github.com/iterative/example-repos-dev/pull/191#discussion_r1147486180
* https://github.com/iterative/dvclive/issues/456
* https://github.com/iterative/dvc.org/pull/4389#discussion_r1140805922

### DIFF
--- a/tests/test_dvc.py
+++ b/tests/test_dvc.py
@@ -199,74 +199,6 @@ def test_get_dvc_stage_template_empty(tmp_dir, mocked_dvc_repo):
             "dvclive": {
                 "cmd": "<python my_code_file.py my_args>",
                 "deps": ["<my_code_file.py>"],
-                "outs": [],
-            }
-        }
-    }
-
-
-def test_get_dvc_stage_template_params(tmp_dir, mocked_dvc_repo):
-    live = Live()
-    live.log_param("foo", 1)
-    template = get_dvc_stage_template(live)
-
-    assert YAML_LOADER.load(template) == {
-        "stages": {
-            "dvclive": {
-                "cmd": "<python my_code_file.py my_args>",
-                "deps": ["<my_code_file.py>"],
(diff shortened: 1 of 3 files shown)


tests/test_dvc.py
```diff
@@ -293,10 +225,6 @@ def test_get_dvc_stage_template_chdir(tmp_dir, mocked_dvc_repo, monkeypatch):
     d.mkdir(parents=True)
     monkeypatch.chdir(d)
     live = Live("live")
-    live.log_param("foo", 1)
-    live.log_metric("bar", 1)
-    live.log_image("img.png", Image.new("RGB", (10, 10), (250, 250, 250)))
-    live.log_sklearn_plot("confusion_matrix", [0, 0, 1, 1], [0, 1, 1, 0])
```
daavoo, line 299: Let's keep these lines so that we test that we are "ignoring" them.

daavoo (approved):

tests/test_dvc.py
```diff
@@ -293,10 +225,6 @@ def test_get_dvc_stage_template_chdir(tmp_dir, mocked_dvc_repo, monkeypatch):
     d.mkdir(parents=True)
     monkeypatch.chdir(d)
     live = Live("live")
```
dberenbaum, line 227:
```suggestion
    live = Live("live")
    live.log_param("foo", 1)
    live.log_metric("bar", 1)
    live.log_image("img.png", Image.new("RGB", (10, 10), (250, 250, 250)))
    live.log_sklearn_plot("confusion_matrix", [0, 0, 1, 1], [0, 1, 1, 0])
```

shcheklein:
Q: If we do this, does it clean up all the DVClive outputs between experimnets when I run it via dvc exp run?

Issue #474, opened 2023-02-27
studio: live metrics for `dvc repro`


dberenbaum:
Opening a new issue so we can consolidate discussion in one place. 

Live metrics theoretically work regardless of how you run your experiment. Someone could run `dvc repro` with DVCLive and `STUDIO_TOKEN` and expect to get live metrics in Studio. In fact, this is maybe the most common way I would expect it to be used 😅 because:

1. Studio live metrics seem most useful in CI.
2. We know that most CI/CML jobs use `dvc repro`.

We could encourage people to use `dvc exp run` instead, but I have some concerns with that approach:

1. Makes it harder for users to onboard since they have to migrate to a different workflow in their CI, which may not seem worth it since it could introduce unexpected problems.
2. `dvc exp run` has known limitations (for example, requiring non-default fetch depth) when running in CI. It doesn't seem to have any real advantages when running in CI.

What are the downsides to allowing Studio live metrics inside `dvc repro`?

daavoo:
> What are the downsides to allowing Studio live metrics inside dvc repro?

No experiment ref.

dberenbaum:
> No experiment ref.

Do I want an exp ref in CML? Is it just going to add yet another row in Studio with the same info? Or should we work towards having the end result of a CML workflow be an exp ref instead of a regular commit? What would be the pros/cons?

daavoo:
> Is it just going to add yet another row in Studio with the same info?

You are already adding one row for the live experiment, even though it is not an exp ref.

> Or should we work towards having the end result of a CML workflow be an exp ref instead of a regular commit? 

IMO, Yes. 

The workflow should serve as a replacement for the current `cml pr/cml report` towards making it more Studio centric. 
CML should just be used for provisioning the runner.

People that don't want to use Studio can still use the old CML workflow, but they obviously don't need live experiments in Studio.

> What would be the pros/cons?

From the top of my mind:

### Cons

- People would need to update the workflow to use the live experiments (although they already need to update it for STUDIO_TOKEN).
-`cml report` supports arbitrary Markdown report.
- Studio doesn't have an individual view per experiment, would need to implement a way to share a useful link for going from GitHub to the experiment in Studio.

### Pros

- Simpler workflow file for onboarding.
- Studio-centric 
- Cleaner UI in Studio (row mess when using `cml pr`).
- Cleaner git history (branch and p.rs by `cml pr` even if you end up discarding the experiment).
- Unification of the different concepts of `experiments`.

[...]

PR #507, opened 2023-03-27
live: Explicitly require `exp` for sending Studio updates.

daavoo:
Requires using either `dvc exp run` or `save_dvc_exp=True`.

Closes https://github.com/iterative/dvclive/issues/474

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -70,7 +70,7 @@ class Live:
             self._init_cleanup()
 
         self._baseline_rev: Optional[str] = None
-        self._exp_name: str = "dvclive-exp"
+        self._exp_name: Optional[str] = None
         self._experiment_rev: Optional[str] = None
         self._inside_dvc_exp: bool = False
         self._dvc_repo = None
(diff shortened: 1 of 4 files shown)


daavoo (commented):
@dberenbaum The discussion in https://github.com/iterative/dvclive/issues/474 was stalling but since we are moving forward in Sudio towards unifying Live Experiments and Experiment refs (i.e. see https://github.com/iterative/studio/pull/5417), I think it's better to stop allowing sending updates without an experiment.

daavoo:
> @daavoo I'm not against this, but I still feel like I'm maybe missing the motivation. 

My main motivation is that I think of the use cases that this P.R forbids as lose ends.

> Does it help simplify something?

It allows Studio to use baseline SHA and experiment name pair as unique identifier for all live experiments, which makes thing simpler on their end and avoid changes to th client. 

It also allows both running and completed experiments to be referenced by URL using the same logic

dberenbaum (approved):
> It allows Studio to use baseline SHA and experiment name pair as unique identifier for all live experiments, which makes thing simpler on their end and sabes the need of changes to the client.
> 
> It also allows both running and completed experiments to be referenced by URL using the same logic

Okay, this seems like enough reason since we don't have any strong argument in the other direction. Thanks for explaining.

PR #518, opened 2023-04-09
Support path to an image in `log_image`

shcheklein:
Log image from an existing file. Helps to convert a repo like YOLO where images are produced already by some core code and we need to capture and transfer them into the DVCLive output dir.

It also adds some test for `log_image`.

Related: https://github.com/shcheklein/ultralytics
Related: https://github.com/ultralytics/ultralytics/issues/1909

Docs https://github.com/iterative/dvc.org/pull/4463

-------

* [X] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -269,6 +269,11 @@ class Live:
         if not Image.could_log(val):
             raise InvalidDataTypeError(name, type(val))
 
+        if isinstance(val, str) or isinstance(val, Path):
+            from PIL import Image as ImagePIL
+
+            val = ImagePIL.open(val)
+
         if name in self._images:
             data = self._images[name]
         else:
--- a/src/dvclive/plots/image.py
+++ b/src/dvclive/plots/image.py
@@ -1,4 +1,4 @@
-from pathlib import Path
+from pathlib import Path, PurePath
 
 from .base import Data
 
@@ -18,6 +18,8 @@ class Image(Data):
         if val.__class__.__module__ == "PIL.Image":
             return True
         if val.__class__.__module__ == "numpy":
+            return True
+        if isinstance(val, PurePath) or isinstance(val, str):
             return True
         return False
 
--- a/tests/plots/test_image.py
+++ b/tests/plots/test_image.py
@@ -24,11 +24,38 @@ def test_invalid_extension(tmp_dir):
 
 @pytest.mark.parametrize("shape", [(10, 10), (10, 10, 3), (10, 10, 4)])
 def test_numpy(tmp_dir, shape):
+    from PIL import Image as ImagePIL
+
     live = Live()
     img = np.ones(shape, np.uint8) * 255
     live.log_image("image.png", img)
 
-    assert (tmp_dir / live.plots_dir / LiveImage.subfolder / "image.png").exists()
+    img_path = tmp_dir / live.plots_dir / LiveImage.subfolder / "image.png"
+    assert img_path.exists()
+
+    val = np.asarray(ImagePIL.open(img_path))
+    assert np.array_equal(val, img)
+
+
+def test_path(tmp_dir):
+    import numpy as np
+    from PIL import Image as ImagePIL
+
+    live = Live()
+    image_data = np.random.randint(0, 255, (100, 100, 3)).astype(np.uint8)
+    pil_image = ImagePIL.fromarray(image_data)
+    image_path = tmp_dir / "temp.png"
+    pil_image.save(image_path)
+
+    live = Live()
+    live.log_image("foo.png", image_path)
+    live.end()
+
+    plot_file = tmp_dir / live.plots_dir / "images" / "foo.png"
+    assert plot_file.exists()
+
+    val = np.asarray(ImagePIL.open(plot_file))
+    assert np.array_equal(val, image_data)
 
 
 def test_override_on_step(tmp_dir):


src/dvclive/live.py
```diff
@@ -269,6 +269,11 @@ def log_image(self, name: str, val):
         if not Image.could_log(val):
             raise InvalidDataTypeError(name, type(val))
 
+        if isinstance(val, str) or isinstance(val, Path):
+            from PIL import Image as ImagePIL
+
+            val = ImagePIL.open(val)
+
```
dberenbaum, line 276: So we end up making our own copy of these images inside `dvclive/plots/images`, right? Should we leave them as is and add `plots: [original/path/to/image/file]` in `dvclive/dvc.yaml`?

Not that strong an opinion here but want to hear your thoughts.

src/dvclive/plots/image.py
```diff
@@ -19,6 +19,8 @@ def could_log(val: object) -> bool:
             return True
         if val.__class__.__module__ == "numpy":
             return True
+        if isinstance(val, PurePath) or isinstance(val, str):
+            return True
```
dberenbaum, line 23: Should we check if the path exists and has a supported extension? Or is this intentionally left to later when we try to read it as an image?

Do we want to support a directory of images here? Not a blocker, just another idea to mention.

src/dvclive/live.py
shcheklein: > Should we leave them as is and add plots: [original/path/to/image/file] in dvclive/dvc.yaml?

I think in this case we can expect that user take care of them on their own (e.g. in the `example-get-started` repo). Also, it might be another option for this, yes. I think can have both eventually. `dvc.yaml` might become very complicated though ...

In this specific case, the YOLO repo creates images in some temporary directory specific to an experiment and we need to capture it from that directory. Ideally, the repo itself should be using `log_image` instead of creating them the regular way, but I don't think it's realistic / practical to expect them to convert it soon. So for now, to unblock it and show the benefits I think it's a reasonable approach.

src/dvclive/plots/image.py
shcheklein: I think it's fine to fail when it tries to read it.

> has a supported extension

Good question. Let me think and see if we want to check it here. Not sure. E.g. we don't check in the `if val.__class__.__module__ == "PIL.Image"` that is has the right format?

> Do we want to support a directory of images here?

Yes, can be a follow up later if this is needed.

tests/test_main.py
```diff
@@ -440,6 +440,43 @@ def checkpoint_never_exists(path):
...
+    import numpy as np
+    from PIL import Image as ImagePIL
+
+    live = Live()
+    image_data = np.random.randint(0, 255, (100, 100, 3)).astype(np.uint8)
+    live.log_image("foo.png", image_data)
+    live.end()
+
+    plot_file = os.path.join(tmp_dir, live.plots_dir, "images", "foo.png")
+    assert os.path.exists(plot_file)
```
daavoo, line 453: This part is already tested in:

https://github.com/iterative/dvclive/blob/1cb31aa2f906396c1a32f1e0bd548d49af851182/tests/plots/test_image.py#L25-L31

The  `assert np.array_equal(val, image_data)` could be added to that test

tests/test_main.py
```diff
@@ -440,6 +440,43 @@ def checkpoint_never_exists(path):
...
+
+    val = np.asarray(ImagePIL.open(plot_file))
+    assert np.array_equal(val, image_data)
+
+
+def test_log_image_path(tmp_dir):
```
daavoo, line 459: I would move this to https://github.com/iterative/dvclive/blob/main/tests/plots/test_image.py

src/dvclive/live.py
```diff
@@ -269,6 +269,11 @@ def log_image(self, name: str, val):
         if not Image.could_log(val):
             raise InvalidDataTypeError(name, type(val))
 
+        if isinstance(val, str) or isinstance(val, Path):
+            from PIL import Image as ImagePIL
+
+            val = ImagePIL.open(val)
```
daavoo, line 275: I would move this to:

https://github.com/iterative/dvclive/blob/1cb31aa2f906396c1a32f1e0bd548d49af851182/src/dvclive/plots/image.py#L25-L30

src/dvclive/plots/image.py
daavoo: > we don't check in the if val.__class__.__module__ == "PIL.Image" that is has the right format?

I think is ok to let PIL handle the checks. Their errors are already explicit

daavoo (approved):
Minor comments about restructuring. I could do those myself in a follow-up

shcheklein:
@daavoo I'll do those changes. Thanks for the review 🙏

shcheklein:
@daavoo @dberenbaum I've addressed the review comments and created a doc update https://github.com/iterative/dvc.org/pull/4463

shcheklein:
@daavoo can I do a release?

daavoo:
> @daavoo can I do a release?

sure

shcheklein:
@daavoo do you draft them manually?

daavoo:
> @daavoo do you draft them manually?

Yes. https://github.com/iterative/dvclive/releases/new -> create a new version tag -> `Generate release notes`

PR #521, opened 2023-04-11
Support `artifacts` section

### LINKED ISSUE #519: make_dvcyaml: Create `artifacts` section
Per https://github.com/iterative/dvc/pull/9220

daavoo:
Closes #519

### DIFF
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -107,6 +107,14 @@ def make_dvcyaml(live):
     if plots:
         dvcyaml["plots"] = plots
 
+    if live._artifacts:
+        dvcyaml["artifacts"] = live._artifacts
+        for artifact in dvcyaml["artifacts"].values():
+            abs_path = os.path.realpath(artifact["path"])
+            abs_dir = os.path.realpath(live.dir)
+            relative_path = os.path.relpath(abs_path, abs_dir)
+            artifact["path"] = Path(relative_path).as_posix()
+
     dump_yaml(dvcyaml, live.dvc_file)
 
 
@@ -164,9 +172,10 @@ def get_dvc_stage_template(live):
         "cmd": "<python my_code_file.py my_args>",
         "deps": ["<my_code_file.py>"],
     }
-    if live._outs:
+    if live._artifacts:
         stage["outs"] = []
-    for o in live._outs:
+    for artifact in live._artifacts.values():
+        o = artifact["path"]
         artifact_path = Path(os.getcwd()) / o
         artifact_path = artifact_path.relative_to(live._dvc_repo.root_dir).as_posix()
         stage["outs"].append(artifact_path)
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -60,7 +60,7 @@ class Live:
         self._images: Dict[str, Any] = {}
         self._params: Dict[str, Any] = {}
         self._plots: Dict[str, Any] = {}
-        self._outs: Set[StrPath] = set()
+        self._artifacts: Dict[str, Dict] = {}
         self._inside_with = False
         self._dvcyaml = dvcyaml
 
@@ -321,19 +321,42 @@ class Live:
         """Saves the given parameter value to yaml"""
         self.log_params({name: val})
 
-    def log_artifact(self, path: StrPath):
+    def log_artifact(
+        self,
+        path: StrPath,
+        type: Optional[str] = None,  # noqa: A002
+        name: Optional[str] = None,
+        desc: Optional[str] = None,  # noqa: ARG002
+        labels: Optional[List[str]] = None,  # noqa: ARG002
+        meta: Optional[Dict[str, Any]] = None,  # noqa: ARG002
+    ):
         """Tracks a local file or directory with DVC"""
         if not isinstance(path, (str, Path)):
             raise InvalidDataTypeError(path, type(path))
 
         if self._dvc_repo is not None:
+            from dvc.repo.artifacts import name_is_compatible
+
             try:
                 stage = self._dvc_repo.add(path)
             except Exception as e:  # noqa: BLE001
                 logger.warning(f"Failed to dvc add {path}: {e}")
                 return
 
-            self._outs.add(path)
+            name = name or Path(path).stem
+            if name_is_compatible(name):
+                self._artifacts[name] = {
+                    k: v
+                    for k, v in locals().items()
+                    if k in ("path", "type", "desc", "labels", "meta") and v is not None
+                }
+            else:
+                logger.warning(
+                    "Can't use '%s' as artifact name (ID)."
+                    " It will not be included in the `artifacts` section.",
+                    name,
+                )
+
             dvc_file = stage[0].addressing
 
             if self._save_dvc_exp:
--- a/tests/test_log_artifact.py
+++ b/tests/test_log_artifact.py
@@ -1,4 +1,5 @@
 from dvclive import Live
+from dvclive.serialize import load_yaml
 
 
 def test_log_artifact(tmp_dir, dvc_repo):
@@ -44,3 +45,61 @@ def test_log_artifact_with_save_dvc_exp(tmp_dir, mocker, mocked_dvc_repo):
         include_untracked=[live.dir, "data", ".gitignore"],
         force=True,
     )
+
+
+def test_log_artifact_type_model(tmp_dir, mocked_dvc_repo):
+    (tmp_dir / "model.pth").touch()
+
+    with Live() as live:
+        live.log_artifact("model.pth", type="model")
+
+    assert load_yaml(live.dvc_file) == {
+        "artifacts": {"model": {"path": "../model.pth", "type": "model"}}
+    }
+
+
+def test_log_artifact_type_model_provided_name(tmp_dir, mocked_dvc_repo):
+    (tmp_dir / "model.pth").touch()
+
+    with Live() as live:
+        live.log_artifact("model.pth", type="model", name="custom")
+
+    assert load_yaml(live.dvc_file) == {
+        "artifacts": {"custom": {"path": "../model.pth", "type": "model"}}
+    }
+
+
+def test_log_artifact_type_model_on_step(tmp_dir, mocked_dvc_repo):
+    (tmp_dir / "model.pth").touch()
+
+    with Live() as live:
+        for _ in range(3):
+            live.log_artifact("model.pth", type="model")
+            live.next_step()
+        live.log_artifact("model.pth", type="model", labels=["final"])
+    assert load_yaml(live.dvc_file) == {
+        "artifacts": {
+            "model": {"path": "../model.pth", "type": "model", "labels": ["final"]},
+        },
+        "metrics": ["metrics.json"],
+    }
+
+
+def test_log_artifact_attrs(tmp_dir, mocked_dvc_repo):
+    (tmp_dir / "model.pth").touch()
+
+    attrs = {
+        "type": "model",
+        "name": "foo",
+        "desc": "bar",
+        "labels": ["foo"],
+        "meta": {"foo": "bar"},
+    }
+    with Live() as live:
+        live.log_artifact("model.pth", **attrs)
+    attrs.pop("name")
+    assert load_yaml(live.dvc_file) == {
+        "artifacts": {
+            "foo": {"path": "../model.pth", **attrs},
+        }
+    }


src/dvclive/live.py
```diff
@@ -316,7 +323,7 @@ def log_param(self, name: str, val: ParamLike):
         """Saves the given parameter value to yaml"""
         self.log_params({name: val})
 
-    def log_artifact(self, path: StrPath):
+    def log_artifact(self, path: StrPath, **kwargs):
```
dberenbaum, line 326: Why not include `type` and `name` as explicit kwargs?

dberenbaum: And maybe even `desc` and `labels`?

dberenbaum:
Thanks @daavoo!

Questions I have:
* Not a blocker, but this reminds me we need to revisit the lightning callback since it's most popular but doesn't auto-log the model artifact.
* Do we know what the artifact names look like for each framework? Should we be inferring them from the path or trying to come up with some better name for them?
* If we log the artifact at the end of each step/epoch, will it raise an error because another artifact exists with the same name?

src/dvclive/live.py
daavoo: I added explicit names. The reasons were that it was less code and `type` is a reserved keyword in Python, but neither are strong reasons

dberenbaum:
> * Do we know what the artifact names look like for each framework? Should we be inferring them from the path or trying to come up with some better name for them?

This would be a breaking change, so can we at least see what the artifact names look like and make a decision on which way to go here?

dberenbaum:
@daavoo We agreed that we would drop the callback support and make this depend on explicit calls to `log_artifact` for now, right? What's the status?

src/dvclive/live.py
```diff
@@ -328,6 +343,24 @@ def log_artifact(self, path: StrPath):
                 logger.warning(f"Failed to dvc add {path}: {e}")
                 return
 
+            if type == "model":
+                if path not in self._models:
+                    self._models[path] = {}
+                    if name is None:
+                        name = get_unmatched_name_from_path(
+                            path, [m.get("name") for m in self._models.values()]
+                        )
```
aguschin, line 352: If I written to `dvclive/dvc.yaml` already, I assume `name`s from there won't make it here, right?

Cautious that they can break the workflow: you log a single model, but the first time you run the script you create `model.pkl`, next time `models/model.pkl`, next time `dvclive/models/model.pkl`, until you run out of names.

src/dvclive/utils.py
```diff
@@ -127,3 +129,19 @@ def inside_notebook() -> bool:
 
         return IPython.__version__ >= "6.0.0"
     return False
+
+
+def get_unmatched_name_from_path(input_path, existing_names):
+    input_path = Path(input_path)
+    name = input_path.stem
```
aguschin, line 136: For now we have a regex checking for names https://github.com/iterative/dvc/blob/main/dvc/repo/artifacts.py#L16

we need to make sure dvclive can't produce names that will be invalidated by that regexp. Depending on when you want to merge this, we can revisit this later.

aguschin (approved):
Looks great overall (and simple - I like that very much after mlem/dvc codebases 😅 ). I see that Dave already posted feedback - have only one few items to add on top of that.

src/dvclive/live.py
daavoo: > If I written to dvclive/dvc.yaml already, I assume names from there won't make it here, right?

We don't load the existing `dvclive/dvc.yaml` on instantiation and we overwrite it with the data logged on the current runtime.

This is not only because of `artifacts` but also makes sense for other sections. For example, a user might remove the code that logs parameters so keeping that section in `dvclive/dvc.yaml` would not make sense/break.

> Cautious that they can break the workflow: you log a single model, but the first time you run the script you create model.pkl, next time models/model.pkl, next time dvclive/models/model.pkl, until you run out of names.

As per the above explanation, this should not happen unless the user calls `live.log_artifact` with the same name inside the same script, at which point I don't think we should try to work around it but consider it a misuse of the API.

src/dvclive/utils.py
daavoo: DVCLive now requires DVC Python API so I think I would update the P.R. to import the name check function from DVC and use it here. Does that sound good to you?

daavoo:
> @daavoo We agreed that we would drop the callback support and make this depend on explicit calls to `log_artifact` for now, right? What's the status?

Dropped from the P.R.

src/dvclive/dvc.py
```diff
@@ -107,6 +107,14 @@ def make_dvcyaml(live):
...
         dvcyaml["plots"] = plots
 
+    if live._models:
+        dvcyaml["artifacts"] = {}
+        for path, attrs in live._models.items():
+            attrs["path"] = path
+            dvcyaml["artifacts"][attrs.get("name")] = {
+                k: v for k, v in attrs.items() if k != "name"
+            }
+
```
dberenbaum, line 117: Why are we not keying `_models` by `name`? Then I think this becomes a one-line like `dvcyaml["artifacts"] = live._models`.

src/dvclive/utils.py
```diff
@@ -125,3 +127,19 @@ def inside_notebook() -> bool:
 
         return IPython.__version__ >= "6.0.0"
     return False
+
+
+def get_unmatched_name_from_path(input_path, existing_names):
```
dberenbaum, line 132: Do we need it to be "unmatched"? For example, if I'm doing my own checkpointing, I might repeatedly call `log_artifact` with the same path and I'd expect it to overwrite each time.

src/dvclive/dvc.py
daavoo: > Why are we not keying _models by name? Then I think this becomes a one-line like dvcyaml["artifacts"] = live._models.

I had it like that and then had to change it, but I don't recall the reason 😅 . 
I will take an extra look

src/dvclive/utils.py
daavoo: I think this was the reason for not using `name` as the key  (answer in the above comment 😅) .

What you describe works as expected in the current P.R. because we use `path` as the key and then check:

```
if path not in self._models:
```

To decide whether a `name` needs to be generated or not

src/dvclive/live.py
```diff
@@ -333,6 +348,24 @@ def log_artifact(self, path: StrPath):
                 logger.warning(f"Failed to dvc add {path}: {e}")
                 return
 
+            if type == "model":
```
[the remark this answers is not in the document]
daavoo: We could, I don't know if we should. 
My motivation was to not include unnecessary/unrelated things in the P.R.

It would not be that hard to extend in the future if/when we need it.

src/dvclive/utils.py
daavoo: To clarify, we could achieve a similar thing using `name`, but the code felt worse to me.
Perhaps I am biased because I still feel that `path` made more sense as primary key in `artifacts` 😅

src/dvclive/live.py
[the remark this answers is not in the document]
dberenbaum: I guess I worry it's harder to document this way. How do you explain what the other artifact metadata args do?

Not a blocker.

src/dvclive/utils.py
dberenbaum: Yeah, it feels to me like the code would be simpler if we use the name key and also don't care about unmatched names. It could infer a name if needed, then add whatever metadata is provided to `_models` as is, and write out `_models` as is to the artifacts section of `dvc.yaml`.

Also not really a blocker, but maybe worth a look since I wonder if the current implementation is biased too much by all the previous discussions and we will struggle to remember why it looks this way when we need to touch it in the future.

daavoo: > also don't care about unmatched names

Do you mean that we should not worry about:

```py
with Live() as live:
    live.log_artifact("classification/model.pth", type="model")
    live.log_artifact("detection/model.pth", type="model")
```

src/dvclive/live.py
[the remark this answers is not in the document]
daavoo: > How do you explain what the other artifact metadata args do?

Write to `.dvc` file?

src/dvclive/utils.py
daavoo: Updated the P.R. to use `name` as primary key

src/dvclive/live.py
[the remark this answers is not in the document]
daavoo: Side note. 

`artifacts` section has a restriction on `names` formatting that is only applicable because of the model registry + `git tag` usage.

It feels harder to document that we would try to write to `artifacts` except if the `name` we found or the one you provide doesn't fulfill a restriction that doesn't apply to your use case.

src/dvclive/utils.py
daavoo: > Updated the P.R. to use `name` as primary key

TBH, it feels to me that we are moving the code from one place to another, not really simplifying

src/dvclive/live.py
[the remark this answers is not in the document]
dberenbaum: > They are written to .dvc file? (I updated the P.R. to do that)

I don't think we should have two different places to add these fields. I'd rather deprecate putting them into `.dvc` files like we are doing with `artifacts.yaml`. This feels like we are only partially committing to the new MR workflow. Is there some hesitation to rely on this new `artifacts` section?

> It feels harder to document that we would try to write to `artifacts` except if the `name` we found or the one you provide doesn't fulfill a restriction that doesn't apply to your use case.

What other use cases do you have in mind?

src/dvclive/dvc.py
```diff
@@ -177,3 +181,29 @@ def get_dvc_stage_template(live):
...
+    else:
+        name = attrs.pop("name")
+
+    attrs["path"] = path
+    if name_is_compatible(name):
+        if name not in models:
+            models[name] = {}
+        for k, v in attrs.items():
+            if v is not None:
+                models[name][k] = v
```
dberenbaum, line 200: I think we should overwrite and leave this up to the user to manage, so we could reduce this to `models[name] = attrs`.

src/dvclive/dvc.py
```diff
@@ -177,3 +181,29 @@ def get_dvc_stage_template(live):
...
+    else:
+        logger.warning(
+            "Can't use '%s' as artifact name (ID)."
+            " You can use letters and numbers, and use '-' as separator"
+            " (but not at the start or end)."
+            " The first character must be a letter.",
+            name,
```
dberenbaum, line 207: Yeah, this looks bad unless we can condense the logic more 😓 . @aguschin WDYT? Can we explain it better? Do we really need it to be this convoluted?

src/dvclive/dvc.py
```diff
@@ -177,3 +181,29 @@ def get_dvc_stage_template(live):
     dvcyaml = output.getvalue()
     output.close()
     return dvcyaml
+
+
+def update_models(path, attrs, models):
+    from dvc.repo.artifacts import name_is_compatible
+
+    if attrs["name"] is None:
+        name = find_unused_name(path, models)
```
dberenbaum, line 190: I think we can drop `find_unused_name`. If users are worried about conflicts, they can explicitly set a name.

src/dvclive/utils.py
dberenbaum: Thanks for trying it this way @daavoo. Left some comments about where I think we can simplify.

src/dvclive/dvc.py
daavoo: The thing is that `attrs` here is coming from the explicit optional arguments in `log_artifact` that we have for each of the attributes. 

So, here or elsewhere we need a loop like this to discard `None` values, unless we are ok with dumping to `artifacts.yaml` having attributes set to None.

(Not having to handle this was one of the "pros" of `kwargs`)

src/dvclive/dvc.py
daavoo: I included it because it was explicitly mentioned as a requirement in our last sync with @aguschin and @omesser .
I am happy to go ahead and remove code if there are no strong objections

daavoo:
> Sorry for so many iterations here @daavoo. The blocker left for me is what to do with attrs if there isn't type=model. IMO we should write any attrs to artifacts. It's simple and even if it doesn't do much now, neither does anything else we do with those attrs (at least it could be useful in the future and an artifact registry would work with the existing artifacts written to artifacts).

Pushed a new version

dberenbaum (approved):

tests/test_log_artifact.py
```diff
@@ -44,3 +45,61 @@ def test_log_artifact_with_save_dvc_exp(tmp_dir, mocker, mocked_dvc_repo):
...
+
+
+def test_log_artifact_type_model(tmp_dir, mocked_dvc_repo):
+    (tmp_dir / "model.pth").touch()
+
+    with Live() as live:
+        live.log_artifact("model.pth", type="model")
+
+    assert load_yaml(live.dvc_file) == {
+        "artifacts": {"model": {"path": "model.pth", "type": "model"}}
```
daavoo, line 57: TODO: Fix the `path` here.
It should be written as relative to the `dvclive/dvc.yaml`

daavoo: I pushed a fix

daavoo:
Don't see much point in merging until we test end-to-end against Studio.

aguschin:
@daavoo, looks like it will be working, but if you want me, I can test this against what I have now. WDYT?

daavoo:
> @daavoo, looks like it will be working, but if you want me, I can test this against what I have now. WDYT?

Sounds good, what just saying that we can wait for merge and release until it works with Studio prod

dberenbaum:
Can we merge and push people to start using it now, letting them know what's coming? We have people today asking to get started with the model registry. I'd like to tell them to use this and their models will appear once Studio is updated. Otherwise, I can only tell them do it the old way but it will break soon, or they can't start using it at all.

PR #529, opened 2023-04-14
studio: Save experiment before posting `done` event.

daavoo:
To ensure `experiment_rev` is sent.

Found that it was the underlying issue of https://github.com/iterative/studio/issues/5673#issuecomment-1508480082. I thought Studio relied on `baseline_sha`, `exp_name` pair to match live experiments with experiment ref.

Regardless, this was a plain bug that had to be fixed and went under the radar.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -388,6 +388,8 @@ class Live:
 
         self._ensure_paths_are_tracked_in_dvc_exp()
 
+        self.save_dvc_exp()
+
         if "done" not in self._studio_events_to_skip:
             response = False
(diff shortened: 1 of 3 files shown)


src/dvclive/live.py
```diff
@@ -388,6 +388,8 @@ def end(self):
 
         self._ensure_paths_are_tracked_in_dvc_exp()
 
+        self.save_dvc_exp()
```
daavoo, line 391: I am on a streak of this kind of bugfixes

dberenbaum (approved):

PR #530, opened 2023-04-14
Add DVCLive-Quickstart

daavoo:
To be used in Colab

Test in:

https://colab.research.google.com/github/iterative/dvclive/blob/colab/examples/DVCLive-Quickstart.ipynb

### DIFF
--- a/examples/DVCLive-Quickstart.ipynb
+++ b/examples/DVCLive-Quickstart.ipynb
@@ -55,7 +55,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Helper code"
+    "## Setup code"
    ]
   },
   {
@@ -64,6 +64,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "#@title Training helpers. { display-mode: \"form\" }\n",
+    "\n",
     "import numpy as np\n",
     "import torch\n",
     "import torch.nn.functional as F\n",
@@ -172,6 +174,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "#@title Initialize model and dataset. { display-mode: \"form\" }\n",
+    "\n",
     "model = torch.nn.Sequential(\n",
     "    torch.nn.Flatten(),\n",
     "    torch.nn.Linear(28 * 28, 128),\n",
--- a/examples/DVCLive-Quickstart.ipynb
+++ b/examples/DVCLive-Quickstart.ipynb
@@ -1,10 +1,11 @@
 {
  "cells": [
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<a href=\"https://colab.research.google.com/github/iterative/dvclive/blob/main/examples/DVCLive_Quickstart.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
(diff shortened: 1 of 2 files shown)


examples/DVCLive-Quickstart.ipynb
```diff
@@ -0,0 +1,354 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<a href=\"https://colab.research.google.com/github/iterative/dvclive/blob/main/examples/DVCLive-Quickstart.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
daavoo, line 8: This link would work once we merge on the main branch, in the meantime use  https://colab.research.google.com/github/iterative/dvclive/blob/colab/examples/DVCLive-Quickstart.ipynb

examples/DVCLive-Quickstart.ipynb
```diff
@@ -0,0 +1,354 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# You can modify these parameters to see how they affect the training\n",
+    "# And run the cell several times\n",
```
daavoo, line 214: Maybe we should include a `for` loop running multiple experiments by default

dberenbaum:
Do you think we could use example-get-started-experiments for this?

shcheklein:
I love the idea and direction and how fast you are improving it, @daavoo . I'm even fine to merge this :)

> Do you think we could use example-get-started-experiments for this?

But I agree on this also. We should try to use the same project. It reduces the surface of things to maintain and remember for us.

dberenbaum:
I think having a colab notebook for example-get-started-experiments would be nice to have anyway.

daavoo:
> I think having a colab notebook for example-get-started-experiments would be nice to have anyway.
> 
> Edit: agreed it's a minor note though and this looks good in general to me.

I did one for the blog post release: 

https://colab.research.google.com/drive/1NTivljRYiySMJn-SHeWQSycBmSOVUbvA

What is hard is to make it "based on"/"linked to" the one from the GitHub repo, as this quickstart, because workflows are just too different in Colab. The notebook would have to be filled with conditions to alter the workflow depending on whether you are executing it in Colab or in a different context like a codespace or local machine.

This quickstart works fine because we don't really expect to also use it in a real git-based workflow. It is a one-time visit where you discard the end results. 

To make it work git-based, you need to run `git clone` in colab, `cd` to the repo, somehow dump the current notebook changes to a file and track it with git, etc. It just becomes nonsense.

dberenbaum:
> I did one for the blog post release:
> 
> https://colab.research.google.com/drive/1NTivljRYiySMJn-SHeWQSycBmSOVUbvA

LGTM. My suggestion was about reusing the use case, not tracking the colab notebook in the repo, and I think that notebook already does a good enough job of it. Can we combine these two notebooks so you use the experiment-get-started use case but highlight the DVCLive stuff and hide the irrelevant stuff?

dberenbaum:
We could optionally add it to example-repos-dev and/or example-get-started-experiments to keep changes in sync, but it the notebook itself doesn't need to "depend" on the repo.

daavoo:
> We could optionally add it to example-repos-dev and/or example-get-started-experiments to keep changes in sync, but it the notebook itself doesn't need to "depend" on the repo.

I was actually thinking on adding it to `dvclive/examples`, and making that folder a little about `Colab` intro, as `example-repos-dev` is more about reproducing and generating a full git repo.

daavoo:
Merging this one, lets discuss improvements and ideas as follow ups

PR #533, opened 2023-04-18
studio: Support `DVC_` env vars.

daavoo:
Per https://github.com/iterative/dvc-studio-client/pull/28

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -29,6 +29,7 @@ package_dir=
 packages = find:
 install_requires=
     dvc>2.45.1
+    dvc-studio-client>=0.7.0,<1
     funcy
     ruamel.yaml
 [options.extras_require]
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -5,7 +5,7 @@ import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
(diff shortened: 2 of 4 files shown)


skshetry:
Should we wait for dvc to release? I am worried if we will get into some conflicts here.

skshetry (approved):

skshetry:
Maybe not. 
https://github.com/iterative/dvc/blob/dd3deb6cbd00ab19f60fb63d7a3ab154242a9826/pyproject.toml#L40

daavoo:
> Maybe not. https://github.com/iterative/dvc/blob/dd3deb6cbd00ab19f60fb63d7a3ab154242a9826/pyproject.toml#L40

wait, how does your DVC P.R. work without bumping?

skshetry:
As dvc has to make decisions also based on config, there was no point to spread out logic into two places.

```python
if not token := (
      os.environ.get(DVC_STUDIO_TOKEN)
      or os.environ.get("STUDIO_TOKEN")
      or config.get("token")
):
  pass
```

If dvclive has to do the same, it might be better to use dvc instead of client (I know you want to reduce dependency on dvc, but we have more bigger issues than Studio client there).

skshetry:
Also using env variables are simple enough, that I am fine with whatever you decide: either importing it or redeclaring here.

PR #543, opened 2023-04-24
live: Add `log_plot`.

### LINKED ISSUE #271: log_plot(): add custom line plot

### LINKED ISSUE #453: sklearn: show custom names in title
Right now, if you use custom names to make multiple ROC curves or other sklearn plots, there is no way to distinguish between the plots:

Screenshot 2023-02-03 at 3 03 50 PM

For `live.log_sklearn_plot("roc", labels, predictions, name=f"roc/test")`, it would be nice for the title (both in the report and in `dvc.yaml`) to be be something like `roc/test::Receiver operating characteristic (ROC)`.

daavoo:
Create DVC plots from datapoints (list of dictionaries) and plot config.

Closes #271
Closes #453

```py
datapoints = [{"foo": 1, "bar": 2}, {"foo": 3, "bar": 4}]
with Live() as live:
    live.log_plot("foo_default", datapoints, x="foo", y="bar", title="Default Linear Plot")
    live.log_plot(
        "foo_scatter",
        datapoints,
        x="foo",
        y="bar",
        template="scatter",
        title="Plot using Scatter template",
        y_label="CUSTOM LABEL: BAR",
        x_label="CUSTOM LABEL: FOO"
    )
```

Example output:

- `dvclive/plots/custom/foo_default.json`

```json
[
    {
        "foo": 1,
        "bar": 2
    },
    {
        "foo": 3,
        "bar": 4
    }
]                                                               
```

- `dvclive/dvc.yaml`

```yaml
plots:
- plots/custom/foo_default.json:
    x: foo
    y: bar
    title: Default Linear Plot
- plots/custom/foo_scatter.json:
    template: scatter
    x: foo
    y: bar
    title: Plot using Scatter template
    x_label: 'CUSTOM LABEL: FOO'
    y_label: 'CUSTOM LABEL: BAR'
```

- `dvc plots show`

Captura de pantalla 2023-04-26 a las 20 16 51

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -26,7 +26,7 @@ from .error import (
     InvalidPlotTypeError,
     InvalidReportModeError,
 )
-from .plots import PLOT_TYPES, SKLEARN_PLOTS, Image, Metric, NumpyEncoder
+from .plots import PLOT_TYPES, SKLEARN_PLOTS, CustomPlot, Image, Metric, NumpyEncoder
 from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
 from .studio import get_studio_updates
@@ -261,15 +261,15 @@ class Live:
             raise InvalidDataTypeError(name, type(val))
 
         if name in self._metrics:
-            data = self._metrics[name]
+            metric = self._metrics[name]
         else:
-            data = Metric(name, self.plots_dir)
-            self._metrics[name] = data
+            metric = Metric(name, self.plots_dir)
+            self._metrics[name] = metric
 
-        data.step = self.step
-        data.dump(val, timestamp=timestamp)
+        metric.step = self.step
+        metric.dump(val, timestamp=timestamp)
 
-        self.summary = set_in(self.summary, data.summary_keys, val)
+        self.summary = set_in(self.summary, metric.summary_keys, val)
         logger.debug(f"Logged {name}: {val}")
 
     def log_image(self, name: str, val):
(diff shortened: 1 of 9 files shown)


src/dvclive/plots/sklearn.py
```diff
@@ -117,24 +95,18 @@ def dump(self, val, **kwargs) -> None:
...
-
-    def get_properties(self):
-        properties = copy.deepcopy(self.DEFAULT_PROPERTIES)
-        if self.normalized:
-            properties["template"] = "confusion_normalized"
-        return properties
+    def __init__(self, name: str, output_folder: str, **plot_config) -> None:
+        plot_config["template"] = (
+            "confusion_normalized"
+            if plot_config.pop("normalized", None)
```
daavoo, line 101: We already released and put on docs so I kept this, but the same could now be achieved with:

```py
live.log_sklearn_plot(
    "confusion_matrix",
    [0, 0, 1, 1],
    [0, 1, 1, 0],
    name="confusion_matrix_normalized",
    template="confusion_normalized"
)
```

src/dvclive/live.py
```diff
@@ -282,29 +282,70 @@ def log_image(self, name: str, val):
...
+        image.step = self.step
+        image.dump(val)
         logger.debug(f"Logged {name}: {val}")
 
+    def log_plot(
+        self,
+        name: str,
+        datapoints: List[Dict],
+        x: str,
+        y: str,
```
dberenbaum, line 299: Should we allow for `y` to be a list?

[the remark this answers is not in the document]
dberenbaum: DVC CLI also supports numeric indexing. Should we allow it here? This would open up more possibilities for data types to accept, like 2d arrays.

dberenbaum (approved):
My biggest question is around what data types to support. List of dicts seems pretty limited. DVC CLI also supports tabular data, so it would be nice to support that here, as well as other dataframe-like or array-like inputs.

I don't think any of those changes would be breaking, so I'm fine to approve and follow up to support more input data formats. In the docs, it might help to give a dataframe example using ~`df.to_json(orient="records")`~ [edit: `df.to_dict(orient="records")`, which I think is what people would want to use here in a pandas example, right?

daavoo:
> My biggest question is around what data types to support. List of dicts seems pretty limited. DVC CLI also supports tabular data, so it would be nice to support that here, as well as other dataframe-like or array-like inputs.
> 
> I don't think any of those changes would be breaking, so I'm fine to approve and follow up to support more input data formats. In the docs, it might help to give a dataframe example using ~`df.to_json(orient="records")`~ [edit: `df.to_dict(orient="records")`], which I think is what people would want to use here in a pandas example, right?

It was a bit of a docs-driven decision to keep the first iteration simple, as explaining all the variations of input data and `x`/`y` behavior felt overcomplicated.

The other motivation was that any of those changes would be a breaking change, as you pointed out, so we can extend the API in the future to add support for those features.

PR #548, opened 2023-04-27
logging artifacts in optuna callback

tibor-mach:
Adding a method to log artifacts. It requires the user to [add user defined attributes](https://optuna.readthedocs.io/en/stable/reference/generated/optuna.trial.Trial.html#optuna.trial.Trial.set_user_attr) called `artifact_path` and (optionally) `artifact_name` to the optuna trial.

I am not very happy with that being necessary but I don't see any other way to integrate this functionality into the callback.

The functionality can be tested by following the README.md using  [this simple example](https://github.com/tibor-mach/dvclive-optuna)

### DIFF
--- a/src/dvclive/optuna.py
+++ b/src/dvclive/optuna.py
@@ -13,6 +13,16 @@ class DVCLiveCallback:
         with Live(save_dvc_exp=True, **self.live_kwargs) as live:
             self._log_metrics(trial.values, live)
             live.log_params(trial.params)
+            self._log_artifact(trial, live)
+
+    def _log_artifact(self, trial, live):
+        artifact_path = trial.user_attrs.get("artifact_path")
+        artifact_name = trial.user_attrs.get("artifact_name")
+
+        if artifact_path:
+            live.log_artifact(artifact_path, name=artifact_name)
+        elif artifact_name:
+            pass  # TODO Raise a warning? Perhaps the user forgot to specify the path...
 
     def _log_metrics(self, values, live):
         if values is None:


dberenbaum:
Thanks @tibor-mach! I agree it's inelegant but also don't see a more elegant way to include it in the callback.

What worries me is that the optuna callback seems useful only for very shallow use cases, and as soon as we start adding any customization like this, I would rather use the generic dvclive api within the objective function. Do you feel like there is value in this callback?

@daavoo WDYT?

tibor-mach:
@dberenbaum I was thinking about that too. I guess the callback's basic functionality is nice because all you really need is to add the callback to the study's optimize method - 5 seconds of work, and you get metrics for each trial in your experiment tracking.

However, whenever you want to do anything more complex than that, you are limited by the attributes available in the `trial` object...you can make workarounds like the one I have in my draft PR and keep adding custom attributes but perhaps it would be more helpful to prepare a few "dvclive optuna recipes" (example scripts like the ones optuna has [here](https://github.com/optuna/optuna-examples)).

I would keep the callback as it is, or maybe include the extra feature in my draft PR (but maybe even that is too much already, it still forces you to remember specific keywords and set up user defined attributes for your optuna trials to be of any use) and I would focus on providing such tutorial examples instead. A nice side effect of that is that it familiarizes users with the dvclive API, is a nice developer advocacy material for blogposts or videos and for many common use-cases it can still be more or less copy-paste. Not quite as comfortableas [`import antigravity`](https://xkcd.com/353/) but still quite good and the advantage is that it is much easier for the users to see what is actually going on (which I always like, because it allows me to change it without forking libraries and then maintaining those forks :-) )

dberenbaum:
I lean towards pushing users to use the underlying ML frameworks for this. For the default case, I don't know that users need all the artifacts. It could be surprisingly expensive to keep the model for each trial, and often users will not care about them because they may only be tuning on a subset of data and need to retrain the final params on the full dataset.

daavoo:
> I would keep the callback as it is,

I agree with this. The behavior is in line with existing built-in  integrations in optuna (https://optuna.readthedocs.io/en/stable/reference/integration.html) where only metrics and parameters are being logged.

---

For anything beyond that, we should guide people towards manual use of the API or use the specific ML Framework callback inside the `objective` function, like:

> perhaps it would be more helpful to prepare a few "dvclive optuna recipes" (example scripts like the ones optuna has [here](https://github.com/optuna/optuna-examples)).

tibor-mach:
Agreed. I think [this](https://github.com/iterative/dvclive/pull/356#issuecomment-1548473971) is also relevant regarding the extra stuff @daavoo I would be interested in your opinion on that.

Unfortunately in order to work with the study object, it would still have to be a part of the callback, but the recipes could essentially contain tutorials how to create Callback classes which inherit from the base one that you made.

If Study is not needed then it can be done directly inside the objective function and without the callback, but it is a rather useful and important object in optuna...

PR #559, opened 2023-05-07
feat(log_artifact): add an option to copy (capture) into dvclive dir

### LINKED ISSUE #550: `log_artifact`: move model file (optionally or by default) into Live dir
In some (a lot?) of case models that are being produced are actually saved in some temporary locations, e.g. in case of YOLO it creates Git-ignored `runs/run15/weights/best.pt`. It doesn't make sense to keep model artifact there, it has to be moved somewhere. I think that we should have an option to move a model into `dvclive/models`.

I can make a PR for this. Just want to run by you @daavoo @dberenbaum @aguschin and see what you think.

shcheklein:
Closes #550

```python
    with Live() as live:
        live.log_artifact("model.pth", copy=True)
```

It copies a file into `dvclive/artifacts/model.pth` and tracks it with DVC in a regular way.

---------

- [X] ❗ I have followed the [Contributing to DVCLive](https://github.com/iterative/dvclive/blob/master/CONTRIBUTING.md) guide.

TODO (when accepted):

- [ ] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I  have created a separate PR (or issue, at least) in
  [dvc.org](https://github.com/iterative/dvc.org) and linked it here.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -30,7 +30,14 @@ from .plots import PLOT_TYPES, SKLEARN_PLOTS, CustomPlot, Image, Metric, NumpyEn
 from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
 from .studio import get_studio_updates
-from .utils import env2bool, inside_notebook, matplotlib_installed, open_file_in_browser
+from .utils import (
+    StrPath,
+    clean_and_copy_into,
+    env2bool,
+    inside_notebook,
+    matplotlib_installed,
+    open_file_in_browser,
+)
 
 logger = logging.getLogger("dvclive")
 handler = logging.StreamHandler()
@@ -40,7 +47,6 @@ logger.addHandler(handler)
 logger.setLevel(os.getenv(env.DVCLIVE_LOGLEVEL, "INFO").upper())
 
 ParamLike = Union[int, float, str, bool, List["ParamLike"], Dict[str, "ParamLike"]]
-StrPath = Union[str, Path]
 
 
 class Live:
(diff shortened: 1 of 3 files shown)


src/dvclive/live.py
```diff
@@ -388,8 +393,15 @@ def log_artifact(
         if self._dvc_repo is not None:
             from dvc.repo.artifacts import name_is_compatible
 
+            if copy:
+                basename = os.path.basename(os.path.normpath(path))
+                output_dir = Path(self.artifacts_dir)
+                output_dir.mkdir(exist_ok=True, parents=True)
+                shutil.copy2(path, output_dir / basename)
```
daavoo, line 400: We should also support directories here or, if we don't want to, log warning and not fail

daavoo (changes requested):

src/dvclive/live.py
```diff
@@ -380,6 +384,7 @@ def log_artifact(
         desc: Optional[str] = None,  # noqa: ARG002
         labels: Optional[List[str]] = None,  # noqa: ARG002
         meta: Optional[Dict[str, Any]] = None,  # noqa: ARG002
+        copy: bool = False,
```
dberenbaum, line 387: What's the path that's being copied in your scenario, and what would happen if you logged it with `copy=False`? Could we infer whether to copy?

shcheklein: > What's the path that's being copied in your scenario

The path that you pass to the logger: `log_artifact('some_path')`

> and what would happen if you logged it with copy=False?

Same behavior as before. `.dvc` created nearby that `some_path` and in `dvc.yaml` we log it as `some_path`.

> Could we infer whether to copy?

not sure tbh, sounds a bit too much magic / complicated? Probably we would need a bit more data to decide.

src/dvclive/live.py
shcheklein: yep, my bad - I though it's supported by `copy2`. Looking into this ...

src/dvclive/live.py
dberenbaum: > > What's the path that's being copied in your scenario
> 
> The path that you pass to the logger: `log_artifact('some_path')`

Sorry, I meant for the example you are working on that prompted this PR. What is the exact path you want to log?

I'm wondering if it's inside/outside the repo, gitignored, etc.

daavoo (approved):

PR #539, opened 2023-04-21
Auto-initialize DVC repo

dberenbaum:
Related to #531. Do you think it's enough @daavoo?

Example with no Git repo (with or without a DVC repo):

```
WARNING:dvclive:Can't save experiment without a Git Repo.
Create a Git repo (`git init`) and commit (`git commit`).
```

Example with an empty Git repo and no DVC repo:

```
Initialized DVC repository.

INFO:dvc.repo.init:Initialized DVC repository.

You can now commit the changes to git.

INFO:dvc.repo.init:You can now commit the changes to git.

WARNING:dvclive:Can't save experiment to an empty Git Repo.
Add a commit (`git commit`) to save experiments.
INFO:dvclive:To run with DVC, add this to /private/tmp/repo/dvc.yaml:
stages:
  dvclive:
    cmd: <python my_code_file.py my_args>
    deps:
    - <my_code_file.py>
```

Example with an empty Git repo and an existing DVC repo:

```
WARNING:dvclive:Can't save experiment to an empty Git Repo.
Add a commit (`git commit`) to save experiments.
INFO:dvclive:To run with DVC, add this to /private/tmp/repo/dvc.yaml:
stages:
  dvclive:
    cmd: <python my_code_file.py my_args>
    deps:
    - <my_code_file.py>
```

Example with a non-empty Git repo and no DVC repo:

```
Initialized DVC repository.

INFO:dvc.repo.init:Initialized DVC repository.

You can now commit the changes to git.

INFO:dvc.repo.init:You can now commit the changes to git.

INFO:dvclive:To run with DVC, add this to /private/tmp/repo/dvc.yaml:
stages:
  dvclive:
    cmd: <python my_code_file.py my_args>
    deps:
    - <my_code_file.py>
```

Edit: also fixes #522

PR #561, opened 2023-05-08
read studio token from dvc config

dberenbaum:
Reads studio token from dvc config and posts live metrics if found

Edit: for reference, context is in https://iterativeai.slack.com/archives/C01SR9Q12LB/p1683213358701869

copilot:all

### DIFF
--- a/.dvc/.gitignore
+++ b/.dvc/.gitignore
@@ -1,3 +1,0 @@
-/config.local
-/tmp
-/cache
--- a/.dvc/.gitignore
+++ b/.dvc/.gitignore
@@ -1,0 +1,3 @@
+/config.local
+/tmp
+/cache
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -5,7 +5,7 @@ import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
-from dvc_studio_client.post_live_metrics import get_studio_config, post_live_metrics
+from dvc_studio_client.post_live_metrics import post_live_metrics
 from funcy import set_in
 from pathspec import PathSpec
 from ruamel.yaml.representer import RepresenterError
@@ -28,7 +28,7 @@ from .error import (
 from .plots import PLOT_TYPES, SKLEARN_PLOTS, CustomPlot, Image, Metric, NumpyEncoder
 from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
-from .studio import get_studio_updates
+from .studio import get_dvc_studio_config, get_studio_updates
 from .utils import env2bool, inside_notebook, matplotlib_installed, open_file_in_browser
 
 logger = logging.getLogger("dvclive")
@@ -80,15 +80,13 @@ class Live:
         self._exp_name: Optional[str] = None
         self._experiment_rev: Optional[str] = None
         self._inside_dvc_exp: bool = False
-        self._dvc_repo = None
+        self._dvc_repo = get_dvc_repo()
         self._include_untracked: List[str] = []
         self._init_dvc()
 
-        self._dvc_config = {}
-        if self._dvc_repo:
-            self._dvc_config = self._dvc_repo.config.get("studio")
         self._latest_studio_step = self.step if resume else -1
         self._studio_events_to_skip: Set[str] = set()
+        self._dvc_studio_config: Dict[str, Any] = get_dvc_studio_config(self)
         self._init_studio()
 
     def _init_resume(self):
@@ -125,7 +123,6 @@ class Live:
                 )
                 self._save_dvc_exp = False
 
-        self._dvc_repo = get_dvc_repo()
         if self._dvc_repo is None:
             if self._save_dvc_exp:
                 logger.warning(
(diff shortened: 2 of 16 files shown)


src/dvclive/live.py
```diff
@@ -151,7 +152,12 @@ def _init_dvc(self):
             self._include_untracked.append(self.dir)
 
     def _init_studio(self):
-        if not any((os.getenv(STUDIO_TOKEN, None), os.getenv(DVC_STUDIO_TOKEN, None))):
+        self._studio_token = os.getenv(STUDIO_TOKEN, None) or os.getenv(
+            DVC_STUDIO_TOKEN, None
+        )
+        if self._dvc_repo and not self._studio_token:
+            self._studio_token = self._dvc_repo.config.get("studio", {}).get("token")
```
shcheklein, line 159: can DVCLive now work without DVC repo? Outside a repo? 

We still could read the global / user config in that case I think.

dberenbaum: DVCLive will try to create a DVC repo now (see #539). It should only fail if it's outside a Git repo or the Git repo is empty. That's a valid use case but handling it is outside this PR IMO (to start, Studio would need to handle a non-Git project).

src/dvclive/studio.py
```diff
@@ -66,3 +68,11 @@ def get_studio_updates(live):
     plots = {k: {"data": v} for k, v in plots.items()}
 
     return metrics, params, plots
+
+
+def get_studio_token(repo):
```
daavoo, line 73: I wonder if this logic should be in dvc_studio_client

dberenbaum: What about in dvc? It needs the repo to get the right config value.

daavoo: >  It needs the repo to get the right config value.

I was thinking just accepting optional dvc_repo. If it is not None, assume is a DVC repo and try to read the config. Else try to get the env var

> What about in dvc?

It also works. It is just the same arguments about having dvc-studio-client at all.
If you put it in DVC, we need to pin and require a newer DVC version which is more restrictive for users that if we put it in the client and bumpt it here and in DVC

src/dvclive/live.py
dberenbaum: On 2nd thought, it shouldn't be that hard to support in the use case where I have a repo and project but want to run a specific experiment outside of the repo in some cloud managed cloud environment.

src/dvclive/studio.py
dberenbaum: Added https://github.com/iterative/dvc-studio-client/pull/35 to support this in both dvc and dvclive.

src/dvclive/live.py
dberenbaum: Done

dberenbaum: @daavoo What do you think about this scenario? I think it's disabled now if there's no DVC repo, but should we allow it for this kind of use case?

daavoo: > @daavoo What do you think about this scenario? I think it's disabled now if there's no DVC repo, but should we allow it for this kind of use case?

Sorry, I am not fully understanding the scenario. Could you elaborate?

dberenbaum: Let's say I have an existing Git repo and Studio project, but I run a specific experiment disconnected from my repo (in colab, databricks, etc.). I could set the token and repo url and still send live metrics, except that I think we currently disable this if no DVC repo is found. Should we allow it?

dberenbaum:
Not sure why this says there are conflicts when I have merged main into it. Otherwise it is ready for review.

.dvc/.gitignore
daavoo, line 1: @dberenbaum looks like these were unintentionally committed. Not sure how they got generated maybe there is some leak in tests?

src/dvclive/live.py
```diff
@@ -87,12 +86,13 @@ def __init__(
...
         self._experiment_rev: Optional[str] = None
         self._inside_dvc_exp: bool = False
-        self._dvc_repo = None
+        self._dvc_repo = get_dvc_repo()
         self._include_untracked: List[str] = []
         self._init_dvc()
 
         self._latest_studio_step = self.step if resume else -1
         self._studio_events_to_skip: Set[str] = set()
+        self._dvc_studio_config: Dict[str, Any] = get_dvc_studio_config(self)
```
daavoo, line 95: Same comment as above

.dvc/.gitignore
dberenbaum: Okay, looks like the same comment I made in #565 🤦

src/dvclive/live.py
```diff
@@ -87,12 +86,13 @@ def __init__(
         self._exp_name: Optional[str] = None
         self._experiment_rev: Optional[str] = None
         self._inside_dvc_exp: bool = False
-        self._dvc_repo = None
+        self._dvc_repo = get_dvc_repo()
```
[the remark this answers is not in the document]
dberenbaum: Sorry, just different code styles. I tend to lean towards a more functional code style where I try to avoid stateful changes within other class methods since it's less clear to me without searching the code when and how `self._dvc_repo` gets set.

Anyway, I changed it back here and below.

daavoo (approved):

PR #563, opened 2023-05-09
support metrics from multiple eval sets in lgbm

raphcoterta:
Using multiple "eval_sets" would result in many points being logged on the same graph of dvc. I extended the DVCLiveCallback class locally and changed the __call__ definition to support logging for different eval_sets in different graphs.

  guide.


daavoo (commented):
Hi @raphcoterta ! Thanks for the contribution. Could you please add a test covering the change in:

https://github.com/iterative/dvclive/blob/main/tests/test_frameworks/test_lgbm.py

PR #571, opened 2023-05-19
lightning: parse nested params

dberenbaum:
Nested params were being stringified instead of kept in nested structure.

Before this PR, they were being saved like:

```yaml
optim_params: '{"lr": 0.01}''
```

After this PR, they are being saved like:

```yaml
optim_params:
  lr: 0.01
```


src/dvclive/lightning.py
```diff
@@ -68,7 +64,6 @@ def name(self):
     def log_hyperparams(self, params, *args, **kwargs):
         params = _convert_params(params)
         params = _sanitize_callable_params(params)
-        params = _sanitize_params(params)
```
daavoo, line 71: I think this function also sanitizes some things that would break the yaml serialization , would be good to double check and see if we can keep that behavior.

Maybe flatten params before this call?

dberenbaum: There is already a [flatten](https://github.com/Lightning-AI/lightning/blob/61246c3b3578de71a0ce33a311f410269b21c1a8/src/lightning/fabric/utilities/logger.py#LL67C1-L67C1) method available, but we don't want them flat here because we expect the params to stay unflattened. Otherwise they will get saved like `optim_params.lr: 0.01`, which doesn't get parsed into multiple levels in vs code or studio.

We could copy and monkeypatch [_sanitize_params](https://github.com/Lightning-AI/lightning/blob/61246c3b3578de71a0ce33a311f410269b21c1a8/src/lightning/fabric/utilities/logger.py#L97) and do our own sanitizing.

dberenbaum: Ended up traversing through the params dict to sanitize.

daavoo (approved):

PR #576, opened 2023-05-22
fix(experiments): use custom `Random` gen in exp names

shcheklein:
It's common in ML frameworks to have seed set for reproducibility. I don't think we want this for experiment names though. It creates collisions, which is fine by itself, but in this case can be extreme.

If this approach sounds good, I'll fix it in DVC also.

https://github.com/iterative/dvclive/assets/3659196/f3304c23-3dee-49a2-a846-71df9fe927a0


daavoo (approved):

daavoo:
> If this approach sounds good, I'll fix it in DVC also.

Actually, we now have DVC as explicit dependency, we should remove this code and import the function from there

daavoo:
Closing in favor of the DVC one and will bump version and import here

PR #583, opened 2023-05-25
suppress dvc logs

dberenbaum:
Drops the default log level from "info" to "warning" for dvc logging messages.


src/dvclive/live.py
```diff
@@ -130,6 +130,10 @@ def _init_dvc(self):
                 self._save_dvc_exp = False
 
         self._dvc_repo = get_dvc_repo()
+
+        dvc_logger = logging.getLogger("dvc")
+        dvc_logger.setLevel(os.getenv(env.DVCLIVE_LOGLEVEL, "WARNING").upper())
```
shcheklein, line 135: do we still use logs to output UX/UI messages? (I thought we were migrating out of it)

any important messages (e.g. does it makes sense to still show this message once) that we might miss because of this?

(in YOLO I can for now do the log artifact once at the end)

dberenbaum: > do we still use logs to output UX/UI messages? (I thought we were migrating out of it)

At least in this case, we are still using logs.

Before:

```
>>> live.log_artifact("images")
100% Adding...|█████████████████████████████████████████████████|1/1 [00:00, 13.25file/s]

To track the changes with git, run:

        git add images.dvc

To enable auto staging, run:

        dvc config core.autostage true

```

After:

```
>>> live.log_artifact("images")
100% Adding...|█████████████████████████████████████████████████|1/1 [00:00, 12.77file/s]
```

In general, it's a good point. I would need to look deeper into when we use logs vs ui and if we need to also think about ui verbosity.

> any important messages (e.g. does it makes sense to still show this message once) that we might miss because of this?

Again, I'd have to look deeper, but this change only drops the level from info to warn. Since info shouldn't have important messages, and warn is the default for python anyway, it _theoretically_ should be fine.

My general feeling is that dvclive is too noisy overall and this change would be for the better, but I could also keep playing with it before merging. We may need to try for a bit to see if it feels like a good change.

daavoo: > any important messages (e.g. does it makes sense to still show this message once) that we might miss because of this?

If there is something really important we should be logging it from DVCLive logger

PR #584, opened 2023-05-25
log_metric: add plot flag

### LINKED ISSUE #574: `summary`: split names into nested dict automatically
```python
live.summary[metric] = value
```

should have the same semnatics as `live.log_metric(metric)` except skipping the plots part.

It means we should split the metric name by `/` (`os.sep`) as we do in `log_metric`.

Btw, why do we use `os.sep` to split metric names? Are they usually platform dependent?

dberenbaum:
Closes #574


tests/test_main.py
```diff
@@ -29,10 +29,10 @@ def read_latest(live, metric_name):
 def test_logging_no_step(tmp_dir):
     dvclive = Live("logs")
 
-    dvclive.log_metric("m1", 1)
+    dvclive.log_metric("m1", 1, plot=False)
     dvclive.make_summary()
 
-    assert not (tmp_dir / "logs" / "m1.tsv").is_file()
+    assert not (tmp_dir / "logs" / "plots" / "metrics" / "m1.tsv").is_file()
```
shcheklein, line 35: why changing the path here?

dberenbaum: The test was outdated and shouldn't have been passing. `tmp_dir / "logs" / "m1.tsv"` is never going to be generated in the current dvclive version. So I adjusted the path to make it fail properly and then added the flag to make it pass again.

shcheklein (approved):

daavoo (approved):

dberenbaum:
@shcheklein Let me know if you need a release for this (or feel free to do it yourself; I'm not aware of any blockers)

PR #587, opened 2023-05-27
xgb: infer metric data names from `evals` and deprecate `metric_data`

sisp:
I've added support for logging metrics for multiple data sets/splits (e.g. train and test/eval) with XGBoost. This is consistent with other supported frameworks, e.g. Keras. It's also a common use case to compare the metrics on multiple data sets/splits, e.g. to check overfitting.

This addition is backwards compatible.

I'll create a PR to the docs project once this addition has been approved.

  guide.


sisp:
Or should we remove the `metrics_data` argument and create a subdirectory for each evaluation set? It seems the other framework integrations work like that. It would be a breaking change though.

Alternatively, `metrics_data` could also accept `None` default to `None` in which case metrics for all evaluation sets would be logged. That would be backwards compatible if we kept omitting the subdirectory for the evaluation set when only one set is used. It seems other framework integrations always create this subdirectory though, so the XGBoost integration is the exception.

daavoo (changes requested):
Hi @sisp thanks for the contribution! 

I took a look and I think what makes sense is to just go ahead and drop `metric_data ` , as you suggested:

> Or should we remove the metric_data argument and create a subdirectory for each evaluation set? It seems the other framework integrations work like that. It would be a breaking change though.

To make it not-so-breaking for now, we can leave the argument and just ignore it + show a warning that it is being ignored.

daavoo:
> We could allow `None` (i.e. `Optional[str]` and write metrics logs to those inferred subdirectories), default to `None`, and show a deprecation warning to notify users and give them time to migrate.

Yes, sorry for the lack of clarity, I meant the option in your last paragraph

sisp:
Great! I'll update the PR tomorrow.

sisp:
@daavoo I've updated the PR as we discussed and resolved merge conflicts with `main`. I've also submitted a PR to the `dvc.org` project to first document the `metric_data` parameter and also add a notice regarding its deprecation: https://github.com/iterative/dvc.org/pull/4578

src/dvclive/xgb.py
```diff
@@ -8,18 +9,35 @@
...
+            [(self._metric_data, "")]
+            if self._metric_data
+            else zip(evals_log, evals_log)
+        ):
+            for key, values in evals_log[name].items():
+                if values:
+                    latest_metric = values[-1]
+                self.live.log_metric(
+                    f"{subdir}/{key}" if subdir else key, latest_metric
+                )
```
daavoo, line 40: Just minor aesthetics but I found this more clear than the `zip`:

```suggestion
        if self._metric_data:
            evals_log = {"": evals_log[self._metric_data]}
        for subdir, data in evals_log.items():
            for key, values in data.items():
                self.live.log_metric(
                    f"{subdir}/{key}" if subdir else key, values[-1]
                )
```

Unrelated, the `latest_metric` check was pointless

sisp: > Unrelated, the latest_metric check was pointless

Yes, I was also wondering about that but didn't want to mix it with this PR. :laughing:

daavoo (approved):
Thank you for contributing and addressing all comments!

PR #588, opened 2023-05-28
docs: add `dev` extra to install command in contributing guide

sisp:
I've updated the contributing guide to instruct a contributor to install also the `dev` extra. Otherwise, `pytest` and all other development dependencies won't be available.

  guide.

- [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. &ndash; **n/a**

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -57,11 +57,11 @@ You need Python 3.8+.
    $ python -m venv .venv
    $ source .venv/bin/activate
 
-Install in editable mode:
+Install in editable mode including development dependencies:
 
 .. code:: console
 
-   $ pip install -e .
+   $ pip install -e .[dev]
 
 
 How to test the project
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -57,11 +57,11 @@ You need Python 3.8+.
    $ python -m venv .venv
    $ source .venv/bin/activate
 
-Install in editable mode:
+Install in editable mode including development dependencies:
 
 .. code:: console
 
-   $ pip install -e .
+   $ pip install -e .[dev]
```
daavoo, line 64: Just because dev also installs all frameworks and it is probably too much in most cases, I think would make more sense to:

```suggestion
   $ pip install -e .[tests]
```

Although I should update the framework tests to be skipped when the framework is missing

daavoo (approved):

daavoo:
Thanks @sisp !

Issue #359, opened 2022-11-04
version attr missing


dberenbaum:
Need to support a version attribute to check the current version. I'd say it's important with the intro of 1.0.

skshetry:
Users can do that easily with:
```python
from importlib.metadata import version

version('dvclive')
```

dberenbaum:
Judging from https://stackoverflow.com/questions/20180543/how-do-i-check-the-versions-of-python-modules, I guess `dvclive.__version__` may be an outdated/not recommended anymore, although it's still unclear. I still default to that for finding a package version. Do you consider it bad practice? Does it hurt to include it?

alex000kim:
ahaha, I came to this repo to create this exact issue :)

skshetry:
See https://stackoverflow.com/questions/72167802/adding-version-attribute-to-python-module.

I don't see much difference between `version('dvclive')`, and that is in-built into the standard library, and is a canonical truth (even if `dvclive.__version__` may sometimes lie and is just a convention).

Do users want to support multiple versions of `dvclive`? Or, is it for constraining to a certain version? In that case, either restricting via a package manager might be a good option, or an ad-hoc `version('dvclive')`.

Again, I am not opposed to this, it's a simple addition. I was merely suggesting that there are other ways to get the _real_ installed version.

alex000kim:
The only argument I can make is the same as what @dberenbaum said:
> I still default to that for finding a package version.

daavoo:
My vote is to just recommend `importlib.metadata` when someone asks.

PR #589, opened 2023-05-29
chore(setup): migrate to pyproject, add version

shcheklein:
Migrate setup to `pyproject.toml`. Adds `__version__` to the package. Needed to do dynamic checks (e.g. check that version is correct in the callback) and overall it's expected that packages have it.

----------------

- [X] ❗ I have followed the [Contributing to DVCLive](https://github.com/iterative/dvclive/blob/master/CONTRIBUTING.md)
  guide.


pyproject.toml
```diff
@@ -1,8 +1,92 @@
...
+    "mypy>=1.1.1"
+]
+mmcv = ["mmcv"]
+tf = ["tensorflow"]
+xgb = ["xgboost"]
+lgbm = ["lightgbm"]
+huggingface = ["transformers", "datasets"]
+catalyst = ["catalyst>22"]
+fastai = ["fastai"]
+pytorch_lightning = ["pytorch-lightning>=1.9", "torch<2.1"] # for backward compatibility
```
shcheklein, line 65: @skshetry @daavoo not sure why, but this gives me an error sometimes. (`pytorch_lightning` extra is not found, it seems that underscore is converted to `-` sometimes) ... kept the previous one for the backward compatibility, but tbh didn't have much time to look into this ... if you have any ideas let me know, or we can move forward with this

daavoo (commented):
> Needed to do dynamic checks (e.g. check that version is correct in the callback) and overall it's expected that packages have it.

See https://github.com/iterative/dvclive/issues/359#issuecomment-1305052717 .

Do you have any reason not to use it in the callback?

I prefer that over adding the `version`, but I have no strong opinion.

pyproject.toml
daavoo: > but this gives me an error sometimes. (pytorch_lightning extra is not found, it seems that underscore is converted to - sometimes)

When did you hit the error?

It appears that they recommend `lightning` as the name now in the docs:

https://lightning.ai/docs/pytorch/stable/starter/installation.html

shcheklein: I was getting `WARNING: dvclive 2.10.2.dev11+g937c870.d20230529 does not provide the extra 'pytorch_lightning'` in tests and locally when trying to `pip install ".[all]" or similar. I think this is specifically related to the target name `pytorch_lightning`.

shcheklein:
> I prefer that over adding the version, but I have no strong opinion.

Also, don't have a strong opinion. I think it can be convenient since people seems to expect it still and most (?) of the package provide it. But I can remove it (I assume toml migration is still meaningful for us)

pyproject.toml
shcheklein: > It appears that they recommend lightning as the name now in the docs:

Let's rename and keep the old one for backward compatibility?

daavoo (approved):

.gitignore
```diff
@@ -1,3 +1,5 @@
+/src/dvclive/_dvclive_version.py
+
```
daavoo, line 2:
```suggestion
```

src/dvclive/__init__.py
```diff
@@ -1 +1,3 @@
+from dvclive.version import __version__, version_tuple  # noqa: F401
+
```
daavoo, line 2:
```suggestion
```

pyproject.toml
```diff
@@ -1,8 +1,92 @@
...
+pytorch-lightning = ["pytorch-lightning>=1.9", "torch<2.1"]
```
daavoo, line 66:
```suggestion
lightning = ["lightning>=1.9", "torch<2.1"]
```

PR #593, opened 2023-06-05
allow dvc>=3.0

dberenbaum:
I don't think there are any breaking changes planned here, so can we start enabling dvclive to work with dvc>=3.0 now?


dberenbaum:
@daavoo PTAL and let's release so we don't have to worry about timing issues with the dvc release 🙏

Issue #586, opened 2023-05-25
`log_model`


dberenbaum:
We already have two closed issues about this 😓 , but I think we should add a `log_model` method that uses mlem to inspect the model artifact.

Benefits:
* Automatically captures metadata about the model artifact that users can see in the model registry to make it more meaningful.
* Makes model registry workflow more useful since users can deploy downstream with mlem.
* Automatically logs it as an artifact with `type=model`.
* Can support different options for how to save the model (related to #572).

dberenbaum:
cc @aguschin

aguschin:
Ok, now I'm getting what we reserved `log_model` for 😅 

I assume `log_model` should simply call `mlem.api.save` and ideally have the same (in practice, almost the same) args https://mlem.ai/doc/api-reference/save

This should be an easy change. Happy to take it once we're done with MR.

dberenbaum:
In addition, it would be nice to also support an already serialized model with https://mlem.ai/doc/api-reference/import_object. In other words, it would be nice to accept either a model object or a path to a model.

Issue #572, opened 2023-05-20
Decouple `dvc add` from `log_artifact`


shcheklein:
There are a bunch of issues in the current implementation, most of them are related to the workflow when one has to run `log_artifact` on each iteration (e.g. saving best model via callback).

- It prints this stuff on each iteration (super noisy):

```
To track the changes with git, run:
        git add dvclive/artifacts/best.pt.dvc
To enable auto staging, run:
        dvc config core.autostage true
```

- It saves a model into cache (takes time + space)
- It requires settings up a DVC remote. Which is by all means optional- there are other ways to handle storage

Related to #551

daavoo:
> workflow when one has to run log_artifact on each iteration (e.g. saving best model via callback)

I think we have discussed this elsewhere but we went with the approach of not doing this in the integrations and instead documenting/explaining to the users how to track the model once the training loop is completed, like in:

https://dvc.org/doc/start/experiments/experiment-tracking?tab=Pytorch-Lightning

This was under the assumption of using `log_artifact` for the model registry scenario.

Are we discussing here a different scenario to use the checkpoint as recovery from interruption?

shcheklein:
Scenarious include updating the best model from time to time (e.g to be able early stop or just in general have the best weights at the end) + recovery. Yes, it can be done at the end also with a copy flag, but it's unrealistic to expect users or callbacks to behave that way.

We can porentially run `dvc add` only at the end at least. That would solve at least some of the issues above. And even if it's interrupted at least model is there.

dberenbaum:
> when one has to run `log_artifact` on each iteration (e.g. saving best model via callback)

Why does `log_artifact` have to get called repeatedly here? And what is the point of doing so if you only want to `dvc add` at the end?

I think doing `dvc add` only at the end makes some sense if we aren't currently focused on checkpointing, but I'm confused about how that connects to calling `log_artifact` multiple times or compatibility with recovery from interruption.

shcheklein:
I think what I'm trying to cover here (and we can discuss the details about the YOLO) is the workflow when a framework has a callback - `on_model_save` or similar and we just add the `log_artifact` there without thinking too much. Scenarios are the same to why people in general save the best model periodically and the last one I think. I feel that it quite limiting to expect users to do the `log_artifact` only at the end + explaining then context, etc.

[...]

Issue #551, opened 2023-05-01
`log_artifact`: external and non-DVC tracked files support


shcheklein:
More of a question for now:

- Should DVC tracking be optional / disabled by default? E.g. if model weights are in Git-lfs? I can still see the model in the registry, but I don't need DVC remote, etc, etc.
- How should we treat S3 files? `dvc.yaml` should support them I think. Do we need to create import file `.dvc` or not?

daavoo:
Feels like the questions are assuming `log_artifact` is coupled with the model registry but strictly speaking, it is not.

As of today, it is decoupled from an implementation perspective (`log_artifact` creates the `.dvc` but it is `make_dvcyaml` that writes the `artifacts` section), but I would also like to think that it should not be coupled from a product perspective.

For those scenarios, why would you want to use `log_artifact` python API for registering the model? It is more convenient than writing the `artifacts` section in the `dvc.yaml` or using the UI?

If we still want a Python API, should we make it part of `dvc.api`? Does it belong in DVCLive *logger*?

dberenbaum:
> * Should DVC tracking be optional / disabled by default? E.g. if model weights are in Git-lfs? I can still see the model in the registry, but I don't need DVC remote, etc, etc.

The use case I can think of is huggingfaces integration. Is that what you have in mind?

Would we also make `dvc get` work with git-lfs? Do you have a use case where model registry is useful without being able to retrieve the artifact?

> * How should we treat S3 files? `dvc.yaml` should support them I think. Do we need to create import file `.dvc` or not?

@daavoo If we import the model, isn't that part of the core functionality of `log_artifact`? I think this is an interesting idea because it helps introduce a way to manage external data, which is a major source of confusion today.

shcheklein:
> Would we also make dvc get work with git-lfs? 

I don't know yet, at the end people might decide on their own also how exactly bring the artifact from a commit.

dberenbaum:
> I don't know yet, at the end people might decide on their own also how exactly bring the artifact from a commit.

Can we come up with a use case where model registry is needed in this scenario?

shcheklein:
> Can we come up with a use case where model registry is needed in this scenario?

To be honest, I don't see the difference is it DVC-tracked or not. All the same scenarios apply, no? Find a specific version of a model (by a tag) and fetch it to deploy. Assign stages, etc, etc. In this case `dvc.yaml` helps to see them in the MR + to see some additional metadata.

[...]

Issue #504, opened 2023-03-20
frameworks: improve log_artifact for model "checkpoints"


dberenbaum:
When auto-logging model artifacts as part of the frameworks in #500, sometimes frameworks save multiple model files at different paths within a directory to allow users to keep multiple model checkpoints:
* https://lightning.ai/docs/pytorch/stable/common/checkpointing_basic.html
* https://huggingface.co/docs/transformers/main_classes/trainer#transformers.Trainer.save_model
* https://docs.fast.ai/callback.tracker.html#savemodelcallback

This present some questions for our callbacks:
1. Should we save the individual files or the entire directory?
2. How should we keep track of the latest model file so that it can be used to resume/analyze/deploy?

dberenbaum:
After #472, `live.log_artifact` should be adding something to `dvclive/dvc.yaml` like:

```yaml
artifacts:
  fast-ai-model:
    path: model.pth
    type: model
```

This `artifacts` section could be used to answer question 2 above. Each call to `live.log_artifact` can update the path in that `artifacts` section from `model_1.pth` to `model_2.pth` and so on, and @aguschin is working on making it possible to do things like `dvc get . fast-ai-model` so that it's possible to get the model without specifying the exact path of the last checkpoint.

In that case, question 1 becomes less important and is mostly about how we want to organize the `.dvc` files: is it okay to have one for each checkpoint, or should we try to track the model directory?

dberenbaum:
I think we can close this as completed with the work we have done so far in lightning and huggingface

PR #613, opened 2023-07-03
auto save model in lightning

dberenbaum:
This PR auto logs models with `dvclive.lightning.DVCLiveLogger(log_model=True)`:

```
trainer = pl.Trainer(logger=[DVCLiveLogger(log_model=True)])
```

`log_model` follows the conventions in [mlflow](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.mlflow.html#module-lightning.pytorch.loggers.mlflow) and [wandb](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.wandb.html):
* `False` saves no models (this is the default).
* `True` saves all model checkpoints at the end of training. 
* `"all"` saves all model checkpoints whenever a model checkpoint is saved.

If `log_model` is `True` or `"all"`, dvclive caches the entire checkpoints folder. 

Dvclive will also add a model artifact named "best" at the end of training that references the best model checkpoint. (edit: this resembles the [best alias](https://docs.wandb.ai/guides/integrations/lightning#model-checkpointing) in wandb)

Edit: example `dvclive/dvc.yaml` output:

```yaml
artifacts:
  best:
    path: ../DvcLiveLogger/dvclive_run/checkpoints/epoch=1-step=4-v1.ckpt
    type: model
```

To support this, `log_artifact` was also changed:
* Artifacts will only be added to `dvc.yaml:artifacts` if some metadata is provided (type, name, desc, labels, or meta). This is a breaking change, but I can't see how anyone is making use of this without any metadata since it won't be used by the model registry.
* Added `cache` kwarg to `log_artifact` (defaults to `True`) so that it's possible to add the artifact metadata without caching the object.

Related:
* #504 
* #551 
* #572 
* #586 

To do:
- [x] Tests
- [ ] Docs (https://github.com/iterative/dvc.org/pull/4714)
- [ ] VS Code snippets (cc @mattseddon)
- [ ] Studio snippets (cc @yathomasi)


dberenbaum:
Overall, I felt opening a PR with the desired behavior would be more effective than explaining and discussing in an issue. I hope this will help resolve some of the rough edges around saving models and that we can work through the other framework callbacks to implement similar functionality that works with the existing framework conventions and resembles mlflow, wandb, etc.

dberenbaum:
One thing to note: lightning will not overwrite existing files or clean up between runs. Instead, it will append a version number, so if you run the same code repeatedly, you will end up with a directory that tracks your entire history of model checkpoints instead of only the latest run:

```
$ tree DvcLiveLogger
DvcLiveLogger
└── dvclive_run
    ├── checkpoints
    │   ├── epoch=0-step=2-v1.ckpt
    │   ├── epoch=0-step=2-v2.ckpt
    │   ├── epoch=0-step=2.ckpt
    │   ├── epoch=1-step=32-v1.ckpt
    │   ├── epoch=1-step=32-v2.ckpt
    │   ├── epoch=1-step=32.ckpt
    │   ├── epoch=1-step=4.ckpt
    │   ├── epoch=2-step=6.ckpt
    │   ├── epoch=4-step=10-v1.ckpt
    │   ├── epoch=4-step=10-v2.ckpt
    │   ├── epoch=4-step=10-v3.ckpt
    │   ├── epoch=4-step=10-v4.ckpt
    │   ├── epoch=4-step=10-v5.ckpt
    │   ├── epoch=4-step=10-v6.ckpt
    │   └── epoch=4-step=10.ckpt
    └── checkpoints.dvc
```

If you are running a pipeline, this is probably fine since you can control if you want to delete that directory each time. We might also want to consider dropping the existing checkpoints directory in the dvclive callback if `resume=False`.

daavoo:
> log_model follows the conventions in [mlflow](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.mlflow.html#module-lightning.pytorch.loggers.mlflow) and [wandb](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.wandb.html):

`True` for MLFLow appears to be different:

```
log_model follows the conventions in [mlflow](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.mlflow.html#module-lightning.pytorch.loggers.mlflow) and [wandb](https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.loggers.wandb.html):
```

dberenbaum:
@daavoo I can add the behavior for `save_top_k == -1` to log every checkpoint. I figured it's a minor detail and a quick google search didn't turn up any justification for why they do it, but wandb does it also, so we can do it for consistency.

dberenbaum:
I think this gets us to a good place with logging models in lightning. In fact, comparing to other trackers, it feels a bit easier to manage the models this way in dvc. On a different machine, you can pull the lightning checkpoints dir and keep using lightning methods to load those checkpoints. With other trackers, once you are on a different machine, the only way to load models is using the experiment tracker's api.

src/dvclive/lightning.py
```diff
@@ -119,6 +123,23 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
...
+        ):
+            self.experiment.log_artifact(checkpoint_callback.dirpath)
+
     @rank_zero_only
     def finalize(self, status: str) -> None:
+        checkpoint_callback = self._checkpoint_callback
+        # Save model checkpoints.
+        if self._log_model is True:
+            self.experiment.log_artifact(checkpoint_callback.dirpath)
+        # Log best model.
```
daavoo, line 139: WDYT of creating a copy in "dvclive" folder (or in the checkpoints folder itself), at least for the best?

It seems that we would be changing the path of the registered model between experiments in the current behavior

daavoo:
> If you are running a pipeline, this is probably fine since you can control if you want to delete that directory each time. We might also want to consider dropping the existing checkpoints directory in the dvclive callback if resume=False.

Didn't look in details, but seems like the other loggers use [_scan_checkpoints ](https://github.com/Lightning-AI/lightning/blob/017262e5e0c65215e9e75121d155d7a07cd9e7bf/src/lightning/pytorch/loggers/utilities.py#L32) to only track the ones related to the current experiment

src/dvclive/lightning.py
dberenbaum: Yes, which I guess is also what other trackers do AFAICT? Do you think the path matters? Maybe it makes it easier to `dvc get` later, although we could make that work by the artifact name. No strong opinion from me.

dberenbaum:
> Didn't look in details, but seems like the other loggers use [_scan_checkpoints ](https://github.com/Lightning-AI/lightning/blob/017262e5e0c65215e9e75121d155d7a07cd9e7bf/src/lightning/pytorch/loggers/utilities.py#L32) to only track the ones related to the current experiment

Great idea. I'll look into it.

dberenbaum:
Lightning [warns](https://github.com/Lightning-AI/lightning/blob/017262e5e0c65215e9e75121d155d7a07cd9e7bf/src/lightning/pytorch/callbacks/model_checkpoint.py#L620) if the directory is not empty:

```console
UserWarning: Checkpoint directory /Users/dave/Code/lstm_seq2seq/model exists and is not empty.
  rank_zero_warn(f"Checkpoint directory {dirpath} exists and is not empty.")
```

`_scan_checkpoints` only returns the latest checkpoint even if you use something like `ModelCheckpoint(save_top_k=-1)`, so I had to save the checkpoint from each scan and then drop the rest of the files in the directory (unless `resume=True`).

Removing the files only happens after the checkpoint is saved, so sometimes the first checkpoint will still get a version number like:
```console
$ tree model
model
├── epoch=0-step=2-v1.ckpt # saved this checkpoint before previous one was dropped
├── epoch=1-step=4.ckpt
├── epoch=2-step=6.ckpt
├── epoch=3-step=8.ckpt
└── epoch=4-step=10.ckpt
```

Overall, this works and probably meets most user's expectations, so I think we should keep it, but I don't feel strongly that it outweighs the added complexity or potential surprise that dvclive is deleting model checkpoints.

dberenbaum:
ping @daavoo

src/dvclive/lightning.py
```diff
@@ -131,6 +141,42 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
...
+        if self._log_model == "all" or (
+            self._log_model is True and checkpoint_callback.save_top_k == -1
+        ):
+            self._save_checkpoints(checkpoint_callback)
+
     @rank_zero_only
     def finalize(self, status: str) -> None:
+        # Log best model.
+        if self._checkpoint_callback:
+            self._scan_checkpoints(self._checkpoint_callback)
```
daavoo, line 157:
```suggestion
```

daavoo (approved):
On a high level, the code and description make sense to me.

I have not actually tried in a project the different options, but the test looks reasonable so trusting that.

src/dvclive/lightning.py
```diff
@@ -131,6 +141,42 @@ def log_metrics(self, metrics: Dict[str, Any], step: Optional[int] = None):
...
+        for t, p, _, _ in checkpoints:
+            self._logged_model_time[p] = t
+            self._all_checkpoint_paths.append(p)
+
+    def _save_checkpoints(self, checkpoint_callback: ModelCheckpoint) -> None:
+        # drop unused checkpoints
+        if not self._experiment._resume:  # noqa: SLF001
+            for p in Path(checkpoint_callback.dirpath).iterdir():
+                if str(p) not in self._all_checkpoint_paths:
+                    p.unlink(missing_ok=True)
```
daavoo, line 179: Let's be clear about this in the docs

daavoo:
I think I would like to have a `best` (or `best_only`) option for `log_model` but we can discuss separately

PR #614, opened 2023-07-03
add cache_images to lightning

dberenbaum:
Adds `dvclive.lightning.DVCLiveLogger(cache_images=True)`

### DIFF
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -53,6 +53,7 @@ class DVCLiveLogger(Logger):
             "report": report,
             "save_dvc_exp": save_dvc_exp,
             "dvcyaml": dvcyaml,
+            "cache_images": cache_images,
         }
         if dir is not None:
             self._live_init["dir"] = dir
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -44,6 +44,7 @@ class DVCLiveLogger(Logger):
         report: Optional[str] = "auto",
         save_dvc_exp: bool = False,
         dvcyaml: bool = True,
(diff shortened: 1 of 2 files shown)


src/dvclive/lightning.py
```diff
@@ -44,6 +44,7 @@ def __init__(
         report: Optional[str] = "auto",
         save_dvc_exp: bool = False,
         dvcyaml: bool = True,
+        cache_images: bool = False,
```
daavoo, line 47: You need to also add it to the `_live_init` dict below.

daavoo (changes requested):

dberenbaum:
@daavoo Any ideas for how to test when adding a new `Live()` option that it also exists in the lightning logger?

daavoo:
> @daavoo Any ideas for how to test when adding a new `Live()` option that it also exists in the lightning logger?

I guess we need to remember to update this test with the new args: https://github.com/iterative/dvclive/blob/a0de508ace2df86413b3d9952abaa652cbbdad5d/tests/test_frameworks/test_lightning.py#L144-L158

Seems like the reasonable thing to do since we don't change or have that many args and it is the only framework where the args are not directly passed like in other frameworks as `Live(**kwargs)`

dberenbaum:
> Seems like the reasonable thing to do since we don't change or have that many args and it is the only framework where the args are not directly passed like in other frameworks as `Live(**kwargs)`

Added a test, but I think not changing it often is why I forget to do it. Anyways, I don't have any better ideas for now. Just a thought.

daavoo (approved):

PR #615, opened 2023-07-03
cleanup dvcyaml

### LINKED ISSUE #611: params_show compatibility
If `dvclive.dvc.yaml` includes a params section but that params file does not yet exist, `dvc.api.params_show()` will fail. This has been reported by a user in https://iterativeai.slack.com/archives/C03JS2V4MQU/p1687449694765559, and I have reproduced it, although I need to narrow down when/why this occurs since I thought the dvc.yaml shouldn't be generated until the end, when the params file should already exist.

dberenbaum:
Fixes #611

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -121,6 +121,9 @@ class Live:
             if f and os.path.exists(f):
                 os.remove(f)
 
+        if self.dvc_file and os.path.exists(self.dvc_file):
+            os.remove(self.dvc_file)
+
     def _init_dvc(self):
         from dvc.scm import NoSCM
 
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -171,6 +171,7 @@ def test_cleanup(tmp_dir, html):
 
     assert (tmp_dir / dvclive.plots_dir / Metric.subfolder / "m1.tsv").is_file()
     assert (tmp_dir / dvclive.metrics_file).is_file()
+    assert (tmp_dir / dvclive.dvc_file).is_file()
     assert html_path.is_file() == html
 
     dvclive = Live("logs")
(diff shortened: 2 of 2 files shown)


src/dvclive/live.py
```diff
@@ -121,6 +121,9 @@ def _init_cleanup(self):
             if f and os.path.exists(f):
                 os.remove(f)
 
+        if self.dvc_file and os.path.exists(self.dvc_file):
```
daavoo, line 124: I remember we had a discussion the first time we added the make_dvcyaml about not wanting to remove it on cleanup https://github.com/iterative/dvclive/pull/366#discussion_r1036232488

Do you think the concerns no longer apply (the config for sklearn is now auto-generated) and we can consider the dvclive/dvc.yaml not modifiable?

daavoo: It looks like we are overriding anyhow in the code, so we never preserved the existing dvcyaml 😅

daavoo (approved):
The code changes look good to me, and it is consistent with the behavior we already applying *https://github.com/iterative/dvclive/pull/615#discussion_r1251854717(.

My only worry is that I am not understanding the issue behind https://github.com/iterative/dvclive/issues/611 and when this is a fix. I assume that this is a fix only if `Live` gets instantiated before `params_show` is called?

dberenbaum:
@daavoo Where are we overwriting in the code? Is this PR doing anything then? 😅

daavoo:
> @daavoo Where are we overwriting in the code? 

https://github.com/iterative/dvclive/blob/d261760a2fc32c36faa3b074a6a14a69e65cee5d/src/dvclive/dvc.py#L86-L116

This only happens when `make_dvcyaml` is called.

> Is this PR doing anything then? 😅

Yes, it makes the behavior consistent with the rest of the DVCLive-generated files, where we clean up at  `Live` instantiation

dberenbaum:
I wonder if we need to be so particular anymore about which files to drop inside the dvclive dir or if we should just drop the whole dir. Seems just as likely that someone will end up with stale files than deleting something someone wanted to keep. AFAIR we used to drop more aggressively before https://github.com/iterative/dvclive/issues/102, but maybe we are more clear now that the dvclive dir should be a separate dir managed only by dvclive.

PR #619, opened 2023-07-05
Cast path to string

### LINKED ISSUE #618: cache_images issues
I'm trying to use the new `cache_images` feature from the ultralytics repo, but I get this warning (though it feels more like an error):
```
WARNING:dvclive:Failed to dvc add dvclive/plots/images: 'PosixPath' object is not iterable
```
This causes the feature to fall apart, and no files are added to dvc.
To fix it, I've hacked a string-cast into the `live.py:cache` function. This solves the issue, but I'm not sure if that's the right way to do this.

CptCaptain:
In some cases this might be a PosixPath, causing issues downstream.

Fixes #618

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -446,7 +446,7 @@ class Live:
 
     def cache(self, path):
         try:
-            stage = self._dvc_repo.add(path)
+            stage = self._dvc_repo.add(str(path))
         except Exception as e:  # noqa: BLE001
             logger.warning(f"Failed to dvc add {path}: {e}")
             return


daavoo:
I am not sure in which scenarios this is failing, though.
Tried to add a test but `self._dvc_repo.add(path)` doesn't fail.

@CptCaptain Could you share your DVC version?

CptCaptain:
Sure, I'm using these versions:
```
dvc==2.58.2
dvc-data==0.51.0
dvc-gs==2.22.1
dvc-http==2.30.2
dvc-objects==0.23.0
dvc-render==0.5.3
dvc-studio-client==0.11.0
dvc-task==0.3.0
dvclive==2.12.0
```
This all runs from the ultralytics repo, if that makes a difference.

daavoo (approved):
Ok, appears to be some issue that has been resolved in more recent DVC versions (in the latest version the error doesn't appear). 

I am able to reproduce this when installing those pinned dependencies.

skshetry:
This was fixed in 3.0.

PR #620, opened 2023-07-05
log_artifact: add cache option, only write to dvc.yaml if metadata ex…

### LINKED ISSUE #551: `log_artifact`: external and non-DVC tracked files support

### LINKED ISSUE #572: Decouple `dvc add` from `log_artifact`

dberenbaum:
Extracted from #613.

* Artifacts will only be annotated in `dvc.yaml:artifacts` if some metadata is provided (type, name, desc, labels, or meta). This is a breaking change, but I can't see how anyone is making use of this without any metadata since it won't be used by the model registry. It's needed in #613 so we can cache a whole directory of checkpoints without annotating the whole directory.
* Added `cache` kwarg to `log_artifact` (defaults to `True`) so that it's possible to annotate the artifact metadata without caching the object. It's needed in #613 so we can annotate the `best` artifact in `dvc.yaml` without re-caching (since it overlaps with the cached checkpoints directory).

There are probably workarounds for both in #613, but I think this flexibility to choose when to cache and when to annotate in `dvc.yaml` is useful. It also should close #572 ~~and #551~~ (edit: and partially address #551) since it decouples caching and annotation.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -413,10 +413,11 @@ class Live:
         path: StrPath,
         type: Optional[str] = None,  # noqa: A002
         name: Optional[str] = None,
-        desc: Optional[str] = None,  # noqa: ARG002
-        labels: Optional[List[str]] = None,  # noqa: ARG002
-        meta: Optional[Dict[str, Any]] = None,  # noqa: ARG002
+        desc: Optional[str] = None,
+        labels: Optional[List[str]] = None,
+        meta: Optional[Dict[str, Any]] = None,
         copy: bool = False,
+        cache: bool = True,
     ):
         """Tracks a local file or directory with DVC"""
         if not isinstance(path, (str, Path)):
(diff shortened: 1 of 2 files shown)


dberenbaum:
Looks like there are unrelated errors coming from the keras callback.

daavoo (approved):
Agree it closes https://github.com/iterative/dvclive/issues/572
But I don't think it closes the external part of https://github.com/iterative/dvclive/issues/551

PR #631, opened 2023-07-19
clean up warnings and move some to info

### LINKED ISSUE #630: Review all warnings
Ensure all warnings are relevant and don't cause confusion. Related to https://github.com/iterative/vscode-dvc/issues/3574.

- [ ]  https://github.com/iterative/dvclive/issues/609

dberenbaum:
Closes #630 

Summary of changes:
* Sets default log level to WARNING (@daavoo any reason we set it to INFO?)
* Downgrades warning about ignoring `save_dvc_exp=True` during `dvc exp run` to info. There's nothing wrong here nor any expected action from the user, so I don't think we need a warning.
* Stops trying to cache artifacts during `dvc exp run`. Instead, it will show a warning to add it as a pipeline output if isn't tracked by dvc or will state (only on info) that it is already tracked.


src/dvclive/live.py
```diff
@@ -449,6 +446,19 @@ def log_artifact(
                     )
 
     def cache(self, path):
+        if self._inside_dvc_exp:
```
daavoo, line 449: I would prefer to include and review this in a separate P.R.

daavoo (commented):
> Sets default log level to WARNING (@daavoo any reason we set it to INFO?)

I just don't understand what default is supposed to be and what level should be used for each thing. I have read and heard as many opinions as individuals

src/dvclive/live.py
```diff
@@ -449,6 +446,19 @@ def log_artifact(
...
+            from dvc.exceptions import OutputNotFoundError
+
+            msg = f"Skipping dvc add {path} because `dvc exp run` is running."
+            try:
+                self._dvc_repo.find_outs_by_path(path)
+                msg += " It is already being tracked automatically."
+                logger.info(msg)
+            except OutputNotFoundError:
```
daavoo, line 457: Let's add a test for this and for the logging

src/dvclive/live.py
dberenbaum: I moved the other parts to #632, so PTAL and let's keep this PR open to discuss the `_inside_dvc_exp` logic.

src/dvclive/live.py
dberenbaum: To test the logging statements? Do you think it's needed, and do you have an idea for how to do it?

daavoo: > To test the logging statements? 

To test the try/except logic via the logging statements.

> Do you think it's needed, and do you have an idea of how to do it?

https://github.com/iterative/dvclive/blob/b2a66c29286a8afc3eedd8082dbe5ed4f0c39bbd/tests/test_main.py#L326

daavoo: It is not strictly needed, but nice to have it tested using the non-mocked `dvc_repo` fixture

dberenbaum: Before I bother 😄 , let's agree on the behavior

daavoo: I think the behavior is ok

dberenbaum: Tweaked the messages and added tests.

If it is a stage output, the message is:
```
INFO: Skipping dvc add {path} because `dvc exp run` is running.
It is already being tracked automatically.
```

If it is another kind of output:
```
WARNING: Skipping dvc add data because `dvc exp run` is running.
To track it automatically during `dvc exp run`:
1. Run `dvc exp remove data.dvc` to stop tracking it outside the pipeline.
3. Add it as an output of the pipeline stage.
```

If it is not an output:
```
WARNING: Skipping dvc add data because `dvc exp run` is running.
To track it automatically during `dvc exp run`, add it as an output of the pipeline stage.
```

daavoo: @dberenbaum I am missing those changes in the P.R.

dberenbaum: Well, I nuked my local repo recently for some reason I can't remember, so looks like I need to redo these changes 🤦

src/dvclive/live.py
```diff
@@ -446,6 +446,33 @@ def log_artifact(
                     )
 
     def cache(self, path):
+        if self._inside_dvc_exp:
+            msg = f"Skipping dvc add {path} because `dvc exp run` is running."
+            path_stage = None
+            for stage in self._dvc_repo.stage.collect():
```
daavoo, line 452:
```suggestion
            for stage in self._dvc_repo.index.stages:
```

daavoo (approved):

src/dvclive/live.py
```diff
@@ -446,6 +446,33 @@ def log_artifact(
                     )
 
     def cache(self, path):
+        if self._inside_dvc_exp:
+            msg = f"Skipping dvc add {path} because `dvc exp run` is running."
+            path_stage = None
+            for stage in self._dvc_repo.index.stages:
```
daavoo, line 452: Leaving as a follow-up: We should probably wrap all operations around `dvc_repo` to not fail when a DVC exception is raised.

src/dvclive/live.py
```diff
@@ -446,6 +446,33 @@ def log_artifact(
...
+                for out in stage.outs:
+                    if out.fspath == str(Path(path).absolute()):
+                        path_stage = stage
+                        break
+            if not path_stage:
```
daavoo, line 457: Maybe it was just my bias from knowing the previous implementation. Still, I ended up with the model tracked in Git because we are skipping the `dvc add` entirely, which was previously adding it to gitignore.
I think we might warning but not skip the `dvc add`, as if we reach this point it means that the user is not tracking it in the `dvc.yaml` and the is no `.dvc` file

dberenbaum: Yeah, this was probably a bit overly aggressive. I'll submit a PR to keep trying `dvc add` unless it's already tracked in the pipeline.

PR #632, opened 2023-07-20
change save_dvc_exp warning to info

dberenbaum:
From https://github.com/iterative/dvclive/pull/631/files#r1269153204


daavoo (approved):

src/dvclive/live.py
```diff
@@ -133,9 +132,7 @@ def _init_dvc(self):
             self._exp_name = os.getenv(env.DVC_EXP_NAME, "")
             self._inside_dvc_exp = True
             if self._save_dvc_exp:
-                logger.warning(
-                    "Ignoring `_save_dvc_exp` because `dvc exp run` is running"
-                )
+                logger.info("Ignoring `_save_dvc_exp` because `dvc exp run` is running")
```
shcheklein, line 135: QQ, is it how actual `_save_dvc_exp` is called for users?

shcheklein: I mean for users? Or is it an internal name?

dberenbaum: Good catch. We should not be using the internal name here.

PR #635, opened 2023-07-23
README: A more engaging code snippet ecample

omesser:
  guide.

As discussed in an issues in [studio#6422](https://github.com/iterative/studio/issues/6422)) (private repo) - suggested some changes to this snippet:
copy-pasting motivation:

> Yes - a working snippet is what I was driving it.
> Looking at the snippet in the readme I suggest to consider updating maybe - that's why I looked at some competitors and compiled a new one before.
> 
> first order problem IMO is having a snippet that works out of the box
> 2nd order problem - polish how this snippet looks  - so it's short and clean and nice, but also interesting.
> things that bother me today a bit about the snippet from the [readme](https://github.com/iterative/dvclive/tree/main#example-code)(⚠️ : opinionated / style!):
> - direct access to `sys.argv` (ugly)
> - I can't run it "cleanly" -`python train.py` - fails in an ugly way (💭 mental mode - I'm just copy-pasting in "autopilot mental mode" to test studio - I don't want to have to make a decision about epoch parameter)
> - 4 metrics/plots - train/eval is redundant (the same, also "eval called "val" ? 😖 )
> - graphs aren't "interesting" to my taste (linear + noise):
>   <details><summary>screenshot</summary>
>     Screenshot 2023-07-09 at 23 47 40
>   </details>
>   that's a personal "peeve"  - it's not bad per-se, just that I saw an easy way to improve (from wandb, see below).
> - exp finishes too quickly - had to run several times to "catch" the "liveness" of the metrics
> 
> 

Graphs for new snippet
<details><summary>screenshot</summary>
  
252168487-e814ec47-8123-4f35-962a-b365f6c716f2

</details>

wdyt about this instead of the existing one @daavoo?

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,23 +52,34 @@ $ git commit -m "DVC init"
 
 ## Example code
 
-Copy the snippet below as a basic example of the API usage:
+Copy the snippet below into `train.py` for a basic API usage example:
 
 ```python
-# train.py
+import time
 import random
-import sys
+
 from dvclive import Live
 
+params = {"learning_rate": 0.002, "optimizer": "Adam", "epochs": 20}
+
 with Live(save_dvc_exp=True) as live:
-    epochs = int(sys.argv[1])
-    live.log_param("epochs", epochs)
-    for epoch in range(epochs):
-        live.log_metric("train/accuracy", epoch + random.random())
-        live.log_metric("train/loss", epochs - epoch - random.random())
-        live.log_metric("val/accuracy",epoch + random.random() )
-        live.log_metric("val/loss", epochs - epoch - random.random())
+
+    # log a parameters
+    for param in params:
+        live.log_param(param, params[param])
+
+    # simulate training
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,36 +52,48 @@ $ git commit -m "DVC init"
...
 
 ```python
-# train.py
+import time
 import random
-import sys
+
 from dvclive import Live
 
+params = {"learning_rate": 0.002, "optimizer": "Adam", "epochs": 20}
```
daavoo, line 63: I don't love that params don't have an effect, but not a blocker

daavoo (approved):

README.md
omesser: Yes, this is a for a fake feel of familiarity / realism. But I've seen this in other places and I think since this is very very explicitly "fake" it's not misleading anyone (personal feel)

PR #645, opened 2023-07-31
ignore version file

dberenbaum:
Not sure what's generating these files, but I'm finding I frequently have to ignore them


daavoo (approved):
It is probably generated when you do a pip install of the local repo

Issue #638, opened 2023-07-25
remote/no-git logging


dberenbaum:
There are several scenarios where it's not realistic to work in a git repo:

* Databricks
* Google colab
* SageMaker training jobs

People may run notebooks or jobs in these environments that are associated with a Git repo, but access to the git repo is awkward when running that notebook or job. In this case, dvclive+studio can still provide a typical experiment tracking experience (logging metrics, plots, etc. to the studio server). If the studio token and project are available, dvclive should still be able to log to studio to see live updates and associate them with an experiment.

It can be out of scope for initial implementation, but it would also be useful to persist the metrics and plots logged in this scenario back to git. If they are not tracked in git, they will be lost when that experiment is merged, reproduced, etc. It might be possible to reverse engineer the metrics and plots from studio and add them back to the git repo.

shcheklein:
Does it mean that on those platforms people don't use Git usually?

dberenbaum:
I can't say with certainty, but from my experience and limited knowledge, it depends. They may not use it at all, or they may use it in a limited way. As a generalization, I think people sometimes track the notebook or script that gets run, but the outputs are not something that they track, and with other experiment trackers, there's not really much reason to do so.

dberenbaum:
> Okay, I see. What is the plan of connecting it back? Those experiments will stay "studio"-only, is it correct?

> And why can't we try to do a git clone? any specific obstacles for that?

_Originally posted by @shcheklein in https://github.com/iterative/dvclive/issues/646#issuecomment-1666235526_
            
@shcheklein Answering here so we don't have the discussion all in the PR, and since I don't have all the answers yet, there is likely to be more discussion.

Let me break it down by different possible use cases since they are all unique (and I'm sure there are plenty more):
* **Databricks**: it's feasible to clone and commit using the Databricks UI, but DVC doesn't have access to the repo, so users have to do this manually. DVCLive sees it as a no-git scenario, so logging to Studio can only work in no-git mode. It's possible for the user to commit the DVCLive output using the Databricks UI and push it, but there's not currently any way to connect that commit to the live experiment sent to Studio.
[...]

PR #646, opened 2023-07-31
Allow for logging to Studio when not inside a repo

dberenbaum:
Partially addresses #638. This does not solve how to copy metrics back to the repo or make the experience completely smooth everywhere, but the goal for this PR is to achieve parity with other trackers where a token and project are all that's needed.

This allows for posting to Studio if the following environment variables are set:

* `DVC_STUDIO_TOKEN`
* `DVC_STUDIO_REPO_URL`
* ~`DVC_EXP_BASELINE_REV`~
* ~`DVC_EXP_NAME`~

TODO:
- [x] Make `DVC_EXP_BASELINE_REV` and/or `DVC_EXP_NAME` optional
- [x] Example
  - [Colab](https://colab.research.google.com/drive/1nkiTAp-bBJ8MZPjLKBU6ywMq74guTRTp?usp=sharing)
  - [Databricks](https://dbc-fc73949b-6595.cloud.databricks.com/?o=8761421539283779#notebook/906466764262798)
  - [Studio](https://studio.iterative.ai/user/dberenbaum/projects/dvclive-exp-tracking-p47yvfdv2w)
  - [SageMaker](https://github.com/dberenbaum/amazon-sagemaker-examples/blob/main/sagemaker-python-sdk/pytorch_mnist/pytorch_mnist.ipynb) and [Studio](https://studio.iterative.ai/user/dberenbaum/projects/amazon-sagemaker-examples-2qxnig9os3)
- [x] [Docs](https://github.com/iterative/dvc.org/pull/5108)
- [x] Update branch


dberenbaum:
Need to look into windows tests not passing. Somehow dvclive expects to pass posix paths even on windows, but not sure where that conversion happens.

dberenbaum:
> Need to look into windows tests not passing. Somehow dvclive expects to pass posix paths even on windows, but not sure where that conversion happens.

The cause was because the conversion only happens within a repo, which brought up that we need to decide where paths should be relative to if there is no repo root. I went with cwd for now.

daavoo:
> > Need to look into windows tests not passing. Somehow dvclive expects to pass posix paths even on windows, but not sure where that conversion happens.
> 
> The cause was because the conversion only happens within a repo, which brought up that we need to decide where paths should be relative to if there is no repo root. I went with cwd for now.

I see. Probably no need to do with `cwd` (not even sure if it might be potentially problematic)

The `posix` and `relpath` operations should be decoupled but I mixed them in a single function (they were actually 2 separate bugfixes one for windows other for monorepo in Studio).
I assume that in norepo scenario we don't really need `relpath` operation because the default values are already as expected but we do need `posix`

src/dvclive/dvc.py
```diff
@@ -147,8 +147,12 @@ def mark_dvclive_only_ended():
...
 def get_random_exp_name(scm, baseline_rev):
+    from dvc.repo.experiments.utils import gen_random_name
     from dvc.repo.experiments.utils import (
         get_random_exp_name as dvc_get_random_exp_name,
     )
 
-    return dvc_get_random_exp_name(scm, baseline_rev)
+    if scm and baseline_rev:
+        return dvc_get_random_exp_name(scm, baseline_rev)
+    # TODO: ping studio for list of existing names to check against
```
dberenbaum, line 157: We should make a follow-up issue to make sure we don't overwrite existing live exps

dberenbaum:
> * Make `DVC_EXP_BASELINE_REV` and/or `DVC_EXP_NAME` optional

This depends on:
* https://github.com/iterative/dvc/pull/9806 (for exp name)
* https://github.com/iterative/dvc-studio-client/pull/72 (for baseline rev)
* Studio work to allow for requests with a missing baseline rev (these would have to be permanently "detached" unless we later find some way to connect them back to a baseline)

dberenbaum:
> I see. Probably no need to do with `cwd` (not even sure if it might be potentially problematic)

Updated to remove that.

dberenbaum:
> * Example

See https://colab.research.google.com/drive/17X5OyyUi73PXGyZl_87Pl4yUoNS5lMWn?usp=sharing

shcheklein:
@dberenbaum do you have a sense on what is the mid-term plan for this? If we are not attached to Git - what is the benefit of using the DVC ecosystem? (sorry, I know it's a non-trivial question, definitely not a blocker of this PR or effort). Data management? something else?

Is there a way to clone the repo in those system under the hood / seamlessly for end-users?

Do we want to cover scenarios like when people don't use Git at all - they don't have a repo at all even? (we pretty much then become two products - Git based and separately MlFlow/W&B, right?

dberenbaum:
> Do we want to cover scenarios like when people don't use Git at all - they don't have a repo at all even? (we pretty much then become two products - Git based and separately MlFlow/W&B, right?

It's not in my plans at this point. This is mainly to cover scenarios where you have a project in a repo but want to run an experiment in an environment where it's painful/impossible to have a full copy of the Git repo. In my experience, it's not uncommon to have a notebook that gets checked into Git but that runs independently in databricks, colab, etc. It's also helpful for the SageMaker scenario where you can kick off a job on a remote machine, but that machine only has access to a single .py file, so this way you can at least see metrics populated in Studio.

shcheklein:
Okay, I see. What is the plan of connecting it back? Those experiments will stay "studio"-only, is it correct?

And why can't we try to do a git clone? any specific obstacles for that?

dberenbaum:
TODO: follow up on why this PR isn't passing on Windows.

dberenbaum:
@mnrozhkov To try this out, install this branch on all the ray workers (`pip install git+https://github.com/iterative/dvclive.git@no-git`) and ensure they have the `DVC_STUDIO_TOKEN` and `DVC_STUDIO_REPO_URL` env vars.

shcheklein:
@mattseddon i think you should be aware by this change as well (baseline is optional)

dberenbaum:
> @mattseddon i think you should be aware by this change as well (baseline is optional)

In the end, baseline is still required to be sent to dvc-studio-client in this PR and defaults to `"0" * 40`. It seems this is an accepted value for a null ref (see https://stackoverflow.com/questions/18977462/git-sha1-0000000000000000000000000000000000000000-all-zeroes-is-this-normal).

src/dvclive/live.py
```diff
@@ -95,8 +95,11 @@ def __init__(
         self._report_notebook = None
         self._init_report()
 
-        self._baseline_rev: Optional[str] = None
-        self._exp_name: Optional[str] = exp_name
+        self._baseline_rev: str = os.getenv(
+            env.DVC_EXP_BASELINE_REV,
+            "0" * 40,  # noqa: PLW1508
```
mattseddon, line 100: [Q] Can you create a variable out of this so that we at least have some breadcrumbs for the knowledge? I think we should add a reference to what 40 0s is in Studio as well because this will either get buried or lost.

mattseddon: could be `NO_REPO_SHA` or similar

AlexandreKempf (approved):

src/dvclive/dvc.py
```diff
@@ -131,9 +132,14 @@ def _update_entries(old, new, key):
 def get_exp_name(name, scm, baseline_rev) -> str:
```
dberenbaum, line 132: Updated `get_exp_name` to always be able to return an experiment name instead of only when inside a repo.

src/dvclive/live.py
```diff
@@ -148,28 +150,36 @@ def _init_cleanup(self):
...
             self._init_dvc_pipeline()
         self._dvc_repo = get_dvc_repo()
 
+        scm = self._dvc_repo.scm if self._dvc_repo else None
+        if isinstance(scm, NoSCM):
+            scm = None
+        if scm:
+            self._baseline_rev = scm.get_rev()
+        self._exp_name = get_exp_name(self._exp_name, scm, self._baseline_rev)
+        logger.info(f"Logging to experiment '{self._exp_name}'")
```
dberenbaum, line 167: Once we have tried to find a repo, we can generate all exp info we need instead of scattering it throughout the code

src/dvclive/studio.py
```diff
@@ -34,9 +34,8 @@ def _cast_to_numbers(datapoints):
 
 
 def _adapt_path(live, name):
-    if live._dvc_repo is not None:
-        name = rel_path(name, live._dvc_repo.root_dir)
-    return name
+    dvc_root_path = live._dvc_repo.root_dir if live._dvc_repo else None
+    return rel_path(name, dvc_root_path)
```
dberenbaum, line 38: Need to adapt path even when no repo exists

src/dvclive/utils.py
```diff
@@ -192,8 +192,12 @@ def wrapper(*args, **kwargs):
 
 
 def rel_path(path, dvc_root_path):
-    absolute_path = Path(path).absolute()
-    return str(Path(os.path.relpath(absolute_path, dvc_root_path)).as_posix())
+    if dvc_root_path:
+        absolute_path = Path(path).absolute()
+        path = os.path.relpath(absolute_path, dvc_root_path)
+    if os.name == "nt":
+        return str(PureWindowsPath(path).as_posix())
```
dberenbaum, line 199: Convert windows-style paths to unix

tests/test_dvc.py
```diff
@@ -29,27 +29,24 @@ def test_get_dvc_repo_subdir(tmp_dir):
 def test_exp_save_on_end(tmp_dir, save, mocked_dvc_repo):
     live = Live(save_dvc_exp=save)
     live.end()
+    assert live._baseline_rev is not None
+    assert live._exp_name is not None
```
dberenbaum, line 33: These now get set always

tests/test_dvc.py
```diff
@@ -60,31 +57,6 @@ def test_exp_save_skip_on_env_vars(tmp_dir, monkeypatch, mocker):
     assert live._inside_dvc_pipeline
 
 
-def test_exp_save_run_on_dvc_repro(tmp_dir, mocker):
```
dberenbaum, line 63: This test was actually outdated and not working. It conflicts with `test_dvc_repro` below, so I consolidated the meaningful assertions into that test.

tests/test_dvc.py
```diff
@@ -29,27 +29,24 @@ def test_get_dvc_repo_subdir(tmp_dir):
...
         mocked_dvc_repo.experiments.save.assert_not_called()
 
 
-def test_exp_save_skip_on_env_vars(tmp_dir, monkeypatch, mocker):
+def test_exp_save_skip_on_env_vars(tmp_dir, monkeypatch):
     monkeypatch.setenv(DVC_EXP_BASELINE_REV, "foo")
     monkeypatch.setenv(DVC_EXP_NAME, "bar")
     monkeypatch.setenv(DVC_ROOT, tmp_dir)
 
-    mocker.patch("dvclive.live.get_dvc_repo", return_value=None)
```
dberenbaum, line 52: Not needed

tests/test_dvc.py
```diff
@@ -166,7 +138,7 @@ def test_errors_on_git_add_are_catched(tmp_dir, mocked_dvc_repo, monkeypatch):
     mocked_dvc_repo.scm.untracked_files.return_value = ["dvclive/metrics.json"]
     mocked_dvc_repo.scm.add.side_effect = DvcException("foo")
 
-    with Live(dvcyaml=False) as live:
```
dberenbaum, line 169: Not needed

src/dvclive/live.py
```diff
@@ -95,8 +97,8 @@ def __init__(
         self._report_notebook = None
         self._init_report()
 
-        self._baseline_rev: Optional[str] = None
-        self._exp_name: Optional[str] = exp_name
+        self._baseline_rev: str = os.getenv(env.DVC_EXP_BASELINE_REV, NULL_SHA)
```
shcheklein, line 100: Q: why using a special SHA? can we just keep it None - that would be more explicit I think

tests/test_dvc.py
```diff
@@ -204,10 +176,12 @@ def test_no_scm_repo(tmp_dir, mocker):
...
 
 
-def test_dvc_repro(tmp_dir, monkeypatch, mocker):
+def test_dvc_repro(tmp_dir, monkeypatch, mocked_dvc_repo, mocked_studio_post):
     monkeypatch.setenv(DVC_ROOT, "root")
-    mocker.patch("dvclive.live.get_dvc_repo", return_value=None)
     live = Live(save_dvc_exp=True)
+    assert live._baseline_rev is not None
+    assert live._exp_name is not None
+    assert not live._studio_events_to_skip
```
dberenbaum, line 184: Consolidated from the deleted `test_exp_save_run_on_dvc_repro` above. On `repro`, everything will still work except actually saving an exp ref.

tests/test_post_to_studio.py
```diff
@@ -310,9 +309,8 @@ def test_post_to_studio_inside_subdir_dvc_exp(
     )
 
 
-def test_post_to_studio_requires_exp(tmp_dir, mocked_dvc_repo, mocked_studio_post):
-    assert Live(save_dvc_exp=False)._studio_events_to_skip == {"start", "data", "done"}
-    assert not Live()._studio_events_to_skip
+def test_post_to_studio_without_exp(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+    assert not Live(save_dvc_exp=False)._studio_events_to_skip
```
dberenbaum, line 313: Live updates to Studio will happen regardless of `save_dvc_exp` value.

src/dvclive/live.py
dberenbaum: Just going for the simplest solution rather than one that requires updates in dvc-studio-client (see https://github.com/iterative/dvc-studio-client/pull/72) and studio. Didn't feel like there is justification to go for the more complex solution.

shcheklein: yep, my concern that we are getting some tech debt here potentially, and then it takes 2x+ time to change it to something. I'm fine to go with this, just want to quickly double check - `dvc-studio-client ` change doesn't look complicated (?), on the Studio side we also can do whatever we want initially (e.g. convert to 40x0 in API if needed - at least we would isolate it to the Studio side vs open source client library?)

dberenbaum: I will leave it up to you and @mattseddon. Not a strong preference from me. I would rather not need the baseline rev in dvclive at all but don't see an easy way to drop it completely.

mattseddon: My $0.02 is that the way `dvc-studio-client`/`dvc`/`dvclive`/`studio` are setup for dev on live experiments at the moment is a complete PITA to work with. Given that `baseline_sha` is required/non-nullable (IIUC) by the model in Studio AND `dvc-studio-client` I would say that we do this for now.

```
class Exp(SoftDeleteModel, LifecycleModel):
    repo_url = models.URLField()
    baseline_sha = models.CharField(max_length=40)
```

```
BASE_SCHEMA = Schema(
    {
        Required("type"): choices("start", "done", "data"),  # No "interrupt" for now
        Required(
            "repo_url",
        ): str,  # TODO: use some url validator, voluptuous.Url is too strict
        Required("baseline_sha"): Sha,
```

I think we should also move `dvc-studio-client` into `dvc` which is (maybe) what was being suggested [here](https://github.com/iterative/dvc/pull/10284#issuecomment-1931266683).

dberenbaum: It took me awhile to understand your comment here @mattseddon. To confirm:
1. You prefer we keep this PR as is with the `NULL_SHA` value?
2. The code blocks you pasted are the existing snippets from `studio` and `dvc-studio-client` that show that `baseline_sha` is required in both of those projects?
3. You want to move `dvc-studio-client` into `dvc`, but it's not a blocker for this PR?

mattseddon: 1. yes
2. yes
3. yes
👍🏻

mattseddon: Note to self: don't try to resolve the baseline information for this sha.

src/dvclive/live.py
```diff
@@ -148,28 +150,36 @@ def _init_cleanup(self):
...
+        if not scm:
             if self._save_dvc_exp:
                 logger.warning(
                     "Can't save experiment without a Git Repo."
                     "\nCreate a Git repo (`git init`) and commit (`git commit`)."
                 )
                 self._save_dvc_exp = False
             return
-        if self._dvc_repo.scm.no_commits:
+        if scm.no_commits:
```
shcheklein, line 182: it seems `scm` can be None here

src/dvclive/utils.py
shcheklein: Was it handled before, caught by tests, or a separate fix along the way?

src/dvclive/studio.py
shcheklein: minor: since we are touching this - can we add types?

src/dvclive/live.py
```diff
@@ -571,7 +558,8 @@ def make_report(self):
 
     @catch_and_warn(DvcException, logger)
     def make_dvcyaml(self):
-        make_dvcyaml(self)
+        if self.dvc_file:
```
shcheklein, line 561: what is the case for this?

shcheklein: I mean when can it be `None`?

src/dvclive/studio.py
shcheklein: what is the semantics of a rel_path with None passed to it? why do we need to adapt it anyways if there is no repo?

dberenbaum: > minor: since we are touching this - can we add types?

Added types

> what is the semantics of a rel_path with None passed to it? why do we need to adapt it anyways if there is no repo?

It is only about conversion to windows-style paths and only impacts studio tests, so I refactored to leave `rel_path` untouched and include the windows formatting directly in `_adapt_path`.

src/dvclive/utils.py
dberenbaum: It was caught by CI tests on Windows. It only impacts posting to Studio in Windows without a repo, so it is new to this scenario. Moved this logic outside of `rel_path`.

shcheklein (approved):

src/dvclive/live.py
shcheklein: this is the only (?) unresolved question? Approved, since it doesn't look anything major ... 


good stuff, Dave. thanks!

dberenbaum: Sorry, missed this one. Good catch. Maybe this came from an earlier iteration. I dropped this condition in the latest commit.

PR #649, opened 2023-08-04
HuggingFace improvements

### LINKED ISSUE #641: HuggingFace: Add `log_model`
Similar to https://github.com/iterative/dvclive/pull/640

daavoo:
- Log `Trainer.args`.

- Add `log_model` argument.
Every framework does its own thing. No strong opinions but I went with the following:

  - If `None` (default) will not log any artifact.
  - If `all` will call log_artifact with `output_dir` at each `on_save` call.
  - If `True` will save the model `on_train_end` and call `log_artifact` with type=model and copy=True.
    Will use `best` as name if `args.load_best_model_at_end` else `last`.

- Add Notebook/Colab example

Closes #641 

---

TODO:

- [ ] dvc.org updates

### DIFF
--- a/src/dvclive/huggingface.py
+++ b/src/dvclive/huggingface.py
@@ -32,15 +32,7 @@ class DVCLiveCallback(TrainerCallback):
         control: TrainerControl,
         **kwargs,
     ):
-        for key, value in args.to_dict().items():
-            if key in (
-                "num_train_epochs",
-                "weight_decay",
-                "max_grad_norm",
-                "warmup_ratio",
-                "warmup_steps",
-            ):
-                self.live.log_param(key, value)
+        self.live.log_params(args.to_dict())
 
     def on_log(
         self,
--- a/src/dvclive/huggingface.py
+++ b/src/dvclive/huggingface.py
@@ -1,6 +1,6 @@
 # ruff: noqa: ARG002
 import os
-from typing import Literal, Optional
+from typing import Literal, Optional, Union
 
 from transformers import (
     TrainerCallback,
@@ -18,7 +18,7 @@ class DVCLiveCallback(TrainerCallback):
     def __init__(
         self,
         live: Optional[Live] = None,
-        log_model: Optional[Literal["all", "last"]] = None,
+        log_model: Optional[Union[Literal["all"], bool]] = None,
         **kwargs,
     ):
         super().__init__()
@@ -67,7 +67,8 @@ class DVCLiveCallback(TrainerCallback):
             fake_trainer = Trainer(
                 args=args, model=kwargs.get("model"), tokenizer=kwargs.get("tokenizer")
             )
-            output_dir = os.path.join(args.output_dir, "last")
+            name = "best" if args.load_best_model_at_end else "last"
+            output_dir = os.path.join(args.output_dir, name)
             fake_trainer.save_model(output_dir)
-            self.live.log_artifact(output_dir, type="model", copy=True)
+            self.live.log_artifact(output_dir, name=name, type="model", copy=True)
         self.live.end()
--- a/src/dvclive/huggingface.py
+++ b/src/dvclive/huggingface.py
@@ -1,5 +1,6 @@
 # ruff: noqa: ARG002
-from typing import Optional
+import os
+from typing import Literal, Optional
 
 from transformers import (
     TrainerCallback,
@@ -14,9 +15,14 @@ from dvclive.utils import standardize_metric_name
 
 
 class DVCLiveCallback(TrainerCallback):
-    def __init__(self, model_file=None, live: Optional[Live] = None, **kwargs):
+    def __init__(
+        self,
+        live: Optional[Live] = None,
+        log_model: Optional[Literal["all", "last"]] = None,
+        **kwargs,
+    ):
         super().__init__()
-        self.model_file = model_file
+        self._log_model = log_model
         self.live = live if live is not None else Live(**kwargs)
 
     def on_train_begin(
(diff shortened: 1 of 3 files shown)


dberenbaum:
> * If `last` will save the model `on_train_end` and call `log_artifact` with type=model and copy=True.

Is it possible to do `best` instead of/in addition to `last`?

src/dvclive/huggingface.py
```diff
@@ -53,10 +71,11 @@ def on_train_end(
         control: TrainerControl,
         **kwargs,
     ):
-        if args.load_best_model_at_end:
-            trainer = Trainer(
+        if self._log_model == "last" and state.is_world_process_zero:
```
dberenbaum, line 74: Is it possible to do `best` instead of/in addition to `last`?

src/dvclive/huggingface.py
```diff
@@ -53,10 +71,11 @@ def on_train_end(
...
+            fake_trainer = Trainer(
                 args=args, model=kwargs.get("model"), tokenizer=kwargs.get("tokenizer")
             )
-            trainer.save_model()
-            self.live.log_artifact(args.output_dir)
+            output_dir = os.path.join(args.output_dir, "last")
+            fake_trainer.save_model(output_dir)
+            self.live.log_artifact(output_dir, type="model", copy=True)
```
dberenbaum, line 80: Cross-framework consistency isn't our highest priority, but should we agree on some common principles for the final artifact, like naming and whether to copy it?

src/dvclive/huggingface.py
```diff
@@ -14,11 +15,33 @@
...
+    ):
+        for key, value in args.to_dict().items():
+            if key in (
+                "num_train_epochs",
+                "weight_decay",
+                "max_grad_norm",
+                "warmup_ratio",
+                "warmup_steps",
+            ):
+                self.live.log_param(key, value)
```
dberenbaum, line 43: I know it feels like a lot of unnecessary params, but this goes against what all the other loggers do, right? I think we would need a really strong reason to do this, and IMHO logging dozens of parameters is typical and unproblematic -- even if I probably never care about a parameter, once in awhile I might want to know what it was and it's better to have than not.

dberenbaum (commented):
Have a few questions where we need to align. 

As far as breaking changes, maybe we should go ahead and make the easy ones from the 3.0 checklist and do a release. I think we will need a major version where we transition to `log_model` and moving callbacks into frameworks and can't do it all at once anyway. The alternative is likely to branch-based development and only make the breaking changes available on the 3.0 branch. WDYT?

src/dvclive/huggingface.py
daavoo: Do you mean call the option `best`?
Whether `last` means `best` or not depends on the user passing https://huggingface.co/docs/transformers/main/en/main_classes/trainer#transformers.TrainingArguments.load_best_model_at_end

src/dvclive/huggingface.py
daavoo: I think I would like for all the integrations to just have 2 options:

- `all/checkpoints`: resuming scenarios.
Log the entire checkpoint folder

- `best`: model registry
Log the best checkpoint on end with `copy=True, name="best", type="model")`

dberenbaum: That's fine with me. Do you want to update the lightning logger to use `copy=True`? AFAIK the rest is consistent.

daavoo: will update this and lightning to use that

src/dvclive/huggingface.py
dberenbaum: Sorry, let me resolve this one and discuss in the other comment

src/dvclive/huggingface.py
dberenbaum: Sorry, are you also suggesting to change the behavior of `log_model=True` in lightning to track only the copied best artifact and not the whole directory? That's fine, just want to make sure I understand what you mean.

For HF, how should we handle the last/best checkpoint? If `args.load_best_model_at_end`, we could add `name=best`? WDYT?

SoyGema (commented):
Thanks for making an example! 🙏Would like to propose some additions if you don´t mind. I can´t comment directly due to the notebook nature, so hopefully this can help. If you are time sensitive, and agree on the proposed additions/changes,  happy to submit a contrib.🎺

1. Add **# Goal/Intro** section 
Cover the full functionality the example is exploring at the beginning of the notebook

Proposal : _Example of fine-tuning sentiment analysis classifier based on imbd and distilbert pretrained model , experiment tracking and results metrics analysis with dvclive and dvc api_ 

2. Add section **# Initialize git and dvc** in 2nd cell
why? It gives context about the necessary requirements beyond the libraries install that a data scientist needs to run some other HF examples 

3. Change # Dataset for **# Dataset and Tokenization** in 3rd cell 
The section includes a Tokenization process 

4. Add **# Evaluation metrics** section 
For consistency / coherence with respect to the process

5. Possible discussion L28
Describe in comment what `log_model` does for user?  which seems to be the improvement of the PR?

6. Explain # Comparing section
Context : unclear right now whats really going on with naming . I´ve lost a few chapters of this exploring other things . Thinking about sharing in [#9709](https://github.com/iterative/dvc/issues/9709) . When exploring this, a question came into my mind, _what is the difference in between dvc.api and dvclive ?_

daavoo:
> Thanks for making an example! 🙏Would like to propose some additions if you don´t mind. I can´t comment directly due to the notebook nature, so hopefully this can help. If you are time sensitive, and agree on the proposed additions/changes, happy to submit a contrib.🎺

Thanks @SoyGema ! All points make sense to me. I have opened a separate issue to address as a follow-up. I believe it applies to all the examples and not only to the huggingface one.

src/dvclive/huggingface.py
daavoo: > or HF, how should we handle the last/best checkpoint? If args.load_best_model_at_end, we could add name=best? WDYT?

Updated with this behavior

daavoo:
> As far as breaking changes, maybe we should go ahead and make the easy ones from the 3.0 checklist and do a release. I think we will need a major version where we transition to `log_model` and moving callbacks into frameworks and can't do it all at once anyway. The alternative is likely to branch-based development and only make the breaking changes available on the 3.0 branch. WDYT?

I kept the `model_file` behavior with a warning about deprecation for now

src/dvclive/huggingface.py
daavoo: Also, dropped `last` option in favor of `True`

PR #652, opened 2023-08-04
make_dvcyaml: skip metrics plots when plot=False

### LINKED ISSUE #651: DVC Live log_metric adds a plot to dvc file even though plot=False
`log_metric(..., plot=False)` still adds to `dvc.yaml`:

```
plots:
- plots/metrics:
    x: ...
```

However the `plots/metrics` folder is not created (since `plot=False`). This also causes issues with DVC plugin in VS Code:

image

dberenbaum:
Fixes #651


daavoo (approved):
Thanks @dberenbaum

PR #654, opened 2023-08-05
dvcyaml: warn if tracked as stage output

dberenbaum:
Proposed implementation for https://github.com/iterative/dvclive/issues/456#issuecomment-1665813858.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -123,7 +123,7 @@ class Live:
         if self.dvc_file and os.path.exists(self.dvc_file):
             os.remove(self.dvc_file)
 
-    def _init_check_dvcyaml(self):
+    def _init_check_dvcyaml_overlap(self):
         for stage in self._dvc_repo.index.stages:
             for out in stage.outs:
                 if str(out.fs_path) in str(Path(self.dvc_file).absolute()):
@@ -169,7 +169,7 @@ class Live:
             return
 
         if self._dvcyaml:
-            self._init_check_dvcyaml()
+            self._init_check_dvcyaml_overlap()
 
         if self._inside_dvc_exp:
             return
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -124,15 +124,15 @@ class Live:
             os.remove(self.dvc_file)
 
     def _init_check_dvcyaml(self):
-        if self._dvcyaml:
-            for stage in self._dvc_repo.index.stages:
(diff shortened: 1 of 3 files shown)


src/dvclive/live.py
```diff
@@ -123,6 +123,17 @@ def _init_cleanup(self):
         if self.dvc_file and os.path.exists(self.dvc_file):
             os.remove(self.dvc_file)
 
+    def _init_check_dvcyaml(self):
+        for stage in self._dvc_repo.index.stages:
+            for out in stage.outs:
+                if str(out.fs_path) in str(Path(self.dvc_file).absolute()):
```
daavoo, line 129: Why `self.dvc_file` and not `self.dir`???

tests/test_dvc.py
```diff
@@ -321,3 +321,27 @@ def test_no_scm_repo(tmp_dir, mocker):
 
         live = Live(save_dvc_exp=True)
         assert live._save_dvc_exp is False
+
+
+@pytest.mark.parametrize("dvcyaml", [True, False])
+def test_warn_on_dvcyaml_output_overlap(tmp_dir, mocker, dvcyaml):
```
daavoo, line 327: You should be able to use `mocked_dvc_repo` and update only the properties needed (see examples above int his file)

src/dvclive/live.py
```diff
@@ -157,6 +168,9 @@ def _init_dvc(self):
                 self._save_dvc_exp = False
             return
 
+        if self._dvcyaml:
+            self._init_check_dvcyaml()
```
daavoo, line 172: minor: Would be nice to find a more explicit name

src/dvclive/live.py
daavoo: WDYT about:

```py
str(Path(self.dir).absolute()).startswith(str(out.fs_path))
```

src/dvclive/live.py
```diff
@@ -123,6 +123,17 @@ def _init_cleanup(self):
         if self.dvc_file and os.path.exists(self.dvc_file):
             os.remove(self.dvc_file)
 
+    def _init_check_dvcyaml_overlap(self):
+        for stage in self._dvc_repo.index.stages:
```
daavoo, line 127: I will address in a follow-up: we should wrap any `dvc` operations to catch and not raise exception but log warning instead

daavoo (approved):

Issue #625, opened 2023-07-13
ModuleNotFoundError: No module named 'lightning'


aschuh-hf:
I installed `dvclive=2.12` from `conda-forge` using `mamba`. I have `pytorch_lightning=2.0.4` installed, but not `lightning`. The installation of DVC did not require it either, and the missing import would also be available through `lightning_fabric.utilities.logger` instead of `lightning.fabric.utilities.logger`.

```
    from dvclive.lightning import DVCLiveLogger
  File "/opt/conda/envs/venv/lib/python3.10/site-packages/dvclive/lightning.py", line 5, in <module>
    from lightning.fabric.utilities.logger import (
ModuleNotFoundError: No module named 'lightning'
```

aschuh-hf:
I see that with `pip`, I would have needed to specify extra `lightning` (https://github.com/iterative/dvclive/blob/28a022422c47b885e074b0d13e0adeee77bd545b/pyproject.toml#L65).

But I wonder if it would be possible to modify DVCLive that it suffices to have `pytorch-lightning` installed rather than the more recent `lightning` package? Ray Train 2.5 still uses `pytorch-lightning`.

daavoo:
> But I wonder if it would be possible to modify DVCLive that it suffices to have pytorch-lightning installed rather than the more recent lightning package

Yes, I think we could add some try/excepts on the imports to make it compatible with both packages.

Sorry for the inconvenience, I didn't pay much attention to how/when they decided to migrate and change package names (couldn't find any guide for libraries depending on that) and thought that imports would be somehow compatible after upgrading to the `lightning` name.

We are aiming to contribute and move the loggers to the ML framework repositories to avoid this kind of things

aschuh-hf:
#627 May need a minor adjustment. The following import differs between `lightning` and `pytorch_lightning`.

New `lightning` import:
```
from lightning.pytorch.loggers.utilities import _scan_checkpoints
```

Current alternative import after #627 which raises a `ModuleNotFoundError`:
```
from pytorch_lightning.utilities.logger import _scan_checkpoints
```

Required fix of `dvclive/lightning.py` import statement to make it work:
```
from pytorch_lightning.loggers.utilities import _scan_checkpoints
```

PR #664, opened 2023-08-15
fix broken pytorch-lightning import

dberenbaum:
Per https://github.com/iterative/dvclive/issues/625#issuecomment-1678757747.

The current import works for `pytorch-lightning<2.0`. Added a nested try/except to work for `pytorch-lightning>2.0`. It's getting a bit ugly, but don't see a good reason not to support as many versions as possible for now.

### DIFF
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -22,7 +22,11 @@ except ImportError:
     from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
     from pytorch_lightning.loggers.logger import Logger, rank_zero_experiment
     from pytorch_lightning.utilities import rank_zero_only
-    from pytorch_lightning.utilities.logger import _scan_checkpoints
+
+    try:
+        from pytorch_lightning.utilities.logger import _scan_checkpoints
+    except ImportError:
+        from pytorch_lightning.loggers.utilities import _scan_checkpoints
 from torch import is_tensor
 
 from dvclive import Live


src/dvclive/lightning.py
```diff
@@ -22,7 +22,11 @@
     from pytorch_lightning.callbacks.model_checkpoint import ModelCheckpoint
     from pytorch_lightning.loggers.logger import Logger, rank_zero_experiment
     from pytorch_lightning.utilities import rank_zero_only
-    from pytorch_lightning.utilities.logger import _scan_checkpoints
+
+    try:
```
daavoo, line 26: We could also do an explicit version check but I guess it is equally "ugly

```py
if pkg.parse_version(version('pytorch_lightning')) < pkg.parse_version('2.0'):
```

daavoo (approved):

aschuh-hf:
Thanks for the quick fix, @dberenbaum.

@daavoo Guess I would agree with the sentiment of the meme, except that you guys are importing a private member (`_scan_checkpoints`) that clearly isn't part of their public API...

daavoo:
> @daavoo Guess I would agree with the sentiment of the meme, except that you guys are importing a private member (`_scan_checkpoints`) that clearly isn't part of their public API...

That is true for this case but the rest of the import patches are for public things, including the pure package name.
Anyhow the main thing we have been doing "wrong" is not moving the logger into the framework repo, although it is in the plans.

aschuh-hf:
Oh, yes, things are definitely not in a great intermediate state on their end.

BTW I meant to put a smily after my comment. I don't have problem with how things are and think it makes sense to limit the effort put into this as long as it works. And if it doesn't, you guys are swift in catching up!

PR #668, opened 2023-08-16
dvc: Extract `find_overlapping_stage`.

daavoo:
- Add type hints.
- Simplify if/else warnings in cache


dberenbaum (approved):
I didn't even remember I had written this twice 🤦

PR #669, opened 2023-08-16
log_artifact: update warning inside exp

dberenbaum:
Some fixes to the warnings when logging an artifact inside an experiment:

1. Check if artifact is a subpath of a tracked output
2. `dvc exp remove` -> `dvc remove`


daavoo (changes requested):
Need to rebase the changes from #668

daavoo (approved):

PR #673, opened 2023-08-17
utils: Add catch_and_warn.

daavoo:
dvc: Extract `ensure_dir_is_tracked` to `dvc`.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -5,13 +5,14 @@ import shutil
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set, Union
 
+from dvc.exceptions import DvcException
 from dvc_studio_client.post_live_metrics import post_live_metrics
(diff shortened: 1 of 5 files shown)


src/dvclive/live.py
```diff
@@ -581,35 +583,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
-                self._experiment_rev = self._dvc_repo.experiments.save(
-                    name=self._exp_name,
-                    include_untracked=self._include_untracked,
-                    force=True,
-                    message=self._exp_message,
-                )
-            except DvcException as e:
-                logger.warning(f"Failed to save experiment:\n{e}")
-            finally:
-                mark_dvclive_only_ended()
```
dberenbaum, line 615: Why was this dropped?

dberenbaum:
Are there more places we can use it?

PR #674, opened 2023-08-17
Add experiment name to DVCLive only signal file

mattseddon:
  guide.

Looking to solidify behaviour for https://github.com/iterative/vscode-dvc/issues/4491


daavoo:
CI failure is unrelated, looks like a huggingface update broke a test

daavoo (approved):

mattseddon:
> @mattseddon I just realized, you need to also store baseline_rev. The exp name alone is not guaranteed to be unique, only the pair of baseline rev and exp name is enforced to be unique

This is only to cover a niche use case. 

When an experiment is running under the DVCLive only context (i.e. no DVC) and the workspace is selected for plotting then we want to move that selection from the workspace onto the new experiment when the experiment (training script) finishes running. I don't think the baseline would make a difference in this case because if we can see an experiment with that name under a different baseline it would already have a status (unselected). The new experiment won't have one yet.

mattseddon:
I can still add it if you think it is necessary

PR #678, opened 2023-08-17
live.logger: Honor DVCLIVE_LOGLEVEL env var

daavoo:
I think leaving this out was unintentional @dberenbaum ?


dberenbaum:
🤔 I think so? I can't remember a reason why I did it intentionally, so let's add it back.

Issue #648, opened 2023-08-03
Transition to pipelines


dberenbaum:
Issues with transitioning to pipelines:
1. Not clear what to track as output 
2. Tracking the [entire output folder](https://github.com/iterative/dvclive/issues/456) causes problems with the dvclive dvc.yaml
3. May want to move params from outputs to deps and drop some/all dvclive params to avoid duplication
4. Anything cached by dvclive (for example, using `log_artifact`) needs to be moved to a stage output and removed as a standalone .dvc file
5. DVCLive writes dvc.yaml to subdir but pipeline should be at root

Some ideas for how to help with the transition to pipelines (none are new, but trying to consolidate them here):

1. [Reintroduce](https://github.com/iterative/dvclive/pull/497) printing out a stage template or write one to a file (`dvclive/stage_template.yaml`).
6. Add a top-level `outs` section in `dvclive/dvc.yaml` and a corresponding `dvclive/dvc.lock` file so all outputs are captured within the `dvclive` dir and can be easily dropped/moved to a stage in the top-level `dvc.yaml` (and outputs get dropped each time so you don't have stale outputs).
7. [CLI entry point](https://github.com/iterative/dvclive/issues/648#issuecomment-1665219052) to do an automated, opinionated migration (i..e dvclive.pipeline_migration or something)?
8. When dvc stage add overlaps with an existing output, it can prompt to do dvc remove automatically instead of fail.

shcheklein:
Thanks, @dberenbaum ! Can we also summarize the requirements / issues that we see now in this ticket (can be "as we go" / from the top of your head). From what I remember - params (becomes duplicated and needed to actually run an experiment), loggers (like YOLO - would need additional way to configure it / recognize the pipeline and don't dump certain things for example).

daavoo:
> Some ideas for how to help with the transition to pipelines (none are new, but trying to consolidate them here):

Why don't we introduce an explicit CLI entry point to do all those things (i..e `dvclive.make_pipeline`)?

dberenbaum:
> loggers (like YOLO - would need additional way to configure it / recognize the pipeline and don't dump certain things for example)

Can you clarify? Trying to understand how much is specific to yolo vs general issues with loggers

dberenbaum:
> Why don't we introduce an explicit (call it when you need) CLI entry point to do an automated, opinionated migration (i..e `dvclive.pipeline_migration` or something)?

[...]

PR #679, opened 2023-08-17
Stage template

dberenbaum:
  guide.

Related to #648. Basically adds back #497, but writes to a file instead of printing to the logger.

### DIFF
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -174,13 +174,14 @@ def get_dvc_stage_template(live):
         "deps": ["<my_code_file.py>"],
     }
     outs = []
-    rel_path = Path(os.path.relpath(os.getcwd(), live._dvc_repo.root_dir))
+    root_dir = live._dvc_repo.root_dir
     if live._images and live._cache_images:
-        images_path = (rel_path / live.plots_dir / Image.subfolder).as_posix()
+        plots_path = Path(os.path.relpath(live.plots_dir, root_dir))
+        images_path = (plots_path / Image.subfolder).as_posix()
         outs.append(images_path)
     for o, cache in live._outs.items():
         artifact_path = Path(os.getcwd()) / o
-        artifact_path = artifact_path.relative_to(live._dvc_repo.root_dir).as_posix()
+        artifact_path = artifact_path.relative_to(root_dir).as_posix()
         if cache:
             outs.append(artifact_path)
         else:
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -180,8 +180,7 @@ def get_dvc_stage_template(live):
         images_path = (plots_path / Image.subfolder).as_posix()
         outs.append(images_path)
     for o, cache in live._outs.items():
-        artifact_path = Path(os.getcwd()) / o
-        artifact_path = artifact_path.relative_to(root_dir).as_posix()
+        artifact_path = Path(os.path.relpath(o, root_dir)).as_posix()
         if cache:
             outs.append(artifact_path)
         else:
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -150,3 +150,26 @@ def find_overlapping_stage(dvc_repo: "Repo", path: StrPath) -> Optional["Stage"]
             if str(out.fs_path) in abs_path:
                 return stage
     return None
+
+
+def get_dvc_stage_template(live):
+    stage = {
+        "cmd": "<python my_code_file.py my_args>",
+        "deps": ["<my_code_file.py>"],
+        "outs": [],
+    }
+    rel_path = Path(os.path.relpath(os.getcwd(), live._dvc_repo.root_dir))
+    if live._params:
+        params_path = (rel_path / live.params_file).as_posix()
+        stage["outs"].append({params_path: {"cache": False}})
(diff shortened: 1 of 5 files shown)


src/dvclive/live.py
```diff
@@ -101,6 +103,8 @@ def __init__(
         self._dvc_studio_config: Dict[str, Any] = {}
         self._init_studio()
 
+        self._stage_template_file = Path(self._dir) / "stage_template.yaml"
```
dberenbaum, line 106: No strong opinion on the name

src/dvclive/dvc.py
```diff
@@ -166,3 +166,25 @@ def ensure_dir_is_tracked(directory: str, dvc_repo: "Repo") -> None:
...
         dvc_repo.scm.add(paths_to_track)
+
+
+def get_dvc_stage_template(live):
+    stage = {
+        "cmd": "<python my_code_file.py my_args>",
+        "deps": ["<my_code_file.py>"],
+    }
+    outs = []
+    rel_path = Path(os.path.relpath(os.getcwd(), live._dvc_repo.root_dir))
```
daavoo, line 177: I haven't thought it through but the `getcwd()` feels potentially problematic. Maybe we can do it relative to the parent dir of `Live.dir`?

src/dvclive/live.py
```diff
@@ -456,6 +460,8 @@ def log_artifact(
                         name,
                     )
 
+        self._outs[str(path)] = cache
```
daavoo, line 463: Shouldn't this be nested inside the if?

src/dvclive/live.py
```diff
@@ -525,6 +531,11 @@ def make_report(self):
     def make_dvcyaml(self):
         make_dvcyaml(self)
 
+    @catch_and_warn(Exception, logger)
```
daavoo, line 534: Can we catch `DvcException` instead of a generic Exception?

src/dvclive/live.py
```diff
@@ -525,6 +531,11 @@ def make_report(self):
...
+    def make_stage_template(self):
+        stage_content = get_dvc_stage_template(self)
```
daavoo, line 536: We should check either here or inside `get_dvc_stage_template` the existence of `self._dvc_repo`. We don't do it in make_dvcyaml because we don't try to access it.

tests/test_dvc.py
```diff
@@ -341,3 +338,106 @@ def test_warn_on_dvcyaml_output_overlap(tmp_dir, mocker, mocked_dvc_repo, dvcyam
...
+
+def test_get_dvc_stage_template_empty(tmp_dir, mocked_dvc_repo):
+    live = Live()
+    template = get_dvc_stage_template(live)
+
+    assert template == {
+        "stages": {
+            "dvclive": {
+                "cmd": "<python my_code_file.py my_args>",
+                "deps": ["<my_code_file.py>"],
```
daavoo, line 351: q: why this mention has surrounding `<>` but not the one in cmd?

tests/test_dvc.py
```diff
@@ -341,3 +338,106 @@ def test_warn_on_dvcyaml_output_overlap(tmp_dir, mocker, mocked_dvc_repo, dvcyam
...
+                "outs": ["dvclive/plots/images"],
+            }
+        }
+    }
+
+
+def test_get_dvc_stage_template_chdir(tmp_dir, mocked_dvc_repo, monkeypatch):
+    d = tmp_dir / "sub" / "dir"
+    d.mkdir(parents=True)
+    monkeypatch.chdir(d)
```
daavoo, line 426: I am confused about this test. In this case `cwd` would be `sub/dir` but we dump the paths relative to the root `.dvc`?

tests/test_dvc.py
daavoo, line 1: Thanks for the tests 🙏

tests/test_dvc.py
dberenbaum: cmd does have `<>`? did i misunderstand you?

btw I'm fine if you want to suggest something different for `cmd` and `deps`

daavoo: oh my bad, I didn't notice that and was wondering why my_code_file.py was not surrounded in cmd

tests/test_dvc.py
daavoo: Then in the code, I am not sure why we do need to get `cwd`  instead of just using repo root dir for relpath

dberenbaum: Got it, thanks for the close review. Dropped it in the latest push.

src/dvclive/live.py
dberenbaum: 🤔 Not sure. Any reason not to always add it? No strong opinion.

daavoo: If we do not enter the if we are not doing anything inside the method

daavoo (approved):

dberenbaum:
Not ready to merge this yet. Need to think again about the suggestions in #648 to have some kind of helper instead of manually editing `dvc.yaml` files. Ideally, users shouldn't have to care about these files, plus this PR still has friction with having to manually run `dvc remove`.

dberenbaum:
I'll plan to rework this as writing `dvc remove` and `dvc stage add` commands and make it a follow-up to 3.0 changes.

PR #683, opened 2023-08-22
studio: Ensure `MIN_SECONDS_BETWEEN_CALLS`.

daavoo:
Naive but simple prevention against spamming `data` updates.


daavoo:
Actually looks like the cloudfare is already configured to take care of it

shcheklein:
> Actually looks like the cloudfare is already configured to take care of it

How will this client though react to that? Does CF return 429 or something? Will it back off and retry?

PR #684, opened 2023-08-22
3.0 updates

### LINKED ISSUE #459: report: auto mode and when to show the report
#432 introduces `report="notebook"`. Turning this on automatically may be too aggressive for some, but others may never discover it if we turn it off by default. I also expect that users rely less on the report now that we have other ways to show live updates, like in the VS Code extension or even Studio. 

Do we still need `auto` mode at all? Should it be the default? What should its behavior be? It gives an obvious intro to the reports DVCLive can generate, and it can always be turned off, but I don't think it's behavior is obvious, and adding more report types to it only makes it more complicated.

### LINKED ISSUE #484: 3.0 updates
```[tasklist]
- [ ] https://github.com/iterative/dvclive/issues/499
- [x] Set `save_dvc_exp=True` by default
- [x] Set `report=none` by default
- [ ] https://github.com/iterative/dvclive/issues/704
```

Post-release:

https://github.com/iterative/dvclive/issues/221

### LINKED ISSUE #499: frameworks: Remove `model_file`

daavoo:
Closes #499
Closes #484 
Closes #459


dberenbaum:
We also need accompanying docs updates.

dberenbaum (changes requested):
Nothing wrong with the PR AFAICT, but let's hold on this for #687

daavoo:
Added:

- [report: Drop "auto" logic.](https://github.com/iterative/dvclive/pull/684/commits/5dcc43be9a90ce10f73227316859b98ffd1225ad)

- https://github.com/iterative/dvclive/pull/705

- [refactor(tests): Split test_main into separate files.](https://github.com/iterative/dvclive/pull/684/commits/a02d16099b072fa86e0947cdc5cfe412864e5fbe)

dberenbaum:
I was thinking of waiting on https://github.com/iterative/dvc/pull/9898 since if we have to release before then, it will be ugly for existing projects

dberenbaum:
Need to update dvc version before merging

PR #685, opened 2023-08-22
studio: Add simple guardrails against spam

daavoo:
- `MIN_SECONDS_BETWEEN_CALLS`
- `MAX_CONSECUTIVE_FAILURES`


tests/test_studio.py
```diff
@@ -485,3 +492,104 @@ def test_post_to_studio_message(tmp_dir, mocked_dvc_repo, mocked_studio_post):
...
+            "client": "dvclive",
+        },
+        headers={
+            "Authorization": "token STUDIO_TOKEN",
+            "Content-type": "application/json",
+        },
+        timeout=(30, 5),
+    )
+
+    live.log_metric("foo", 2)
```
shcheklein, line 549: is there a way to capture the log warning? also check assert not called or something like that.

daavoo: Yes

tests/test_studio.py
```diff
@@ -485,3 +492,104 @@ def test_post_to_studio_message(tmp_dir, mocked_dvc_repo, mocked_studio_post):
...
+    live.next_step()
+
+    time.sleep(MIN_SECONDS_BETWEEN_CALLS)
+
+    live.log_metric("foo", 3)
+    live.next_step()
+
+    mocked_post.assert_called_with(
```
shcheklein, line 557: can the request it was called with taken out and only the needed info compared with M matcher? (to simplify the test - easier to maintain later)

shcheklein:
I think we should try to make it a bit more robust by not completely stopping sending all the metrics, but rather dropping events (data) but still trying to send it from time to time.

Also, minimum time between events - should it be handled by Studio rate limiter - send 429 back where the client then can skip the send for example. It feels that a few seconds hard limit is a bit heavy handed and artificial tbh.

daavoo:
The call is on each `next_step` so the 

> send 429 back where the client then can skip the send for example. It feels that a few seconds hard limit is a bit heavy handed and artificial tbh.

how do you know how much to wait when receiving 429? Wouldn't we still need some hardcoded limit there?

daavoo:
> completely stopping sending all the metrics, but rather dropping events (data) but still trying to send it from time to time.

So, we could drop `MAX_CONSECUTIVE_FAILURES` and convert `MIN_SECONDS_BETWEEN_CALLS` to `MIN_SECONDS_BETWEEN_FAILURES`.

I already add `MAX_NUMBER_OF_DATAPOINTS` which can hard limit any call

shcheklein:
> how do you know how much to wait when receiving 429? Wouldn't we still need some hardcoded limit there?

good question, @iterative/studio-backend what is the rate limit at the moment for the `api/live`?

I would need to check, but I think it should be still `{'capacity': 1000, 'restore_rate_hz': 600 / 60}` per user. Means that 1000 requests can be done immediately and then it will be restoring it at 10 requests per second. We are using the token bucket algorithm behind it.

It can be tuned for the live APIs if we need to.

What it practically means here - we can expect that after one second wait we are getting more "tokens" to run queries.

amritghimire:
> > how do you know how much to wait when receiving 429? Wouldn't we still need some hardcoded limit there?
> 
> good question, @iterative/studio-backend what is the rate limit at the moment for the `api/live`?
> 
> I would need to check, but I think it should be still `{'capacity': 1000, 'restore_rate_hz': 600 / 60}` per user. Means that 1000 requests can be done immediately and then it will be restoring it at 10 requests per second. We are using the token bucket algorithm behind it.
> 
> It can be tuned for the live APIs if we need to.
> 
> What it practically means here - we can expect that after one second wait we are getting more "tokens" to run queries.

It is combined value for DEFAULT_IP_RATE_LIMIT and DEFAULT_USER_RATE_LIMIT.

The default value are the `{'capacity': 10000, 'restore_rate_hz': 6000 / 60}` for IP limit and `{'capacity': 1000, 'restore_rate_hz': 600 / 60}` for user rate limit if not changed in production.

daavoo:
I have been testing against the rate limiter and it kind of works as expected. This was originally raised due to misidentifying the source of the Studio breakage as being coming from multiple calls instead of the backend code querying the PlotRow database.

Closing for now but I might open to handle different error codes returned by Studio differently

PR #687, opened 2023-08-25
Write to root dvc.yaml

dberenbaum:
This is still WIP and needs tests, but I started working on moving `dvc.yaml` contents to the root `dvc.yaml` after all. It looks like writing to `dvclive/dvc.yaml` was the wrong decision since it causes a number of issues, like:
1. Making it hard to track the dvclive outputs properly, and relatedly making the pipelines transition harder #456
2. Making model artifacts require a subpath to specify the `dvc.yaml` path where to find them
3. Increasing the likelihood of duplicate metrics, plots, etc.
4. Hard to explain and goes against DVC conventions


src/dvclive/dvc.py
```diff
@@ -71,25 +71,28 @@ def get_dvc_repo() -> Optional["Repo"]:
             return None
 
 
-def make_dvcyaml(live) -> None:
+def make_dvcyaml(live) -> None:  # noqa: C901
```
dberenbaum, line 74: Changes here are to change the paths to be relative to the new `dvc.yaml` location

src/dvclive/dvc.py
```diff
@@ -98,11 +101,53 @@ def make_dvcyaml(live) -> None:
...
             artifact["path"] = Path(relative_path).as_posix()
 
-    dump_yaml(dvcyaml, live.dvc_file)
+    if not os.path.exists(live.dvc_file):
+        dump_yaml(dvcyaml, live.dvc_file)
+    else:
+        update_dvcyaml(live, dvcyaml)
+
+
+def update_dvcyaml(live, updates):  # noqa: C901
```
dberenbaum, line 114: This contains most of the new logic to update an existing `dvc.yaml` file. The basic logic is:
1. For each section in the old `dvc.yaml`, drop the old dvclive entries
2. Merge the remaining old dvclive entries with the new ones

src/dvclive/live.py
```diff
@@ -58,7 +58,7 @@ def __init__(
         resume: bool = False,
         report: Optional[str] = "auto",
         save_dvc_exp: bool = False,
-        dvcyaml: bool = True,
+        dvcyaml: Union[str, bool] = True,
```
dberenbaum, line 61: Changed `dvcyaml` to accept a path so any `dvc.yaml` path can be specified, but defaults to the DVC repo root.

dberenbaum:
Should be ready to review now

shcheklein:
@dberenbaum 😎 I really like the attempt and the idea. If that works and makes it's better - that's fantastic, Dave.

Questions, most of it will be about questions in this PR, I guess. (don't take it as a blocker, I'm 100% to try it):

- To clarify, if the file exists already and we use `dvc exp run` - the new logic will try to update that file (it will keep editing it as the experiment runs), is it correct?
- Do we plan to use / modify stages, outputs? (e.g. to track plots, models).
- Are we trying (or is there a plan) to detect the existing output in stages to avoid `dvc add`-in plots / weigths? What if file is using templates?
- What happens if I do a live experiment, and it creates a record for a plot in the `dvc.yaml`, then I run a new one where I renamed that plot. Will I end up with two records

dberenbaum:
>   * To clarify, if the file exists already and we use `dvc exp run` - the new logic will try to update that file (it will keep editing it as the experiment runs), is it correct?

Yup 🤷 . It won't touch `stages` but will edit the file while the pipeline runs.
 
>   * Do we plan to use / modify stages, outputs? (e.g. to track plots, models).

No, this PR is limited to writing what gets written now to `dvclive/dvc.yaml`. I don't have any immediate intentions of going any further.
 
>   * Are we trying (or is there a plan) to detect the existing output in stages to avoid `dvc add`-in plots / weigths? What if file is using templates?

Sorry, I don't follow what you mean.

>   * What happens if I do a live experiment, and it creates a record for a plot in the `dvc.yaml`, then I run a new one where I renamed that plot. Will I end up with two records

No. To try to match current behavior, this PR will drop any existing plots, metrics, or params which start with the dvclive dir (`dvclive/`), and DVCLive never writes out plots that aren't keyed by that dir. It will also drop any artifacts with a `path` field that starts with `dvclive/`.

For example, all of these will get dropped when the file is updated:

```yaml
metrics:
  - dvclive/metrics.json
plots:
  - dvclive/plots/metrics: 
      x: step
  - dvclive/plots/images
  - dvclive/plots/sklearn/confusion_matrix.json:
      template: confusion
      ...
artifacts:
  model:
    path: dvclive/artifacts/model.pth
    type: model
    desc: seq2seq
```

These won't get dropped:

```yaml
plots:
  - Confusion Matrix: # custom plot id (dvclive never write this)
      template: confusion
      x: actual
      y:
        dvclive/plots/sklearn/confusion_matrix.json: predicted
      ...
artifacts:
  model: # could be logged by dvclive, but we can't know since the path is outside of dvclive/
    path: model.pth
    type: model
    desc: seq2seq # any of these fields might be overwritten if dvclive logs an artifact with name 'model'
```

shcheklein:
@dberenbaum 

> Yup 🤷 . It won't touch stages but will edit the file while the pipeline runs.

that worries me the most, tbh :) It might create some simple issues (e.g. I have a dvc.yaml open and I edit it while another experiment is running but don't save it, then when I'm trying to save it I can hit many different situations since it's now written by machine). More complicated stuff - something that DVC doesn't expect (e.g. it or some DVC Python API reads it in between and hits some inconsistent state (we can mitigate by writing atomically, or it calculates its hash, or ...)

> It will also drop any artifacts with a path field that starts with dvclive/.

does it mean we'll need to do copy=True when we log artifacts? at least by default? 

> Sorry, I don't follow what you mean.

Sorry, I mean if I do log_artifact for example, it can try to check if dir / file is already DVC-tracked and we can skip `dvc add`. Probably it's not related to this PR.

src/dvclive/live.py
```diff
@@ -111,22 +112,7 @@ def _init_resume(self):
...
-            self.params_file,
-            os.path.join(self.dir, "report.html"),
-            os.path.join(self.dir, "report.md"),
-        ):
-            if f and os.path.exists(f):
-                os.remove(f)
-
-        if self.dvc_file and os.path.exists(self.dvc_file):
-            os.remove(self.dvc_file)
+        shutil.rmtree(self.dir)
```
dberenbaum, line 115: If we do this, let's cleanup the entire dvclive dir so that stale `dvclive/dvc.yaml` files get removed. AFAIR, there was not that strong a motivation to selectively cleanup files in the dir, and I think we have established the convention that this dir will be overwritten each run.

dberenbaum: Okay, I went with globbing for any `dvc.yaml` files inside the dvclive dir and deleting those.

dberenbaum:
The PR should be ready now, but would still like to play with it a bit more before deciding if this is the right direction.

@daavoo Would also be curious to hear your thoughts on this change overall.

dberenbaum:
When making this change for an existing project, this causes a plots mismatch in Studio due to the change in `dvc.yaml` path:

Screenshot 2023-08-29 at 8 29 36 PM

Related: https://github.com/iterative/studio/issues/5503

src/dvclive/live.py
```diff
@@ -536,8 +549,10 @@ def make_report(self):
             if self._report_mode == "html" and env2bool(env.DVCLIVE_OPEN):
                 open_file_in_browser(self.report_file)
 
+    @catch_and_warn(DvcException, logger)
     def make_dvcyaml(self):
-        make_dvcyaml(self)
+        if self.dvc_file:
```
daavoo, line 554: I am missing something or we are now ignoring `dvcyaml=False`?

tests/test_dvc.py
daavoo, line 1: Let's create a separate `test_make_dvcyaml.py`

tests/test_log_artifact.py
daavoo, line 1: Same as above, I would prefer to parameterized instead of update the values

src/dvclive/live.py
daavoo: Why not drop the `bool` and make a default string?

src/dvclive/live.py
daavoo: I was missing something

daavoo:
I merged main and that caused some test failures

> When making this change for an existing project, this causes a plots mismatch in Studio due to the change in `dvc.yaml` path:
> 
> Screenshot 2023-08-29 at 8 29 36 PM
> Related: [iterative/studio#5503](https://github.com/iterative/studio/issues/5503)

To me that is more related to the idea of dropping the dvcyaml prefix unless there is a name conflict

daavoo:
> The PR should be ready now, but would still like to play with it a bit more before deciding if this is the right direction.
> 
> @daavoo Would also be curious to hear your thoughts on this change overall.

I am on board with the idea of the change.
I still haven't but want to do a proper review of the code changes.
My main worry has already been shared, modifying the dvc.yaml while the experiment runs. I want manually try out some scenarios

tests/test_log_artifact.py
```diff
@@ -67,45 +67,45 @@ def test_log_artifact_with_save_dvc_exp(tmp_dir, mocker, mocked_dvc_repo):
 def test_log_artifact_type_model(tmp_dir, mocked_dvc_repo):
     (tmp_dir / "model.pth").touch()
 
-    with Live() as live:
+    with Live(dvcyaml="dvc.yaml") as live:
```
daavoo, line 70: Shouldn't this be the default behavior when no argument is passed? Why do we need to explicitly pass it in the test?

dberenbaum:
> To me that is more related to the idea of dropping the dvcyaml prefix unless there is a name conflict

Yes, sorry, that's what I meant. Is there an issue for this?

tests/test_vscode.py
```diff
@@ -91,23 +91,3 @@ def test_vscode_dvclive_only_signal_file(tmp_dir, dvc_root, mocker):
...
-    dvc_stage.outs = [dvc_out]
-    mocked_dvc_repo.index.stages = [dvc_stage]
-    live = Live(dvcyaml=dvcyaml)
-
-    if dvcyaml:
-        msg = f"'{live.dvc_file}' is in outputs of stage 'train'.\n"
-        msg += "Remove it from outputs to make DVCLive work as expected."
-        logger.warning.assert_called_with(msg)
-    else:
-        logger.warning.assert_not_called()
```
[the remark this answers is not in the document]
dberenbaum: Okay, looks like we have a similar test [here](https://github.com/iterative/dvclive/blob/0d9dd2cfd639833a5e90b1269497fac494f13e82/tests/test_dvc.py#L649-L665). Should we add the vs code mark?

tests/test_dvc.py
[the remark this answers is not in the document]
dberenbaum: Couldn't see exactly where you meant here, but tried to parametrize where it made sense.

tests/test_vscode.py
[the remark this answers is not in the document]
daavoo: I think I unintentionally put it in vscode but it is not really vscode related

dberenbaum:
> When making this change for an existing project, this causes a plots mismatch in Studio due to the change in `dvc.yaml` path:
> Screenshot 2023-08-29 at 8 29 36 PM
> 
> Related: [iterative/studio#5503](https://github.com/iterative/studio/issues/5503)

This isn't only a Studio issue. In DVC CLI and VS Code, the plots are also split:

visualization(12)
visualization(13)

Screenshot 2023-08-31 at 2 14 49 PM

I can try to start by dropping the `dvc.yaml` prefix on the DVC side and see if that helps.

Any other blockers to merging this?

src/dvclive/dvc.py
daavoo: What would be the harm of simplifying the logic and only adding new content if there is no matching "key" in the dvc.yaml? Not trying to drop "outdated" sections, not trying to update props, etc.

We are defining what is the excepted thing here as it is the first time we have done it, so I think it would be okay to start with that and wait for people to ask for more complex update logic.

We would be simplifying code and the number of occasions we would "dynamically" update the `dvc.yaml`

daavoo:
> This isn't only a Studio issue. In DVC CLI and VS Code, the plots are also split:

Yes, it all originated in DVC with the addition of top-level plots where we started adding the prefix.

> Any other blockers to merging this?

I left a comment about the `update_yaml` logic. Not a blocker, but I would prefer to lie on the side of doing less and simpler code to begin with

src/dvclive/live.py
dberenbaum: How would you turn it off? https://github.com/iterative/example-get-started for example doesn't use it and adding it would create duplicate plots.

src/dvclive/dvc.py
dberenbaum: > What would be the harm of simplifying the logic and only adding new content if there is no matching "key" in the dvc.yaml?

What do you suggest if there is a matching key? Let's say the artifact labels change. We would at least want to update that artifact, right?

I basically had 2 reasons for doing it this way:
1. It was the closest to mimicking current behavior (we have sort of defined expectations from what we do today in `dvclive/dvc.yaml`).
2. I didn't see how to simplify the logic much more.

Most of the complexity is because:
- metrics, plots, and params are all lists even though they work like dicts (keys must be unique), which makes it hard to merge them. We can't make a set because we could end up with something like `[{"dvclive/plots/custom/roc.json": {"title": "ROC"}}, {"dvclive/plots/custom/roc.json": {"title": "ROC Curve"}}]`, which is invalid for dvc.
- artifacts are easy to merge since they are dicts but they are keyed by name rather than path, so I went through the extra trouble to also make sure we dropped other dvclive paths [here](https://github.com/iterative/dvclive/blob/7e9554a366c59fce7d185193f18b2cc51bbb7a7e/src/dvclive/dvc.py#L120-L122). We could drop this part, but it felt consistent with the rest of the logic and only added 3 lines/1 loop.

src/dvclive/live.py
daavoo: Optional[str] = "dvc.yaml"

And you pass dvcyaml=None.

Anyhow, it is a minor thing, no strong opinion

dberenbaum: It would be slightly more "breaking" but is slightly cleaner 🤔 . No strong opinion from me either.

daavoo (approved):
Minor comment but not blocker

src/dvclive/live.py
daavoo: Let's go with this for now, not major

src/dvclive/dvc.py
daavoo: It makes sense, I couldn't find a simpler way to cover all cases. As discussed offline, let's go with this and see how it works

src/dvclive/live.py
```diff
@@ -263,7 +265,18 @@ def metrics_file(self) -> str:
...
+                return self._dvcyaml
+            raise InvalidDvcyamlError
+        if self._dvc_repo is not None:
+            return os.path.join(self._dvc_repo.root_dir, "dvc.yaml")
+        self._dvcyaml = False
+        logger.warning(
+            "Can't infer dvcyaml path without a DVC repo. "
+            "`dvc.yaml` file will not be written."
+        )
+        return ""
```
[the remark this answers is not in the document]
dberenbaum: Can we use https://docs.python.org/3/library/functools.html#functools.cached_property?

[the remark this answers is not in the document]
dberenbaum: Okay, moved it inside a new `_init_dvc_file` method since `_init_dvc` is already convoluted. PTAL. Not sure if there's value in keeping the `dvc_file` property method here vs. setting `self.dvc_file` directly in `_init_dvc`, but no strong opinion.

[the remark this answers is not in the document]
daavoo: > Not sure if there's value in keeping the dvc_file property method here vs. setting self.dvc_file directly in _init_dvc, but no strong opinion.

I think the value is that users can't modify/set the value of `dvc_file` directly, which is what we want

PR #688, opened 2023-08-27
Add step completed signal file for VS Code

mattseddon:
  guide.

----

In `vscode-dvc`, we have [this issue](https://github.com/iterative/vscode-dvc/issues/4528) which is caused by VS Code dropping events for files within deeply nested folders when running on a remote machine. This can happen when a user has a slightly nested folder structure and is running experiments in the queue. An example of a file path that should trigger a watcher event but doesn't is: `{root}/.dvc/tmp/exps/{tmp_queue_dir}/data/train/metrics.json`

My suggestion [here](https://iterativeai.slack.com/archives/C05K8642QGZ/p1692917483377809?thread_ts=1692873983.370429&cid=C05K8642QGZ) to the user was for him to update his training script to touch one of the files further up the tree (e.g. the root `dvc.yaml`) but we can do the same thing with the DVCLive meaning that the extension should be guaranteed to work out of the box for users using DVCLive as a logger.


src/dvclive/dvc.py
```diff
@@ -46,6 +51,22 @@ def _find_dvc_root(root: Optional[StrPath] = None) -> Optional[str]:
...
+
+    if not non_queue_root_dir:
+        return None
+
+    temp_dir_path = os.path.join(".dvc", "tmp", "exps")
+
+    is_temp_queue_dir = temp_dir_path in non_queue_root_dir
+    if is_temp_queue_dir:
+        non_queue_root_dir = _find_non_queue_root_dir(
+            os.path.abspath(os.path.join(non_queue_root_dir, "..", "..", "..", ".."))
```
mattseddon, line 65: Happy for pointers on how to make this less ugly.

tests/test_main.py
```diff
@@ -400,6 +400,55 @@ def test_context_manager_skips_end_calls(tmp_dir):
     assert (tmp_dir / live.metrics_file).exists()
 
 
+@pytest.mark.vscode()
+@pytest.mark.parametrize("dvc_root", [True, False])
+def test_vscode_dvclive_step_completed_signal_file(tmp_dir, dvc_root, mocker):
+    signal_file = os.path.join(
+        tmp_dir, ".dvc", "tmp", "exps", "run", "DVCLIVE_STEP_COMPLETED"
```
daavoo, line 407: High-level question, how does this work when queued experiments are running in parallel?

mattseddon: The file will get overwritten as each experiment completes a step. 

IMO this is okay because all we need is for an event to be created so that the extension can reliably run `exp show` at the correct time. We are not currently interrogating the contents of the file but we could use the PID to clean it up when it is stale.

src/dvclive/dvc.py
daavoo: We already do this for the exp name and baseline rev:

https://github.com/iterative/dvc/blob/ef260eaf2e2c207edc26505f510e4f79b11ba6a4/dvc/repo/experiments/queue/base.py#L344-L350

In those lines, you could access `self.repo.root_dir` and it should point to the right place

mattseddon: Would that be correct for the temp directories running from the queue? Would it have to be exposed by `dvc queue start`? (I have no idea that's why I'm asking)

tests/test_main.py
daavoo: To elaborate, we have a single signal file but we could have multiple running experiments in parallel writing to the same file

src/dvclive/dvc.py
daavoo: > Would that be correct for the temp directories running from the queue? Would it have to be exposed by dvc queue start and then would it need to be passed to each worker? (I have no idea that's why I'm asking)

I believe it would be correct in any case. The `run_env` gets serialized when adding experiments to the queue (i.e. `dvc exp run --queue`).
We rely on a mechanism that deserializes and exposes those env vars so the code in the stages can access them.
That is why we can retrieve the right `exp_name` even when there are multiple exps running in the queue

tests/test_main.py
mattseddon: For the experiments table in the extension, we watch for file events in order to call for data updates. These watched files include params and metrics files (gathered from the `exp show` output). When there are experiments running in the queue on remote machines some of the necessary events are dropped which means the table can get stale. 

The single purpose of this file is to make sure that the extension calls `exp show` at the right time. If there are multiple experiments running then we want to call for a data update whenever one of the experiments completes a step.

With all of that said maybe the contents of the file are wrong. I was just concerned that if we didn't actually change the contents of the file that VS Code would drop those events as well.

src/dvclive/dvc.py
mattseddon: This would also mean that users would have to update to the most recent version of DVC and DVCLive when it is released, right?

daavoo: > This would also mean that users would have to update to the most recent version of DVC and DVCLive when it is released, right?

Yes

pyproject.toml
```diff
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "dvc>=2.58.0",
+    "dvc>=3.17.0",
```
daavoo, line 34: Added the bump

daavoo (approved):

PR #689, opened 2023-08-28
updates test for dvc-studio-client changes

dberenbaum:
Fixes tests after https://github.com/iterative/dvc-studio-client/pull/78


shcheklein (approved):

pyproject.toml
```diff
@@ -33,7 +33,7 @@ dynamic = ["version"]
 dependencies = [
     "dvc>=2.58.0",
     "dvc-render>=0.5.0,<1.0",
-    "dvc-studio-client>=0.10.0,<1",
+    "dvc-studio-client>=0.13.0,<1",
```
dberenbaum, line 36: Guess we would need a new dvc-studio-client release and to bump the version here

dberenbaum:
@daavoo Anything to add here? Any objections to a `dvc-studio-client` version bump? Should we do 1.0?

daavoo:
> @daavoo Anything to add here?

tests are failing. I think the import is wrong

>  Any objections to a `dvc-studio-client` version bump? Should we do 1.0?

I think this is only used in tests so it should not be really a major version bump, strictly speaking. The "public" methods didn't change

dberenbaum:
> tests are failing. I think the import is wrong

It's because we haven't released those changes yet, so I am pointing to the old version without them.

> I think this is and env var that we just happen to use in tests here, so it should not be really a major version bump, strictly speaking. The "public" methods didn't change

👍

daavoo:
> > tests are failing. I think the import is wrong
> 
> It's because we haven't released those changes yet, so I am pointing to the old version without them.
> 
> > I think this is and env var that we just happen to use in tests here, so it should not be really a major version bump, strictly speaking. The "public" methods didn't change
> 
> 👍
> 
> So any objection to a dvc-studio-client 0.14.0 release?

no objections

pyproject.toml
dberenbaum, line 36:
```suggestion
    "dvc-studio-client>=0.14.0,<1",
```

Issue #670, opened 2023-08-16
warn on `repro` usage


dberenbaum:
Saving the experiment and other features of dvclive won't work when using `repro` instead of `exp run`. We need to detect when `repro` is being used and warn the users to do `exp run` instead.

omesser:
@dberenbaum @daavoo  - 2 questions:
- warn vs fail - Do we want to warn just to not break existing user workflows - should we state that in the future (next major release) this will be a failure and not supported? are there any downside of asking users to do `exp run` ?
- verifying I'm not missing anything - `dvc repro` is problematic, `dvc exp run` isn't, and running with just Python (directly, via shebang, or some wrapper like [pdm](https://pdm.fming.dev/latest/)) is safe? obvious example: `python script.py`

dberenbaum:
> * warn vs fail - Do we want to warn just to not break existing user workflows - should we state that in the future (next major release) this will be a failure and not supported? are there any downside of asking users to do `exp run` ?

We err on the side of warning over failing in dvclive because the logger is often a pretty minor part of an often expensive, long-running training job, and it can be frustrating to have it break (possibly after hours) because of an error with your logger setup.

> * verifying I'm not missing anything - `dvc repro` is problematic, `dvc exp run` isn't, and running with just Python (directly, via shebang, or some wrapper like [pdm](https://pdm.fming.dev/latest/)) is safe? obvious example: `python script.py`

Correct. With `dvc exp run`, DVC already takes care of saving the experiment and DVCLive can skip these operations. With Python only, DVCLive can save the experiment itself. With `dvc repro`, DVC is already running but not saving the experiment, and we have no way to intervene in the existing DVC process to save it.

omesser:
Got it, thanks for the clarifications @dberenbaum !

PR #702, opened 2023-09-06
live: warn when DVC_ROOT is set but DVC_EXP_BASELINE_REV is not

pmrowla:
  guide.

Will close https://github.com/iterative/dvclive/issues/670 (requires https://github.com/iterative/dvc/pull/9879)

- Generates a warning message when dvclive is init'ed inside a `dvc repro` pipeline stage (but not when inside `dvc exp run`)

With https://github.com/iterative/lstm_seq2seq:
```
$ dvc repro
Running stage 'download':
...
Running stage 'train':
> python train.py
Number of samples: 10000
Number of unique input tokens: 72
Number of unique output tokens: 101
Max sequence length for inputs: 117
Max sequence length for outputs: 149
WARNING:dvclive:Some DVCLive features are unsupported in `dvc repro`.
To use DVCLive with a DVC Pipeline, run it with `dvc exp run`.
GPU available: True (mps), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
...
```

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -186,6 +186,9 @@ class Live:
                 self._save_dvc_exp = False
         elif os.getenv(env.DVC_ROOT, None):
             # `dvc repro` execution
+            if self._save_dvc_exp:
+                logger.info("Ignoring `save_dvc_exp` because `dvc repro` is running")
+                self._save_dvc_exp = False
             logger.warning(
                 "Some DVCLive features are unsupported in `dvc repro`."
                 "\nTo use DVCLive with a DVC Pipeline, run it with `dvc exp run`."
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -135,15 +135,7 @@ class Live:
     def _init_dvc(self):
         from dvc.scm import NoSCM
 
-        if os.getenv(env.DVC_EXP_BASELINE_REV, None):
-            # `dvc exp` execution
-            self._baseline_rev = os.getenv(env.DVC_EXP_BASELINE_REV, "")
-            self._exp_name = os.getenv(env.DVC_EXP_NAME, "")
-            self._inside_dvc_exp = True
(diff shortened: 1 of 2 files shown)


src/dvclive/live.py
```diff
@@ -183,6 +175,22 @@ def _init_dvc(self):
...
+        if os.getenv(env.DVC_EXP_BASELINE_REV, None):
+            # `dvc exp` execution
+            self._baseline_rev = os.getenv(env.DVC_EXP_BASELINE_REV, "")
+            self._exp_name = os.getenv(env.DVC_EXP_NAME, "")
+            self._inside_dvc_exp = True
+            if self._save_dvc_exp:
+                logger.info("Ignoring `save_dvc_exp` because `dvc exp run` is running")
+                self._save_dvc_exp = False
+        elif os.getenv(env.DVC_ROOT, None):
+            # `dvc repro` execution
```
daavoo, line 188: @dberenbaum we would also want to ignore `save_dvc_exp` here, right? Like above:

```py
            if self._save_dvc_exp:
                logger.info("Ignoring `save_dvc_exp` because `dvc exp run` is running")
                self._save_dvc_exp = False
```

dberenbaum: @daavoo Are you aware of any other features that don't work with `dvc repro`? Do we need a warning if `save_dvc_exp=False`?

daavoo (approved):

shcheklein:
I feel we overdid this a bit. I'm not sure why it should be showing a warning if I don't even want to save experiments. Are we completely deprecating `dvc repro` this way? cc @dberenbaum

dberenbaum:
I see I had a [similar concern](https://github.com/iterative/dvclive/pull/702#discussion_r1317790442) back when we introduced it. I don't see why we need two different messages here either.

PR #709, opened 2023-09-12
Studio dvcyaml

dberenbaum:
Make posts to studio drop `dvc.yaml::` from plots

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -137,4 +137,6 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+.dvc/
+.dvcignore
 src/dvclive/_dvclive_version.py
--- a/.dvc/.gitignore
+++ b/.dvc/.gitignore
@@ -1,3 +1,0 @@
-/config.local
-/tmp
-/cache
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -137,4 +137,6 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 
+.dvc/
+.dvcignore
```
dberenbaum, line 141: Unrelated, but I keep getting these generated when running tests

daavoo (approved):
We should wait and bump min DVC, right?

dberenbaum:
Yes, this depends on both https://github.com/iterative/dvc/pull/9898 and https://github.com/iterative/studio/pull/7628. We can either wait to merge it into the 3.0 branch, or merge now and wait for those changes to release 3.0.

dberenbaum:
Actually, we don't technically need a dvc version bump in dvclive, at least for this PR, since this only affects Studio. Probably good to do it anyway to avoid weird plots issues in 3.0, so I added it here.

PR #710, opened 2023-09-14
track dvcyaml file

dberenbaum:
Now that `dvc.yaml` will be written to the root, this PR ensures it is git-tracked.


src/dvclive/live.py
```diff
@@ -193,7 +193,7 @@ def _init_dvc_file(self) -> str:
                 return self._dvcyaml
             raise InvalidDvcyamlError
         if self._dvc_repo is not None:
-            return os.path.join(self._dvc_repo.root_dir, "dvc.yaml")
+            return os.path.relpath(os.path.join(self._dvc_repo.root_dir, "dvc.yaml"))
```
daavoo, line 196: whi is the `relpath` needed?

dberenbaum: Without it, we end up doing `git add /var/tmp/.../dvc.yaml`, which I guess is fine, so I can update the tests to expect that.

src/dvclive/live.py
```diff
@@ -548,6 +548,10 @@ def end(self):
             catch_and_warn(DvcException, logger)(ensure_dir_is_tracked)(
                 self.dir, self._dvc_repo
             )
+            if self._dvcyaml:
+                catch_and_warn(DvcException, logger)(self._dvc_repo.scm.add)(
```
daavoo, line 552: This is missing some test

daavoo (approved):

dberenbaum:
Not sure why the tests are failing. Some error relating to the version.

dberenbaum:
@skshetry WDYT? Should the failure block merging?

src/dvclive/live.py
daavoo: I meant the `scm.add`, it seems that we only test the `include_untracked` change

skshetry:
@dberenbaum, let me try to rerun, it is probably fixed in setuptools_scm. If not, we can merge it and pin it separately. :)

skshetry (approved):

src/dvclive/live.py
dberenbaum: Good catch. Added it to `test_untracked_dvclive_files_inside_dvc_exp_run_are_added`.

PR #715, opened 2023-09-25
log_artifact: Use name validation from GTO.

daavoo:
The function we were using from `dvc.repo.artifacts` was dropped in https://github.com/iterative/dvc/pull/9770

deps: bump  "dvc>=3.22.1"
deps: Add `gto`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    "dvc>=3.22.0",
+    "dvc>3.22.1",
     "dvc-render>=0.6.0,<1.0",
     "dvc-studio-client>=0.15.0,<1",
     "funcy",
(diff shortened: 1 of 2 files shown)


daavoo:
This needs to be merged and released with the next DVC release.
log_artifact is currently broken if used alongside DVC main branch

dberenbaum:
Should we add back an alias for `dvc.repo.artifacts.name_is_compatible` to prevent breaking changes for users?

dberenbaum:
LGTM but will wait until we clean up the DVC side and can pin to an actual DVC release before approving.

daavoo:
> LGTM but will wait until we clean up the DVC side and can pin to an actual DVC release before approving.

I think this P.R. is not needed if we merge https://github.com/iterative/dvc/pull/9972

dberenbaum:
We might still want this PR so we don't have to keep that unnecessary alias in dvc indefinitely, right?

If we do keep this PR, should we explicitly add gto to dependencies?

PR #718, opened 2023-09-28
Use `DVC_ROOT` to detect pipeline

dberenbaum:
Before `DVC_ROOT`, we often checked for whether dvclive was running inside `dvc exp run` because we had no way to check whether running inside `dvc repro`, but often we really want the same behavior in both scenarios. This PR expands the use of `DVC_ROOT` to check whether dvclive is running inside a pipeline.


dberenbaum:
@iterative/dvc Can anyone take a look at this please?

efiop (approved):

PR #725, opened 2023-10-18
custom exp name

### LINKED ISSUE #582: Set exp name
Can we also add exp name? Seems odd to me that we have support for message but not name.

_Originally posted by @dberenbaum in https://github.com/iterative/dvclive/issues/565#issuecomment-1551350667_

dberenbaum:
Closes #582

### DIFF
--- a/src/dvclive/dvc.py
+++ b/src/dvclive/dvc.py
@@ -1,5 +1,6 @@
 # ruff: noqa: SLF001
 import copy
+import logging
 import os
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, List, Optional
@@ -11,6 +12,8 @@ from dvclive.utils import StrPath, rel_path
 if TYPE_CHECKING:
     from dvc.repo import Repo
     from dvc.stage import Stage
+
+logger = logging.getLogger("dvclive")
 
 
 def _dvc_dir(dirname: StrPath) -> str:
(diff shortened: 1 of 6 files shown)


src/dvclive/lightning.py
```diff
@@ -62,6 +62,7 @@ def __init__(  # noqa: PLR0913
         save_dvc_exp: bool = False,
         dvcyaml: bool = True,
         cache_images: bool = False,
+        exp_name: Optional[str] = None,
```
dberenbaum, line 65: Will add `exp_message` and update other args for this callback in a separate PR.

src/dvclive/live.py
```diff
@@ -178,7 +179,10 @@ def _init_dvc(self):
 
         self._baseline_rev = self._dvc_repo.scm.get_rev()
         if self._save_dvc_exp:
-            self._exp_name = get_random_exp_name(self._dvc_repo.scm, self._baseline_rev)
+            if not self._exp_name:
```
shcheklein, line 182: let's check for the name to be valid?

src/dvclive/dvc.py
```diff
@@ -125,12 +128,23 @@ def _update_entries(old, new, key):
...
-    return dvc_get_random_exp_name(scm, baseline_rev)
+def get_exp_name(name, scm, baseline_rev) -> str:
+    from dvc.exceptions import InvalidArgumentError
+    from dvc.repo.experiments.refs import ExpRefInfo
+    from dvc.repo.experiments.utils import check_ref_format, get_random_exp_name
+
+    if name:
+        ref = ExpRefInfo(baseline_sha=baseline_rev, name=name)
+        if scm.get_ref(str(ref)):
+            logger.warning(f"Experiment conflicts with existing experiment '{name}'.")
```
shcheklein, line 139: btw., what happens if two people name the experiment the same way simultaneously (feels that it would have been better to have still some random id / hash in the name or something).

dberenbaum: Nothing prevents us from adding a random suffix, but it's outside the scope of this PR, right? How about we open an issue in dvc to discuss it?

shcheklein: yep, sorry, not a blocker for this PR

shcheklein (approved):

Issue #594, opened 2023-06-07
DVCLiveLogger not compatible with Ray LightningTrainer


schuhschuh:
The `dvclive.lightning.DVCLiveLogger` initializes the `Live` object in it's `__init__` function [since this change](https://github.com/iterative/dvclive/commit/41d12d7936c7a781a30da27e7fd12faa00b34a1e#diff-39138c9a8bea172474941b1d19b40de9c7c2c375392ee0c3522c63bffab201a8R40-R41). This causes a "cannot pickle io.BufferedReader" error when trying to use this logger with the new [LightningTrainer](https://docs.ray.io/en/latest/train/api/doc/ray.train.lightning.LightningTrainer.html) of Ray Train.

It is working with a custom Live logger for PyTorch Lightning I had implemented few months ago, still based on dvclive 1.3 at the time.

Find below a stack trace of the error when adding `logger=DVCLiveLogger()` to `LightningConfigBuilder.trainer()`. This is because Ray needs to transfer the config to the distributed compute node that is executing the Ray Actor which is running the PyTorch Lightning trainer. An alternative might be for Ray's `LightningTrainer` to allow for more customization of the worker training loop, maybe including support for callables to set up loggers.

However, in this instance, it seems it would be easier for DVCLive not to initialize the `Live` object this early.

```
    # 2023-06-07 16:27:37,549 ERROR ray_train.py:164 -- Traceback (most recent call last):
    # File "/opt/conda/envs/mult-ray-train/lib/python3.10/site-packages/ray/_private/worker.py", line 676, in put_object
    #     serialized_value = self.get_serialization_context().serialize(value)
    # File "/opt/conda/envs/mult-ray-train/lib/python3.10/site-packages/ray/_private/serialization.py", line 466, in serialize
    #     return self._serialize_to_msgpack(value)
    # File "/opt/conda/envs/mult-ray-train/lib/python3.10/site-packages/ray/_private/serialization.py", line 444, in _serialize_to_msgpack
    #     pickle5_serialized_object = self._serialize_to_pickle5(
    # File "/opt/conda/envs/mult-ray-train/lib/python3.10/site-packages/ray/_private/serialization.py", line 406, in _serialize_to_pickle5
    #     raise e
    # File "/opt/conda/envs/mult-ray-train/lib/python3.10/site-packages/ray/_private/serialization.py", line 401, in _serialize_to_pickle5
    #     inband = pickle.dumps(
[...]

Issue #178, opened 2021-10-18
DcvLiveLogger error with LightningCLI


alealv:
# Error

```bash
➜ python test.py --config test.yml
usage: test.py [-h] [--config CONFIG] [--print_config [={comments,skip_null}+]] [--seed_everything SEED_EVERYTHING] [--trainer.logger LOGGER] [--trainer.checkpoint_callback {true,false}] [--trainer.callbacks CALLBACKS] [--trainer.default_root_dir DEFAULT_ROOT_DIR]
               [--trainer.gradient_clip_val GRADIENT_CLIP_VAL] [--trainer.gradient_clip_algorithm GRADIENT_CLIP_ALGORITHM] [--trainer.process_position PROCESS_POSITION] [--trainer.num_nodes NUM_NODES] [--trainer.num_processes NUM_PROCESSES] [--trainer.devices DEVICES]
               [--trainer.gpus GPUS] [--trainer.auto_select_gpus {true,false}] [--trainer.tpu_cores TPU_CORES] [--trainer.ipus IPUS] [--trainer.log_gpu_memory LOG_GPU_MEMORY] [--trainer.progress_bar_refresh_rate PROGRESS_BAR_REFRESH_RATE] [--trainer.overfit_batches OVERFIT_BATCHES]
               [--trainer.track_grad_norm TRACK_GRAD_NORM] [--trainer.check_val_every_n_epoch CHECK_VAL_EVERY_N_EPOCH] [--trainer.fast_dev_run FAST_DEV_RUN] [--trainer.accumulate_grad_batches ACCUMULATE_GRAD_BATCHES] [--trainer.max_epochs MAX_EPOCHS]
               [--trainer.min_epochs MIN_EPOCHS] [--trainer.max_steps MAX_STEPS] [--trainer.min_steps MIN_STEPS] [--trainer.max_time MAX_TIME] [--trainer.limit_train_batches LIMIT_TRAIN_BATCHES] [--trainer.limit_val_batches LIMIT_VAL_BATCHES]
               [--trainer.limit_test_batches LIMIT_TEST_BATCHES] [--trainer.limit_predict_batches LIMIT_PREDICT_BATCHES] [--trainer.val_check_interval VAL_CHECK_INTERVAL] [--trainer.flush_logs_every_n_steps FLUSH_LOGS_EVERY_N_STEPS] [--trainer.log_every_n_steps LOG_EVERY_N_STEPS]
[...]

PR #727, opened 2023-10-18
Lightning 3.0 updates

### LINKED ISSUE #726: lightning: args passed to dvclive are outdated
Since https://github.com/iterative/dvclive/pull/181, the lightning callback relies on hard-coded kwargs that get passed to dvclive. These were not updated in 3.0 and updating them has been missed multiple times. This will be even worse if we move this callback into the lightning codebase, so we need to find a better way to keep the lightning callback in sync with dvclive kwargs.

dberenbaum:
Fixes #726 

Update lightning logger kwargs to reflect dvclive 3.0 updates and other missed changes

### DIFF
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -50,36 +50,19 @@ def _should_call_next_step():
 
 
 class DVCLiveLogger(Logger):
-    def __init__(  # noqa: PLR0913
+    def __init__(
         self,
         run_name: Optional[str] = "dvclive_run",
         prefix="",
         log_model: Union[str, bool] = False,
         experiment=None,
-        dir: str = "dvclive",  # noqa: A002
-        resume: bool = False,
-        report: Optional[str] = None,
-        save_dvc_exp: bool = True,
-        dvcyaml: Union[str, bool] = True,
-        cache_images: bool = False,
-        exp_message: Optional[str] = None,
+        **kwargs,
     ):
         super().__init__()
         self._prefix = prefix
-        self._live_init: Dict[str, Any] = {
-            "resume": resume,
-            "report": report,
-            "save_dvc_exp": save_dvc_exp,
-            "dvcyaml": dvcyaml,
-            "cache_images": cache_images,
-        }
-        if dir is not None:
-            self._live_init["dir"] = dir
+        self._live_init: Dict[str, Any] = kwargs
         self._experiment = experiment
         self._version = run_name
-        if report == "notebook":
-            # Force Live instantiation
-            self.experiment  # noqa: B018
         self._log_model = log_model
         self._logged_model_time: Dict[str, float] = {}
         self._checkpoint_callback: Optional[ModelCheckpoint] = None
--- a/src/dvclive/lightning.py
+++ b/src/dvclive/lightning.py
@@ -56,12 +56,13 @@ class DVCLiveLogger(Logger):
         prefix="",
         log_model: Union[str, bool] = False,
         experiment=None,
-        dir: Optional[str] = None,  # noqa: A002
+        dir: str = "dvclive",  # noqa: A002
         resume: bool = False,
         report: Optional[str] = None,
-        save_dvc_exp: bool = False,
-        dvcyaml: bool = True,
+        save_dvc_exp: bool = True,
+        dvcyaml: Union[str, bool] = True,
         cache_images: bool = False,
+        exp_message: Optional[str] = None,
     ):
         super().__init__()
         self._prefix = prefix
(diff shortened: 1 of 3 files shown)


src/dvclive/lightning.py
```diff
@@ -50,35 +50,19 @@ def _should_call_next_step():
...
         prefix="",
         log_model: Union[str, bool] = False,
         experiment=None,
-        dir: Optional[str] = None,  # noqa: A002
-        resume: bool = False,
-        report: Optional[str] = None,
-        save_dvc_exp: bool = False,
-        dvcyaml: bool = True,
-        cache_images: bool = False,
+        **kwargs,
```
dberenbaum, line 59: Explicit kwargs initially came from https://github.com/iterative/dvclive/issues/178. This is no longer an issue in newer versions of lightning, and there are CLI tests already in lightning that I added below.

src/dvclive/lightning.py
```diff
@@ -50,35 +50,19 @@ def _should_call_next_step():
...
-            "cache_images": cache_images,
-        }
-        if dir is not None:
-            self._live_init["dir"] = dir
+        self._live_init: Dict[str, Any] = kwargs
         self._experiment = experiment
         self._version = run_name
-        if report == "notebook":
-            # Force Live instantiation
-            self.experiment  # noqa: B018
```
dberenbaum, line 81: This was added in https://github.com/iterative/dvclive/pull/432 but had to be updated after https://github.com/iterative/dvclive/issues/594. I don't think it's worthwhile since this still could cause issues in distributed compute. The only thing it does AFAICT is make sure the dvclive notebook report gets output at the very top of the cell output. 

@daavoo If you see this and have any more context, please chime in.

Example of output with this notebook-specific logic:

Screenshot 2023-10-19 at 3 07 31 PM

Example without:

Screenshot 2023-10-19 at 3 11 02 PM

tests/frameworks/test_lightning.py
```diff
@@ -277,22 +283,62 @@ def test_lightning_val_udpates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio
...
+
+    out = StringIO()
+    with mock.patch(
+        "sys.argv", ["any.py"] + cli_args  # noqa: RUF005
+    ), redirect_stdout(  # noqa: RUF100
+        out
+    ), pytest.raises(
+        SystemExit
+    ):
+        LightningCLI(TestModel, run=False)
```
dberenbaum, line 329: Once we move this whole integration to lightning, we won't have to copy all this from lightning tests.

dberenbaum:
Tests are failing because of an issue with pip cache? Don't know if it's related to the PR.

shcheklein:
`Received 88080384 of 3940402683 (2.2%), 41.9 MBs/sec` - that's what, 3-4TB? :) What is happening here? It was running for 29 minutes ...

shcheklein:
I've cleaned up the cache in Actions -> Cached (there was one for 3.7GB !) and rerun it.

src/dvclive/lightning.py
shcheklein: Just to double check - will it break some existing Colabs / notebook examples?

src/dvclive/lightning.py
```diff
@@ -50,36 +50,19 @@ def _should_call_next_step():
...
-        self._live_init: Dict[str, Any] = {
-            "resume": resume,
-            "report": report,
-            "save_dvc_exp": save_dvc_exp,
-            "dvcyaml": dvcyaml,
-            "cache_images": cache_images,
-        }
-        if dir is not None:
-            self._live_init["dir"] = dir
+        self._live_init: Dict[str, Any] = kwargs
```
shcheklein, line 63: double check:

- default values in kwargs, e.g. will we still consider `dvcyaml` as True if it's not specified in kwargs
- might worth creating copy of kwargs, so that we 100% that's it's frozen for the external world

shcheklein (commented):
Thanks, @dberenbaum ! I've reviewed it quickly.

src/dvclive/lightning.py
dberenbaum: I tested it with the lightning colab example we have, which is what's shown above, so AFAIK it shouldn't break anything.

src/dvclive/lightning.py
dberenbaum: Added tests for both of those cases. With the test, I didn't think it's necessary to make a copy of kwargs, and I don't see it happening in other loggers like https://github.com/Lightning-AI/lightning/blob/master/src/lightning/pytorch/loggers/comet.py, but not that strong an opinion if you really feel it's necessary.

shcheklein: yeah, I think we overdid this :) kwargs creates a new dict object when it's being passed I think we don't need a test (no need to test Python I guess :) ). I was asking primarily to check that tbh

same with default values, if it's Python behavior - no need to test this

shcheklein (approved):

PR #729, opened 2023-10-30
dvcyaml: write to cwd instead of git root

dberenbaum:
Instead of defaulting to the git root for dvc.yaml, use the current path. This is technically a breaking change, but it simplifies the dvc.yaml logic and should have the same behavior for most users.

Rationale for this change:
* Avoids unnecessary git logic that can cause [unexpected behavior](https://iterativeai.slack.com/archives/C03JS2V4MQU/p1698052870762279)
* More consistent with dvc commands (for example, `dvc stage add` writes `dvc.yaml` in the current directory, not the git root)
* Simpler code

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -63,7 +63,7 @@ class Live:
         resume: bool = False,
         report: Optional[str] = None,
         save_dvc_exp: bool = True,
-        dvcyaml: Union[str, bool] = True,
+        dvcyaml: Union[str, bool] = "dvc.yaml",
         cache_images: bool = False,
         exp_name: Optional[str] = None,
         exp_message: Optional[str] = None,
@@ -109,9 +109,6 @@ class Live:
         self._dvc_studio_config: Dict[str, Any] = {}
         self._init_studio()
 
-        # Fail fast for invalid dvc.yaml path.
-        self.dvc_file  # noqa: B018
-
     def _init_resume(self):
         self._read_params()
         self._step = self.read_step()
@@ -150,6 +147,8 @@ class Live:
         dvc_logger = logging.getLogger("dvc")
         dvc_logger.setLevel(os.getenv(env.DVCLIVE_LOGLEVEL, "WARNING").upper())
 
+        self._dvc_file = self._init_dvc_file()
+
         if (self._dvc_repo is None) or isinstance(self._dvc_repo.scm, NoSCM):
             if self._save_dvc_exp:
                 logger.warning(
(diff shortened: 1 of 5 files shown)


src/dvclive/live.py
```diff
@@ -63,7 +63,7 @@ def __init__(
         resume: bool = False,
         report: Optional[str] = None,
         save_dvc_exp: bool = True,
-        dvcyaml: Union[str, bool] = True,
+        dvcyaml: Union[str, bool] = "dvc.yaml",
```
dberenbaum, line 66: Now that there is a constant value, it seems more transparent to make the default value a string. Keeping a boolean option for now to not break existing behavior.

shcheklein:
does it meant that depending on where I run the command (e.g. yolo train ...) I might end up with a few `dvc.yaml` files? Also, how do we determine the path for the `dvclive` directory now? It is also usually in the root, how do we do that?

dberenbaum:
> does it meant that depending on where I run the command (e.g. yolo train ...) I might end up with a few `dvc.yaml` files?

Yes, it's possible if you run it from different places. I can think of cases where each behavior is "better." For example, if I have multiple pipelines, always writing to the root could be worse.

> Also, how do we determine the path for the `dvclive` directory now? It is also usually in the root, how do we do that?

It is based on the current path, not the root.

shcheklein (approved):

src/dvclive/live.py
```diff
@@ -191,13 +191,7 @@ def _init_dvc_file(self) -> str:
             if os.path.basename(self._dvcyaml) == "dvc.yaml":
                 return self._dvcyaml
             raise InvalidDvcyamlError
-        if self._dvc_repo is not None:
-            return os.path.join(self._dvc_repo.root_dir, "dvc.yaml")
-        logger.warning(
-            "Can't infer dvcyaml path without a DVC repo. "
-            "`dvc.yaml` file will not be written."
-        )
-        return ""
```
skshetry, line 200: >  dvcyaml: write to cwd instead of git root 

Looking at the diff, it seems it was dvc's root, not git root, right?

skshetry:
@dberenbaum, also does this need to be updated as well?

https://github.com/iterative/dvclive/blob/0df9dc5b92cd909900d6302df26afe96cf87deb7/src/dvclive/live.py#L170

skshetry (approved):

src/dvclive/live.py
dberenbaum: Yes, you are right

dberenbaum:
> `dvc stage add` is a CLI command, which comes with the expectation that it respects your current working directory (and it's a one-off command). I don't understand changes here yet, but in the scripts, you might invoke it from any location, not just your cwd. Shouldn't dvclive be consistent/predictable in that regard? No strong opinion though.

Yes, sorry, the description is a bit sloppy. This does not guarantee it uses your cwd. As discussed above with @shcheklein, this PR matches the behavior for creating the dvclive directory, so it seems most consistent with existing dvclive behavior (the dvc.yaml and dvclive directory should be in the same place).

BastiQ:
My solution and online course are built on top of DVC and DVC Live. I really appreciate your work and understand that you have to continue developing the tool, but breaking changes like this result in a lot of overhead on my end. Maybe there is a way to reduce those in the future. Thanks a lot!

dberenbaum:
Sorry for the incovenience @BastiQ. Obviously, this isn't best practice and we do our best to avoid breaking changes in minor releases. If this or other changes are causing specific issues for you, feel free to follow up with more details.

PR #732, opened 2023-11-03
adds hf accelerate tracker

### LINKED ISSUE #731: integrations: hf accelerate
DVCLive already supports `transformers`, but we should add support for [accelerate](https://github.com/huggingface/accelerate/blob/main/src/accelerate/tracking.py).

dberenbaum:
Fixes #731 

TODO:
- [x] Tests
- [ ] Docs


dberenbaum:
There

PR #735, opened 2023-11-10
add test mode

dberenbaum:
Related to test failures in https://github.com/huggingface/accelerate/pull/2139. DVCLive automatically depends a lot on the Git and DVC repos, so this PR adds a test mode enabled by the environment variable `DVCLIVE_TEST` to disable all of that functionality and only perform the dvclive operations that aren't repo-specific. This could also probably simplify internal tests, although I haven't gone through to check in this PR.


dberenbaum:
@iterative/dvc Any objections to merging this? It's blocking an external PR in https://github.com/huggingface/accelerate/pull/2139.

efiop (approved):

PR #738, opened 2023-11-14
post data to studio on end

dberenbaum:
Before we only post the `done` event during `live.end()`, which doesn't include any data. We should also be posting a `data` event before that to make sure the latest data is also sent.

### DIFF
--- a/tests/test_post_to_studio.py
+++ b/tests/test_post_to_studio.py
@@ -132,15 +132,24 @@ def test_post_to_studio_failed_start_request(
     assert mocked_post.call_count == 1
 
 
-def test_post_to_studio_end_only_once(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+def test_post_to_studio_done_only_once(tmp_dir, mocked_dvc_repo, mocked_studio_post):
     mocked_post, _ = mocked_studio_post
     with Live() as live:
         live.log_metric("foo", 1)
         live.next_step()
 
-    assert mocked_post.call_count == 4
+    expected_done_calls = [
+        call
+        for call in mocked_post.call_args_list
+        if call.kwargs["json"]["type"] == "done"
+    ]
     live.end()
-    assert mocked_post.call_count == 4
+    actual_done_calls = [
+        call
+        for call in mocked_post.call_args_list
+        if call.kwargs["json"]["type"] == "done"
+    ]
+    assert expected_done_calls == actual_done_calls
 
 
 @pytest.mark.studio()
@@ -157,7 +166,9 @@ def test_post_to_studio_skip_start_and_done_on_env_var(
         live.log_metric("foo", 1)
         live.next_step()
 
-    assert mocked_post.call_count == 2
+    call_types = [call.kwargs["json"]["type"] for call in mocked_post.call_args_list]
+    assert "start" not in call_types
+    assert "done" not in call_types
 
 
 @pytest.mark.studio()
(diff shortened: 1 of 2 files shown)


tests/test_post_to_studio.py
```diff
@@ -132,15 +132,24 @@ def test_post_to_studio_failed_start_request(
     assert mocked_post.call_count == 1
 
 
-def test_post_to_studio_end_only_once(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+def test_post_to_studio_done_only_once(tmp_dir, mocked_dvc_repo, mocked_studio_post):
```
dberenbaum, line 135: End may be called multiple times but the `done` event should only be called once.

tests/test_post_to_studio.py
```diff
@@ -132,15 +132,24 @@ def test_post_to_studio_failed_start_request(
...
+        if call.kwargs["json"]["type"] == "done"
+    ]
     live.end()
-    assert mocked_post.call_count == 4
+    actual_done_calls = [
+        call
+        for call in mocked_post.call_args_list
+        if call.kwargs["json"]["type"] == "done"
+    ]
+    assert expected_done_calls == actual_done_calls
```
dberenbaum, line 152: Better to test that this doesn't call `done` again rather than test call counts.

tests/test_post_to_studio.py
```diff
@@ -157,7 +166,9 @@ def test_post_to_studio_skip_start_and_done_on_env_var(
         live.log_metric("foo", 1)
         live.next_step()
 
-    assert mocked_post.call_count == 2
+    call_types = [call.kwargs["json"]["type"] for call in mocked_post.call_args_list]
+    assert "start" not in call_types
+    assert "done" not in call_types
```
dberenbaum, line 171: Better to test that `start` and `done` events aren't called rather than call counts.

tests/test_post_to_studio.py
```diff
@@ -236,7 +247,9 @@ def test_post_to_studio_inside_dvc_exp(
...
```
dberenbaum, line 252: Same as above

tests/test_post_to_studio.py
```diff
@@ -370,3 +383,15 @@ def test_post_to_studio_name(tmp_dir, mocked_dvc_repo, mocked_studio_post):
...
+def test_post_to_studio_if_done_skipped(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+    live = Live()
+    live._studio_events_to_skip.add("start")
+    live._studio_events_to_skip.add("done")
+    live.log_metric("foo", 1)
+    live.end()
+
+    mocked_post, _ = mocked_studio_post
+    call_types = [call.kwargs["json"]["type"] for call in mocked_post.call_args_list]
+    assert "data" in call_types
```
dberenbaum, line 397: Added test to check the use case from [slack](https://iterativeai.slack.com/archives/CUSNDR35K/p1697630212171659), where `live.next_step()` is never called and `done` events are skipped.

tests/test_post_to_studio.py
```diff
@@ -169,14 +180,15 @@ def test_post_to_studio_dvc_studio_config(
...
+    monkeypatch.delenv(DVC_STUDIO_TOKEN)
 
     mocked_dvc_repo.config = {"studio": {"token": "token"}}
 
     with Live() as live:
         live.log_metric("foo", 1)
         live.next_step()
 
-    assert mocked_post.call_count == 2
+    assert mocked_post.call_args.kwargs["headers"]["Authorization"] == "token token"
```
dberenbaum, line 191: This test wasn't actually working as intended since the env var was overriding the dvc config. Now it actually checks the token value to make sure it uses the dvc config value.

shcheklein (approved):

PR #744, opened 2023-11-30
fix: Improve safety of logging images with implicit formats

### LINKED ISSUE #743: `log_image` fails to save an image if the path was provided, but the `name` does not have an extension
In an example like this: 

```python
import time
import random

from dvclive import Live

params = {"learning_rate": 0.002, "optimizer": "Adam", "epochs": 20}

with Live() as live:
    # log a parameters
    for param in params:
        live.log_param(param, params[param])

    # simulate training
    offset = random.uniform(0.02, 0.1)
    for epoch in range(1, params["epochs"]):
        fuzz = random.uniform(0.01, 0.1)
        accuracy = 1 - (2**-epoch) - fuzz - offset
        loss = (2**-epoch) + fuzz + offset

        # ... 

        # here's the critical piece: 
        live.log_image("image", "/Users/nategadzhi/Downloads/Untitled.jpg")
        live.next_step()
        time.sleep(0.2)
```

when [`live.log_image`](https://github.com/natikgadzhi/dvclive-scratchpad/blob/main/.venv/lib/python3.11/site-packages/dvclive/live.py#L372) is trying to save an image, we provide:
- `name` = `"image"`
- `val` = `imagine/a/valid/path`

1. DVCLive will try and open the image with PIL
  - Related documentation fix: https://github.com/iterative/dvc.org/pull/5007
3. DVCLive then makes `dvc_live.Image` object with the provided `name`. **Notice that the name is not a valid filename, it does not have an extension.**
4. Then, `Image` will try to `dump` itself, but it's `output_path` does not have an extension either, because it's basically `output_dir/subdir/name`. 
5. PIL fails to save the image with `ValueError: unknown file extension: `

### Document and provide a better error
We can document this behavior, make it obvious that `name` will become the filename of the saved image, and that it must be a valid filename with an extension. And we can fail earlier with a nicer error message. 

### Try to infer the extension from the image
Alternatively, we can infer the format of the image from the image itself. If the `name` does not have an extension, we can apply `f"{name}/{val.format}"`. 

I _think_ inferring the type from the image file is a better idea, but I'm not an expert in ML context specifically, and how folks are used to working with their data. 

Happy to help with a fix if you point me in the right direction! /cc @dberenbaum @shcheklein

natikgadzhi:
## Summary

This pull request makes it so:
- If `live.log_image` is called without image format in the `name`, we will try to infer the format from the image
- If we could not infer the format, we will raise an error message early in `could_log` instead of falling through to `PIL.Image.save`

Closes #743 

## Changes

@dberenbaum, I felt that both approaches need to be implemented together. Inferring the format is nice, but only works in _some_ situations and not others. So for a situation where the format could not be inferred, I added an additional check that will fail early and won't attempt saving the image.

My Python must be _not optimal_. The changes are split into two commits, feature-wise. 
- I've also added unit tests for the behavior.
- Behavior change: in `could_log` I suggest we take `name` and `val`, not just val. This might be a bad idea if it's used elsewhere (not in this repository, I checked). It's not in public API, though. 

I'm not strongly attached to the `could_log` idea. In the end of the day, it's only saving us a bit of time, and clarifies the error, but it's not strictly required.

## TODO
  guide.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -379,12 +379,12 @@ class Live:
 
         # If the provided image name does not have a format on it,
         # try to infer the format from PIL Image.
-        if len(name.split(".")) <= 1:
-            if (
-                isinstance_without_import(val, "PIL.Image", "Image")
-                and f".{str(val.format).lower()}" in Image.suffixes
-            ):
-                name = f"{name}.{str(val.format).lower()}"
+        if (
+            len(name.split(".")) <= 1
+            and isinstance_without_import(val, "PIL.Image", "Image")
+            and f".{str(val.format).lower()}" in Image.suffixes
+        ):
+            name = f"{name}.{str(val.format).lower()}"
 
         # See if the image format and image name are valid
         if not Image.could_log(name, val):
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -371,9 +371,7 @@ class Live:
         logger.debug(f"Logged {name}: {val}")
 
     def log_image(self, name: str, val):
-        if not Image.could_log(val):
-            raise InvalidDataTypeError(name, type(val))
-
+        # If we're given a path, try loading the image first. This might error out.
         if isinstance(val, (str, Path)):
             from PIL import Image as ImagePIL
 
@@ -387,6 +385,10 @@ class Live:
                 and f".{str(val.format).lower()}" in Image.suffixes
             ):
                 name = f"{name}.{str(val.format).lower()}"
+
+        # See if the image format and image name are valid
+        if not Image.could_log(name, val):
+            raise InvalidDataTypeError(name, type(val))
 
         if name in self._images:
             image = self._images[name]
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -28,6 +28,7 @@ from .dvc import (
 from .error import (
     InvalidDataTypeError,
     InvalidDvcyamlError,
+    InvalidImageNameError,
     InvalidParameterTypeError,
     InvalidPlotTypeError,
     InvalidReportModeError,
@@ -43,7 +44,6 @@ from .utils import (
     convert_datapoints_to_list_of_dicts,
     env2bool,
     inside_notebook,
-    isinstance_without_import,
     matplotlib_installed,
     open_file_in_browser,
 )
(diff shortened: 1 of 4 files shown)


natikgadzhi:
I wonder if that's worth documenting? Sort of a minor change, but happy to put a PR into the docs site, too.

natikgadzhi:
Also, happy to write up the python docs in all the live.* functions if that's something that you think is useful.

src/dvclive/live.py
```diff
@@ -370,14 +371,25 @@ def log_metric(
         logger.debug(f"Logged {name}: {val}")
 
     def log_image(self, name: str, val):
-        if not Image.could_log(val):
-            raise InvalidDataTypeError(name, type(val))
```
dberenbaum, line 374: WDYT about this instead?
1. Leaving `could_log` as is (keep it here and don't change the method)
2. Where you moved `could_log` below, check the name suffix(`if not Path(name).suffix in Image.suffixes`)

It's less changes to the existing codebase, and I think it would be better to have separate errors for invalid image input values and invalid output names.

src/dvclive/live.py
```diff
@@ -370,14 +371,25 @@ def log_metric(
...
             val = ImagePIL.open(val)
 
+        # If the provided image name does not have a format on it,
+        # try to infer the format from PIL Image.
+        if (
+            len(name.split(".")) <= 1
+            and isinstance_without_import(val, "PIL.Image", "Image")
+            and f".{str(val.format).lower()}" in Image.suffixes
+        ):
+            name = f"{name}.{str(val.format).lower()}"
```
dberenbaum, line 387: As mentioned above, I think we could drop this since I don't think it works for anything but existing image paths.

tests/plots/test_image.py
```diff
@@ -24,10 +25,33 @@ def test_pil(tmp_dir):
     assert (tmp_dir / live.plots_dir / LiveImage.subfolder / "image.png").exists()
 
 
+def test_pil_omitting_extension_doesnt_save_without_valid_format(tmp_dir):
+    live = Live()
+    img = Image.new("RGB", (10, 10), (250, 250, 250))
+    with pytest.raises(InvalidDataTypeError, match="has not supported type"):
```
dberenbaum, line 31: As mentioned above, would be better to have a separate error here for invalid/missing output extension.

AlexandreKempf (approved):

dberenbaum (approved):

PR #749, opened 2023-12-07
Adds lightning fabric integration

dberenbaum:
Closes #742. Fabric is Pytorch Lightning's attempt to abstract out the lightweight parts of their codebase so users can benefit from those without having to entirely adapt Lightning.

- Example notebook is included: https://github.com/iterative/dvclive/blob/38d1f40843be5adc49ed718884f04496a7f376ff/examples/DVCLive-Fabric.ipynb
- Docs PR: https://github.com/iterative/dvc.org/pull/5039

The existing Lightning logger was updated to inherit from this one. To make everything work with the existing Lightning logger, I added a `sync` method that decouples `next_step` from saving all the info to Studio, dvc.yaml, etc. I made it public, but no strong opinion on whether it should be public or documented.

See the Tensorboard logger in lightning for an example of another logger:
- [Docs](https://lightning.ai/docs/fabric/stable/api/generated/lightning.fabric.loggers.TensorBoardLogger.html#lightning.fabric.loggers.TensorBoardLogger)
- [Fabric logger](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/fabric/loggers/tensorboard.py)
- [Lightning logger](https://github.com/Lightning-AI/pytorch-lightning/blob/master/src/lightning/pytorch/loggers/tensorboard.py)

PR #752, opened 2023-12-09
fix(pyproject): proper reference to the lighting extra

shcheklein:
Found a small issue while reviewing https://github.com/iterative/dvclive/pull/749

------------------

- [X] ❗ I have followed the [Contributing to DVCLive](https://github.com/iterative/dvclive/blob/main/CONTRIBUTING.rst)
  guide.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ fastai = ["fastai"]
 lightning = ["lightning>=2.0", "torch"]
 optuna = ["optuna"]
 all = [
-   "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,pytorch-lightning,optuna,plots,markdown]"
+   "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,lightning,optuna,plots,markdown]"
 ]
 
 [project.urls]


shcheklein:
Actually, I think it's better to be fixed right in the #749 since we are changing the lighting module there and here CI fails with mypy on it.

dberenbaum:
@shcheklein What issue did you git using `pytorch-lightning`? The codebase tries to be compatible with both `lightning` and `pytorch-lightning`, so wondering if there is something we need to address here.

shcheklein:
I think the `dvclive[all]` just doesn't install it. Since there is not `pytorch-lightning` anymore AFAIR.

dberenbaum:
Okay, I see. It looks like they both still exist but `lightning` is a superset that includes both `pytorch-lightning` and `fabric`. See https://github.com/Lightning-AI/pytorch-lightning/discussions/16688#discussioncomment-6750981.

shcheklein:
To clarify, I think it us:

```
tf = ["tensorflow"]
xgb = ["xgboost"]
lgbm = ["lightgbm"]
huggingface = ["transformers", "datasets"]
catalyst = ["catalyst>22"]
fastai = ["fastai"]
lightning = ["lightning>=2.0", "torch"]
optuna = ["optuna"]
all = [
   "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,lightning,optuna,plots,markdown]"
]
```

mind the `lightning` and it's definition. `all` was referring to non-existent extra in _our_ config ASFAI.

dberenbaum:
🤦 Got it, thanks for the clarification

Issue #750, opened 2023-12-07
Add user friendly warning/error messages and helpers for log_plot() 


mnrozhkov:
When people start using `Live.log_plot()`, they could struggle with getting an expected visualization because of 2 reasons 
1. `log_plot()` is very opinionated about the data format required for every template
2. there are not user-friendly data checks and warning messages 

Here are some ideas to help with DVCLive onboarding: 

**1. "Relax" requirements for data formats supported** 

For example, the `bar_horizontal` template expects smth like this: 
```python
datapoints = [
    {"name": "petal_width", "importance": 0.4},
    {"name": "petal_length", "importance": 0.33},
    {"name": "sepal_width", "importance": 0.24},
    {"name": "sepal_length", "importance": 0.03}
]
``` 

It would be cool to support other formats like: 
1) Pandas DataFrame 
<img width="230" alt="image" src="https://github.com/iterative/dvclive/assets/11489539/df6e663e-7b6c-4a20-a921-95174dac1dd7">

2) Dict  with automatically extracts keys as `y' and values as `x.` 
```python
{'petal_width': 0.4,
 'petal_length': 0.33,
 'sepal_width': 0.24,
 'sepal_length': 0.03}
 ``` 
 
 **2. Provide minimal sanity checks for data/configs provides** 
 For example, if I run this code snippet: 
 ```python
 from dvclive import Live

datapoints = [
    {"name": "petal_width", "importance": 0.4},
    {"name": "petal_length", "importance": 0.33},
    {"name": "sepal_width", "importance": 0.24},
    {"name": "sepal_length", "importance": 0.03}
]

with Live() as live:
    live.log_plot(
        "iris_feature_importance",
        datapoints,
        x="name",
        y="importance",
        template="bar_horizontal",
        title="Iris Dataset: Feature Importance",
        y_label="Feature Name",
        x_label="Feature Importance"
    )
``` 

I'll not get any error, but there is nothing showing in VSCode after that: 
<img width="623" alt="image" src="https://github.com/iterative/dvclive/assets/11489539/9b27a76f-a298-4171-8d62-5953b7adae10">

Reason? There is a mistake in `x` and `y` arguments assignment, the correct is `y="name", x="importance"`. But, it's very easy to oversee this typo and spend a lot of time trying to figure it out. 

How can we help? 
- check that the `bar_horizontal` template expects numerical data for `x` 
 
 **3. Provide good warning messages and hints if formats incompatible** 
[...]

PR #754, opened 2023-12-12
Add support for DataFrames and Numpy Arrays for log_plot()

mnrozhkov:
This PR is a part of this issue: https://github.com/iterative/dvclive/issues/750 

Changes: 
- add support for Add support for DataFrames and Numpy Arrays in `log_plot()` 
- implemented a new function for convert datapoints to List[Dict] `dvclive/utils.py/convert_datapoints_to_list_of_dicts` 
- added new Python dependencies: `pandas` and `numpy`
- added a new argument to the `log_plot()`: `columns: Optional[List[str]] = None,` - optional list of column names for data in Numpy Array format

JN to test new feature: [log_plots.ipynb](https://drive.google.com/file/d/1tDoigZrpo_YzPiS-fXsrDHSJuSmgM-uT/view?usp=sharing)

### TODO
  guide.
- [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR: https://github.com/iterative/dvc.org/pull/5055

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -2,7 +2,9 @@ import glob
 import json
 import logging
 import math
+import numpy as np
 import os
+import pandas as pd
 import shutil
 import tempfile
 from pathlib import Path
@@ -35,6 +37,7 @@ from .utils import (
     StrPath,
     catch_and_warn,
     clean_and_copy_into,
+    convert_datapoints_to_list_of_dicts,
     env2bool,
     inside_notebook,
     matplotlib_installed,
@@ -391,14 +394,19 @@ class Live:
     def log_plot(
         self,
         name: str,
-        datapoints: List[Dict],
+        datapoints: pd.DataFrame | np.ndarray | List[Dict],
         x: str,
         y: str,
         template: Optional[str] = None,
         title: Optional[str] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
+        columns: Optional[List[str]] = None,
     ):
+        
+        # Convert the given datapoints to List[Dict]
+        datapoints = convert_datapoints_to_list_of_dicts(datapoints=datapoints, columns=columns)
+
         if not CustomPlot.could_log(datapoints):
             raise InvalidDataTypeError(name, type(datapoints))
 
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -403,9 +403,10 @@ class Live:
         y_label: Optional[str] = None,
         columns: Optional[List[str]] = None,
     ):
-        
         # Convert the given datapoints to List[Dict]
-        datapoints = convert_datapoints_to_list_of_dicts(datapoints=datapoints, columns=columns)
+        datapoints = convert_datapoints_to_list_of_dicts(
+            datapoints=datapoints, columns=columns
+        )
 
         if not CustomPlot.could_log(datapoints):
             raise InvalidDataTypeError(name, type(datapoints))
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -1,14 +1,17 @@
+from __future__ import annotations
 import glob
 import json
 import logging
 import math
-import numpy as np
 import os
-import pandas as pd
 import shutil
 import tempfile
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Union
+from typing import Any, Dict, List, Optional, Set, Union, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import numpy as np
+    import pandas as pd
 
 from dvc.exceptions import DvcException
 from funcy import set_in
@@ -401,12 +404,9 @@ class Live:
         title: Optional[str] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
-        columns: Optional[List[str]] = None,
     ):
         # Convert the given datapoints to List[Dict]
-        datapoints = convert_datapoints_to_list_of_dicts(
-            datapoints=datapoints, columns=columns
-        )
+        datapoints = convert_datapoints_to_list_of_dicts(datapoints=datapoints)
 
         if not CustomPlot.could_log(datapoints):
             raise InvalidDataTypeError(name, type(datapoints))
--- a/src/dvclive/utils.py
+++ b/src/dvclive/utils.py
@@ -1,12 +1,14 @@
 import csv
 import json
+import numpy as np
 import os
+import pandas as pd
 import re
 import shutil
-import webbrowser
 from pathlib import Path
 from platform import uname
-from typing import Union
+from typing import Union, List, Dict, Optional
+import webbrowser
 
 StrPath = Union[str, Path]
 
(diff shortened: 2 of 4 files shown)


src/dvclive/live.py
```diff
@@ -2,7 +2,9 @@
 import json
 import logging
 import math
+import numpy as np
 import os
+import pandas as pd
```
dberenbaum, line 7: Can we make these optional dependencies and add them here?

https://github.com/iterative/dvclive/blob/84069208144be711ef0feccc38064652d732eac7/pyproject.toml#L46

You can still handle type checking here without always importing those deps: https://stackoverflow.com/questions/61384752/how-to-type-hint-with-an-optional-import

src/dvclive/live.py
```diff
@@ -391,14 +394,20 @@ def log_image(self, name: str, val):
...
         template: Optional[str] = None,
         title: Optional[str] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
+        columns: Optional[List[str]] = None,
     ):
+        # Convert the given datapoints to List[Dict]
+        datapoints = convert_datapoints_to_list_of_dicts(
+            datapoints=datapoints, columns=columns
+        )
```
dberenbaum, line 409: Can we check here if it's a list first and skip in that case? That way, you can still plot a list without needing pandas and numpy installed.

dberenbaum: Or you can leave this and check it within that function itself. For example:

https://github.com/iterative/dvclive/blob/84069208144be711ef0feccc38064652d732eac7/src/dvclive/live.py#L372-L389

src/dvclive/utils.py
```diff
@@ -1,12 +1,14 @@
 import csv
 import json
+import numpy as np
 import os
+import pandas as pd
```
dberenbaum, line 5: Let's import these only as needed inside the function itself. See the example above from `log_image()`.

src/dvclive/live.py
```diff
@@ -391,14 +394,20 @@ def log_image(self, name: str, val):
...
         name: str,
-        datapoints: List[Dict],
+        datapoints: pd.DataFrame | np.ndarray | List[Dict],
         x: str,
         y: str,
         template: Optional[str] = None,
         title: Optional[str] = None,
         x_label: Optional[str] = None,
         y_label: Optional[str] = None,
+        columns: Optional[List[str]] = None,
```
dberenbaum, line 404: I'm not sure we need this. It's only useful for the specific scenario of labeling numpy array columns AFAICT, and there are already several ways to handle that which are at least as straightforward IMO:

1. Convert to a pd dataframe yourself
2. Use [structured arrays](https://numpy.org/doc/stable/user/basics.rec.html)

dberenbaum (changes requested):
Thanks @mnrozhkov! Very clean and simple. I have just two overall suggestions:

1. Try to require numpy and pandas as narrowly as possible
2. Drop `columns` since I think users can achieve the same without it

src/dvclive/live.py
mnrozhkov: Ok, it makes sense!  Should we raise an error if none of the supported formats `List[Dict] |pd.DataFrame | np.ndarray`  provided?

dberenbaum: I guess an error will already get raised below:

https://github.com/iterative/dvclive/blob/19b71b4c05c0e24f747ee4df4e331f960a4dbd97/src/dvclive/live.py#L402

However, I don't think it's a bad idea to check the types here and raise if it's not one of those. Not a blocker to me however you want to handle it.

mnrozhkov:
> Thanks @mnrozhkov! Very clean and simple. I have just two overall suggestions:
> 
> 1. Try to require numpy and pandas as narrowly as possible
> 2. Drop `columns` since I think users can achieve the same without it

Thanks @dberenbaum! I resolved all comments. Pls, let me know if there are other places to improve 🙌

src/dvclive/utils.py
```diff
@@ -194,3 +199,39 @@ def read_history(live, metric):
...
+    Raises:
+        ValueError: If the `datapoints` argument is not of type pd.DataFrame, np.ndarray, or List[Dict].
+    """
+    if isinstance(datapoints, list):
+        return datapoints
+
+    import pandas as pd
+
+    if isinstance(datapoints, pd.DataFrame):
+        return datapoints.to_dict(orient="records")
```
dberenbaum, line 225: I think you could use this:

https://github.com/iterative/dvclive/blob/19b71b4c05c0e24f747ee4df4e331f960a4dbd97/src/dvclive/utils.py#L153-L157

That way you can check instance types without importing the packages. Hopefully that means we would only import pandas when passed a pandas dataframe.

dberenbaum:
Thanks @mnrozhkov! Left one last comment and then I think it's ready to merge. Do you plan to also do the docs PR?

mnrozhkov:
> Thanks @mnrozhkov! Left one last comment and then I think it's ready to merge. Do you plan to also do the docs PR?

Yes, I plan to update the docs as well.

src/dvclive/utils.py
skshetry: I'll suggest importing at the top and suppressing on import error.

```python
# top level

try:
   import pandas as pd
except ImportError:
   pd = None
```

And guard/check for None.
```python
if pd and isinstance(datapoints, pd.DataFrame):
    pass
```

dberenbaum: @mnrozhkov I'll defer to @skshetry here as the engineer.

@skshetry would you advise we drop `is_instance_without_import()` (in a separate PR)?

mnrozhkov: @skshetry, thanks for the suggestions! I dropped `is_instance_without_import()` and made it work with this imports: 

```python
if TYPE_CHECKING:
    import numpy as np
    import pandas as pd
else:
    try:
        import pandas as pd
    except ImportError:
        pd = None

    try:
        import numpy as np
    except ImportError:
        np = None
```

It doesn't look nice, but it's the only solution works to meet 2 goals 
- use proper type annotations for `datapoints` 
- meet `mypy` checks

skshetry: It's okay to silence mypy here here.

src/dvclive/utils.py
```diff
@@ -1,12 +1,31 @@
...
+        pd = None
+
+    try:
+        import numpy as np
+    except ImportError:
+        np = None
+
+from .error import (
+    InvalidDataTypeError,
+)
```
skshetry, line 28: Please move this before `if TYPE_CHECKING` block. Also, adding a comma at the end hangs this, so you can remove it and keep it in a single line.
```suggestion
from .error import InvalidDataTypeError
```

tests/test_utils.py
```diff
@@ -15,3 +18,31 @@
 )
 def test_standardize_metric_name(framework, logged, standardized):
     assert standardize_metric_name(logged, framework) == standardized
+
+
+class TestConvertDatapointsToListOfDicts:
```
skshetry, line 23: Please see if you can use `parametrize` here for these four tests.
Also there will be no need to put them into a class after that, as they will drop to just 2 tests: one for valid outputs and other for invalid outputs.

dberenbaum (approved):
Thanks @mnrozhkov! Great improvement here.

PR #740, opened 2023-11-17
hf: warn of deprecating internal callback

dberenbaum:
Adds a deprecation notice about the internal HF callback and suggests to use the new external callback, which will be available in the next release of https://github.com/huggingface/transformers.

Also adds the external callback to tests since there weren't similar tests in the transformers library, and updates the example notebook.

PR #756, opened 2023-12-15
preserve summary metrics on resume

dberenbaum:
Dvclive was not preserving summary metrics when `resume=True`.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -121,6 +121,7 @@ class Live:
 
     def _init_resume(self):
         self._read_params()
+        self.summary = self.read_latest()
         self._step = self.read_step()
         if self._step != 0:
             logger.info(f"Resuming from step {self._step}")
@@ -607,14 +608,14 @@ class Live:
         cleanup_dvclive_step_completed()
 
     def read_step(self):
-        if Path(self.metrics_file).exists():
-            latest = self.read_latest()
-            return latest.get("step", 0)
-        return 0
+        latest = self.read_latest()
+        return latest.get("step", 0)
 
     def read_latest(self):
-        with open(self.metrics_file, encoding="utf-8") as fobj:
-            return json.load(fobj)
+        if Path(self.metrics_file).exists():
+            with open(self.metrics_file, encoding="utf-8") as fobj:
+                return json.load(fobj)
+        return {}
 
     def __enter__(self):
         self._inside_with = True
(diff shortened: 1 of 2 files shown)


src/dvclive/live.py
```diff
@@ -607,14 +608,14 @@ def end(self):
...
+        latest = self.read_latest()
+        return latest.get("step", 0)
 
     def read_latest(self):
-        with open(self.metrics_file, encoding="utf-8") as fobj:
-            return json.load(fobj)
+        if Path(self.metrics_file).exists():
+            with open(self.metrics_file, encoding="utf-8") as fobj:
+                return json.load(fobj)
+        return {}
```
dberenbaum, line 618: This was just refactoring so I could use `read_latest` elsewhere without having to check if the summary metrics file exists.

tests/test_resume.py
```diff
@@ -24,9 +26,14 @@ def test_resume(tmp_dir, resume, steps, metrics):
     for new_metric in [0.7, 0.6]:
         dvclive.log_metric("metric", new_metric)
         dvclive.next_step()
+    dvclive.end()
 
     assert read_history(dvclive, "metric") == (steps, metrics)
     assert read_latest(dvclive, "metric") == (steps[-1], metrics[-1])
+    if resume:
+        assert dvclive.read_latest()["summary"] == 1
```
dberenbaum, line 34: Without this PR, the `summary` metric would have been lost

dberenbaum:
Test failure is unrelated (see [this comment](https://github.com/iterative/dvclive/pull/740#issuecomment-1852846093))

tests/test_resume.py
```diff
@@ -24,9 +26,14 @@ def test_resume(tmp_dir, resume, steps, metrics):
     for new_metric in [0.7, 0.6]:
         dvclive.log_metric("metric", new_metric)
         dvclive.next_step()
+    dvclive.end()
```
shcheklein, line 29: QQ: is this one actually needed or it's just for a symmetry?

dberenbaum: Just for symmetry/best practice. I think the original year might have been from before we had and end method.

shcheklein (approved):

shcheklein:
We have a test failure, needs to be fixed. Otherwise good catch, thanks @dberenbaum !

dberenbaum:
@shcheklein See the comment above about the test failure being unrelated. We will keep having that test failure until:
1. #740 is merged.
2. https://github.com/huggingface/transformers/pull/27983 is released.

We in fact have probably had the test failure for a long time, but we weren't testing the frameworks in CI until recently.

PR #757, opened 2023-12-15
add back nox

dberenbaum:
Adding back nox per https://github.com/iterative/dvclive/pull/749#issuecomment-1857226519. Looks like it was removed in https://github.com/iterative/dvclive/pull/527 without too much thought beyond one user running into an issue with it.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -1,0 +1,61 @@
+"""Automation using nox."""
+import glob
+import os
+
+import nox
+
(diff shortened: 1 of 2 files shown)


dberenbaum:
See https://github.com/iterative/dvclive/pull/756#issuecomment-1858793979 for why a test is failing - it's unrelated to this PR.

skshetry (approved):

PR #762, opened 2024-01-04
Fix `TypeError` by explictly calling shadowed `builtins.type`

### LINKED ISSUE #761: Invalid path causes `TypeError` due to shadowed builtin `type`
When you pass an invalid path type to `Live.log_artifact`, it's supposed to raise an `InvalidDataTypeError`:

https://github.com/iterative/dvclive/blob/47c87b3de0d727fe4bb4e46fbe76839bfa692e0e/src/dvclive/live.py#L493-L494

Instead, it raises `TypeError: str is not callable` because the builtin `type` is shadowed by the parameter:

https://github.com/iterative/dvclive/blob/47c87b3de0d727fe4bb4e46fbe76839bfa692e0e/src/dvclive/live.py#L484

It's a pretty minor issue, but I thought I'd point it out since it could cause some confusion. Thanks!

aazuspan:
  guide.

Closes #761 by explicitly calling `builtins.type` instead of the `type` argument that's shadowing it.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+import builtins
 import glob
 import json
 import logging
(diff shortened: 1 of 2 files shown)


dberenbaum (approved):
Thanks @aazuspan!

PR #770, opened 2024-02-05
add typehints to Live's public methods

AlexandreKempf:
  guide.

- [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here. https://github.com/iterative/dvc.org/pull/5103

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -15,7 +15,6 @@ if TYPE_CHECKING:
     import pandas as pd
     import matplotlib
     import PIL
-    from dvc.repo import Repo
 
 from dvc.exceptions import DvcException
 from funcy import set_in
@@ -104,7 +103,7 @@ class Live:
         self._experiment_rev: Optional[str] = None
         self._inside_dvc_exp: bool = False
         self._inside_dvc_pipeline: bool = False
-        self._dvc_repo: Optional[Repo] = None
+        self._dvc_repo = None
         self._include_untracked: List[str] = []
         if env2bool(env.DVCLIVE_TEST):
             self._init_test()
@@ -541,7 +540,7 @@ class Live:
             )
 
     @catch_and_warn(DvcException, logger)
-    def cache(self, path: str):
+    def cache(self, path):
         if self._inside_dvc_pipeline:
             existing_stage = find_overlapping_stage(self._dvc_repo, path)
 
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -38,7 +38,7 @@ from .error import (
 from .plots import PLOT_TYPES, SKLEARN_PLOTS, CustomPlot, Image, Metric, NumpyEncoder
 from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
-from .studio import StudioEventKind, get_dvc_studio_config, post_to_studio
+from .studio import get_dvc_studio_config, post_to_studio
 from .utils import (
     StrPath,
     catch_and_warn,
@@ -64,7 +64,7 @@ handler.setFormatter(formatter)
 logger.addHandler(handler)
 
 ParamLike = Union[int, float, str, bool, List["ParamLike"], Dict[str, "ParamLike"]]
-SkleanPlotKind = [*SKLEARN_PLOTS.keys()]
+SklearnPlotKind = [*SKLEARN_PLOTS.keys()]
 
 
 class Live:
@@ -443,7 +443,7 @@ class Live:
 
     def log_sklearn_plot(
         self,
-        kind: SkleanPlotKind,
+        kind: SklearnPlotKind,
         labels: Union[List, np.ndarray],
         predictions: Union[List, Tuple, np.ndarray],
         name: Optional[str] = None,
@@ -588,7 +588,7 @@ class Live:
         make_dvcyaml(self)
 
     @catch_and_warn(DvcException, logger)
-    def post_to_studio(self, event: StudioEventKind):
+    def post_to_studio(self, event: str):
         post_to_studio(self, event)
 
     def end(self):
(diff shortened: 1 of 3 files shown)


src/dvclive/live.py
```diff
@@ -379,7 +391,11 @@ def log_metric(
         self.summary = set_in(self.summary, metric.summary_keys, val)
         logger.debug(f"Logged {name}: {val}")
 
-    def log_image(self, name: str, val):
+    def log_image(
+        self,
+        name: str,
+        val: Union[np.ndarray, matplotlib.figure.Figure, PIL.Image, StrPath],
```
dberenbaum, line 397: It looks like `StrPath` expects `Path`:

https://github.com/iterative/dvclive/blob/38f2f51c300a847e60894eff47f65fc36e4a9a25/src/dvclive/utils.py#L29

The `could_log()` method instead checks for `PurePath`, which is the [parent](https://docs.python.org/3/library/pathlib.html) of `Path`:

https://github.com/iterative/dvclive/blob/38f2f51c300a847e60894eff47f65fc36e4a9a25/src/dvclive/plots/image.py#L28

Maybe we should change `StrPath` to use `PurePath`?

src/dvclive/live.py
```diff
@@ -574,7 +597,7 @@ def make_dvcyaml(self):
         make_dvcyaml(self)
 
     @catch_and_warn(DvcException, logger)
-    def post_to_studio(self, event):
+    def post_to_studio(self, event: Literal["start", "data", "done"]):
```
dberenbaum, line 600: Consider importing from https://github.com/iterative/dvc-studio-client/blob/2bac67ace4d33f013ce247f71ead5df1e6cedd87/src/dvc_studio_client/schema.py#L16

src/dvclive/live.py
```diff
@@ -434,7 +450,14 @@ def log_plot(
         plot.dump(datapoints)
         logger.debug(f"Logged {name}")
 
-    def log_sklearn_plot(self, kind, labels, predictions, name=None, **kwargs):
+    def log_sklearn_plot(
+        self,
+        kind: Literal["calibration", "confusion_matrix", "precision_recall", "roc"],
```
dberenbaum, line 455: I think it's missing `"det"`? 

Also, can we import from here:

https://github.com/iterative/dvclive/blob/38f2f51c300a847e60894eff47f65fc36e4a9a25/src/dvclive/plots/__init__.py#L7-L13

dberenbaum: I see it's missing in the docs also

AlexandreKempf:
@dberenbaum  Thanks for the feedback. I'll take your suggestions into account.
Meanwhile, I have a broader question on type hints. I agree that the current implementation for StrPath or ParamLike is cleaner the current way. But when an editor displays the help for each function, it is not very informative at first sight to have a `StrPath` type as a user, and I would rather have a `str | pathlib.Path`. Do you know a way to have the best of both worlds?

dberenbaum:
> Do you know a way to have the best of both worlds?

🤔 Good point. I don't know a way, but might be worth a little research. If you think it's helpful to have the more explicit types, I'm fine to keep them that way.

src/dvclive/live.py
AlexandreKempf: Great catch. I don't know 100% of the codebase, but if we are not using `pathlib.Path` methods to read/write files, we should be good using only `PurePath` ;)

AlexandreKempf: I also change two instance checks in `log_image` and `log_artifact`.

AlexandreKempf:
I asked the question to tech people; we'll see.

I believe it will take more effort to maintain, but it will be way more user-friendly to have a list of possible plots like `["linear," "simple, ...]` for the user to play with and discover the tool. Rather than an opaque `[*BASE_SCHEMA.schema["type"].validators]`. (I took this one from the studio-event data format). 

You see my point, I guess, we just need to figure out own much a pain it will be to maintain to know if it is worth it :)

src/dvclive/live.py
AlexandreKempf: https://github.com/iterative/dvc.org/pull/5103

src/dvclive/live.py
```diff
@@ -62,16 +64,17 @@
 logger.addHandler(handler)
 
 ParamLike = Union[int, float, str, bool, List["ParamLike"], Dict[str, "ParamLike"]]
+SkleanPlotKind = [*SKLEARN_PLOTS.keys()]
```
dberenbaum, line 67:
```suggestion
SklearnPlotKind = [*SKLEARN_PLOTS.keys()]
```

dberenbaum:
As mentioned by @jonburdo, I take back importing from elsewhere for type checking. It seems like an anti-pattern with more thought, and we are better off listing the choices or probably even just leaving the type hints more general for now.

dberenbaum:
@AlexandreKempf Could you check the CI failures?

jonburdo:
> I believe it will take more effort to maintain, but it will be way more user-friendly to have a list of possible plots like `["linear," "simple, ...]` for the user to play with and discover the tool.

I think you're right for a relatively short list like this. I sometimes take a look at other popular libraries for questions like this. Running `rg '\bLiteral\['` (ripgrep) in the numpy and typeshed repos shows they generally just repeat the whole thing rather than using a variable. I imagine it won't be too hard to maintain.

AlexandreKempf:
Thanks for the advice @jonburdo. What do you think about that @dberenbaum? 
I would say we go with Jon's advice here: our lists are small and useful for users for editors' suggestions. If it is too hard to maintain, we can always revert it at no cost :)

dberenbaum:
Yes, I don't think it's worth worrying about the maintenance costs at this point

PR #773, opened 2024-02-06
Monitors CPU, RAM, and disk usage

AlexandreKempf:
# New feature: monitoring for CPU

Link to [GPU monitoring](https://github.com/iterative/dvclive/pull/785).
Link to [fix Studio](https://github.com/iterative/dvclive/pull/788)

### Monitoring CPU hardware

In this PR we add the possibility for the user to monitor the CPU, RAM, and disk during one experiment.

To use this feature you can use it with a simple argument:
```python
from dvclive import Live

with Live(monitor_system=True) as live:
    ...
```

If you want to use advance features you can specify each parameter this way:
```python
from dvclive import Live
from dvclive.monitor_system import CPUMonitor

with Live(monitor_system=True) as live:
    live.cpu_monitor = CPUMonitor(interval = 0.1, num_samples=15, directories_to_monitor={"data": "/path/to/data/disk", "home": "/home"}))
```
And this is how the editor help should looks like: 
Screenshot from 2024-02-15 12-27-04

If you allow the monitoring of your system, if will track:
 - `system/cpu/count` -> number of CPU cores
 - `system/cpu/usage (%)` ->  the average usage of the CPUs. 
 - `system/cpu/parallelization (%)` -> How many CPU cores use more than 20% of their possibilities? It is useful when you're looking to parallelize your code to train your model or process your data faster. 
  - `system/ram/usage (%)` -> percentage of the RAM used. Useful to increase batch size or data processed at the same time in the RAM.
  - `system/ram/usage (GB)` -> RAM used. Useful to increase batch size or data processed at the same time. 
  - `system/ram/total (GB)` -> Total RAM in your system
  - `system/disk/usage (%)` -> Amount of disk used at a given path in %. By default uses "." meaning where the python executor was launched. You can specify the paths you want to monitor. For instance the code example above monitor `/data` and `/home`. Data and code often lives in very different paths/volumes, so it is useful for the user to be able to specify its own path. Note that as their can be several paths specified, the full metric name is `system/disk/usage (%)/<user defined name>`. For instance it would be `system/disk/usage (%)/data` for the `/path/to/data/disk` and `system/disk/usage (%)/home` for `/home`.
  - `system/disk/usage (GB)` ->  Amount of disk used at a given path.
  - `system/disk/total (GB)`  ->  Amount of total disk storage at a given path.

All the values that can change during an experiment can be saved as plots. Timestamps are automatically recorded with the metrics. Other metrics (that don't change) such as CPU count, RAM total and disk total are saved as metrics but cannot be saved as plots. 

I decided to split the usage in % and GB. First, because it is more consistent with the other loggers out there. Second, both are extremely relevant based on which cloud instance you run your experiment. If you only run your experiment on the same hardware, the distinction is not really interesting.

### Files generated
The metrics about the CPU are stored with the `log_metric` function. It means that the `.tsv` files are stored in the `dvclive/plots` folder. A specific folder, `system`, contains all the metrics about the CPU to distinguish them from the user-defined metrics. The metrics are also saved in the `dvclive/metrics.json` file. 

### Plot display
Here is what VScode extension looks like:
image

Here is what Studio looks like:
image
image

Note that studio update is a little buggy, but it is fixed in [this PR](https://github.com/iterative/dvclive/pull/788)

---

### DIFF
--- a/src/dvclive/monitor_system.py
+++ b/src/dvclive/monitor_system.py
@@ -52,7 +52,7 @@ class _SystemMonitor(abc.ABC):
 
         min_num_samples = 1
         max_num_samples = 30
-        if min_num_samples < num_samples < max_num_samples:
+        if not min_num_samples < num_samples < max_num_samples:
             num_samples = max(min(num_samples, max_num_samples), min_num_samples)
             logger.warning(
                 f"System monitoring `num_samples` should be between {min_num_samples} "


pyproject.toml
```diff
@@ -53,11 +54,7 @@ tests = [
     "dvclive[image,plots,markdown]",
     "ipython",
 ]
-dev = [
-    "dvclive[all,tests]",
-    "mypy>=1.1.1",
-    "types-PyYAML",
-]
+dev = ["dvclive[all,tests]", "mypy>=1.1.1", "types-PyYAML"]
```
shcheklein, line 57: minor: do other changes to this file come from applying some linter? it should be then automated and ideally (not a blocker, no need to change anything in this PR) should be done as a separate PR. The reason is that we can't rely on some local personal settings / setup. People have different preferences, editors - we'll keep getting formatting updates.

AlexandreKempf: I reverted the formatting changes in `pyproject.toml` to introduce it in a cleaner way: https://github.com/iterative/dvclive/pull/775

dberenbaum:
@AlexandreKempf Sorry for the delay here. For any PR like this where you introduce a new feature, it would be great if you could include a demo of the feature in the PR. Some useful ways to demo it can be:
- Record a video via screencast, https://www.loom.com, https://asciinema.org, etc.
- Include an example notebook, script, repo, etc.
- Write a draft of the docs PR

Once it's merged, you can reuse it by posting it to https://iterativeai.slack.com/archives/C064XEWG2BD

dberenbaum:
Also note that windows tests are failing

src/dvclive/studio.py
```diff
@@ -14,7 +14,7 @@
 
 
 def _get_unsent_datapoints(plot, latest_step):
-    return [x for x in plot if int(x["step"]) > latest_step]
+    return [x for x in plot if int(x["step"]) >= latest_step]
```
AlexandreKempf, line 17: Best solution I could think of to send CPU metrics to studio. Otherwise `latest_step` == the step they are logged at and nothing is sent to studio.

src/dvclive/studio.py
```diff
@@ -53,7 +53,7 @@ def _adapt_images(live):
     return {
         _adapt_path(live, image.output_path): {"image": _adapt_image(image.output_path)}
         for image in live._images.values()
-        if image.step > live._latest_studio_step
+        if image.step >= live._latest_studio_step
```
AlexandreKempf, line 56: same logic = same change ^^

AlexandreKempf: It is not needed for the PR tho. Tell me what is the policy. Should I change it to `>=` or keep it as `>` ?

src/dvclive/studio.py
AlexandreKempf: After a quick thought, it is not a good solution for the problem.

AlexandreKempf: The real solution is probably close to the hack in the `log_metrics` method in the lightning logger. But I'm not 100% sure on why this hack was introduced.

AlexandreKempf: for info: https://github.com/iterative/dvclive/issues/477

src/dvclive/live.py
```diff
@@ -345,6 +362,7 @@ def sync(self):
 
         self.make_report()
 
+        logger.warning(f"post_to_studio step{self.step}")
```
dberenbaum, line 365: Do we want to keep these or were they only for debugging? If we keep them, let's lower the logging level so they don't show by default and aren't considered warnings.

dberenbaum:
Sorry if I forget this from your previous work on this, but is GPU usage a separate callback you plan to add later?

> * plots use steps as `x` by default when we log metrics. Using timestamps makes more sense if available (`timestamp=True`). It is weird for the monitoring resources to be plotted with steps.

Can we configure these plots to use timestamps as the x-axis? WDYT about putting these into a separate dir like `dvclive/plots/system`? 

I guess it goes beyond the idea of a generic callback system, but seems like a cleaner UX. I'm also unsure whether we need to expose a whole callbacks system to users for this, or whether there's a simpler way for them to enable/disable logging system metrics (even if we use callbacks internally).

> * it can be nice to have a time flag system to know where we are in training (steps are not very informative). This is very useful to align the hardware resources with specific algorithm parts.

Sorry, I don't follow. How is it different from the previous point?

src/dvclive/live.py
AlexandreKempf: They are only for debugging propose :)

shcheklein:
Thanks for the description and video @AlexandreKempf . It's way easier to review it now and understand.

A few questions, before reviewing the code itself in details:

- Why do we need a separate callback? It's a bit confusing to pass a callback into a callback. Is it what other product do? E.g. W&B , neptune, comet? Can we review and summarize how they do this?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def monitoring_loop(self):
+        while not self._shutdown_event.is_set():
+            self._cpu_metrics = {}
+            for _ in range(int(self.duration // self.interval)):
+                last_cpu_metrics = {}
+                try:
+                    last_cpu_metrics = get_cpus_metrics()
+                except psutil.Error:
+                    logger.exception("Failed to monitor CPU metrics:")
```
shcheklein, line 44: should it be logged only once?

shcheklein: have you tried to hit it? how will the ":" at the end be handled? is it required?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class CPUMetricsCallback:
+    def __init__(
+        self,
+        duration: Union[int, float] = 5,
+        interval: Union[int, float] = 1.0,
+        plot: bool = True,
+    ):
+        self.duration = duration
```
shcheklein, line 24: do we expect those to be accessed / changed? let's make the protected with `_duration`?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+        self._shutdown_event.set()
+
+
+def get_cpus_metrics() -> Dict[str, Union[float, int]]:
+    ram_info = psutil.virtual_memory()
+    io_info = psutil.disk_io_counters()
+    nb_cpus = psutil.cpu_count()
+    cpus_percent = psutil.cpu_percent(percpu=True)
+    return {
+        "system/cpu/usage_avg_percent": mean(cpus_percent),
```
shcheklein, line 69: how does it looks like across different products? I mean all these different names. Also, what do they show as X?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+                    plot=self.plot
+                    if metric_name not in self._no_plot_metrics
+                    else False,
+                )
+
+    def end(self):
+        self._shutdown_event.set()
+
+
+def get_cpus_metrics() -> Dict[str, Union[float, int]]:
```
shcheklein, line 63: make it protected?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+
+class CPUMetricsCallback:
+    def __init__(
+        self,
+        duration: Union[int, float] = 5,
+        interval: Union[int, float] = 1.0,
+        plot: bool = True,
+    ):
+        self.duration = duration
+        self.interval = interval
```
shcheklein, line 25: should do at least basic sanity checks for values ?

tests/test_system_metrics.py
```diff
@@ -0,0 +1,137 @@
...
+        (2.0, 1, True),
+    ],
+)
+def test_cpumetricscallback_with_plot(tmp_dir, duration, interval, plot):
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        callbacks=[CPUMetricsCallback(duration, interval, plot)],
+    ) as live:
+        time.sleep(duration * 2)
```
shcheklein, line 70: can we do w/o sleep?

also, I'm afraid that logic where we just wait for thread with a sleep can be unreliable - there is not real guarantee for it to exit? it can lead to flakey tests

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    def monitoring_loop(self):
+        while not self._shutdown_event.is_set():
+            self._cpu_metrics = {}
+            for _ in range(int(self.duration // self.interval)):
+                last_cpu_metrics = {}
+                try:
+                    last_cpu_metrics = get_cpus_metrics()
+                except psutil.Error:
+                    logger.exception("Failed to monitor CPU metrics:")
+                self._cpu_metrics = append_dict(self._cpu_metrics, last_cpu_metrics)
```
shcheklein, line 45: pls, check funcy lib ... just in case .. potentially zip or something like that can be used? (so that we don't have additional helpers or we can simplify them)

AlexandreKempf:
@dberenbaum 
> Can we configure these plots to use timestamps as the x-axis? WDYT about putting these into a separate dir like `dvclive/plots/system`?
Yeah, that is a cool idea. Initially, I put them into `dvclive/plots/metrics/system` because I use the `log_metric` function that naively puts them here. And I tried to change the dvc.yaml to have the `system` subdirectory to take `timestamp` as X. But this doesn't work if a parent directory is set to another X (in that case, `dvclive/plots/metrics` was set to `step`. 

> I guess it goes beyond the idea of a generic callback system, but seems like a cleaner UX. I'm also unsure whether we need to expose a whole callbacks system to users for this, or whether there's a simpler way for them to enable/disable logging system metrics (even if we use callbacks internally).
I agree with the UX, and I also find it more modular and easy to extend (with GPU, TPU, pricing for cloud ...). But I agree with you and Ivan, the user doesn't need to know how it works under the hood. And the callback in a callback is not great either. I'll change it to only take a `monitor_system`  boolean. 

> Sorry, I don't follow. How is it different from the previous point?

I was thinking of something like this:
image

src/dvclive/system_metrics.py
AlexandreKempf: Yeap, problem resolved with `funcy`, thanks! I didn't want to introduce another dependency to the project, but it was already used.

src/dvclive/system_metrics.py
AlexandreKempf: WandB are doing some clipping on the values: https://github.com/wandb/wandb/blob/cdb7ffe67e19402d4a0728de1f1be4471cef702d/wandb/sdk/internal/system/assets/interfaces.py#L120
If we [agree that the user shouldn't have access to the values](https://github.com/iterative/dvclive/pull/773#discussion_r1485707542), then there is no need to sanity check these (but we can check for their validity in unit tests)

src/dvclive/system_metrics.py
AlexandreKempf: Fine by me. As a datascientist, I wouldn't care about these values. I would just like this monitoring feature to take as little CPU usage as possible.

src/dvclive/studio.py
AlexandreKempf: Still working on that ...

src/dvclive/system_metrics.py
AlexandreKempf: ### W&B
docs and naming: https://docs.wandb.ai/guides/app/features/system-metrics
W&B dashboard: https://wandb.ai/stacey/estuary?workspace=user-lavanyashukla

They show the timestamp in the X axis, but it is in hours and in respect of the beginning of the experiment. I really like their X axis to be honest. Names are spaced by " ". They do have the % usage per CPU core. It is a nice feature, but we will need to generate a complicated `dvc.yaml` configuration to display everything in one graph.

### labML
visual: https://github.com/labmlai/labml/blob/master/guides/hardware_monitoring.md
code: https://github.com/labmlai/labml/blob/6dc2a2ad3bd3a46c5dab9be80f3030c5ac8cb133/client/labml/internal/computer/monitor/scanner.py#L141
Their interface is way more simple than W&B. They use timestamp as the x-axis, but it is formatted nicely (day and hour, not a large int). They group metrics (like read/write on disk, memory available/remaining ...) which is nice and we should probably do the same. 

### Neptune
Dashboard: https://app.neptune.ai/o/showcase/org/onboarding-project/runs/details?viewId=98f66b32-2279-4b73-8210-863021c440ac&detailsTab=monitoring&shortId=IMG-155
doc: https://docs.neptune.ai/logging/system_metrics/

They are using very simple metrics, displayed by step (which I don't like). But it is quite easy to change the x-axis on their dashboard and they have relative timestamp and absolute timestamp as options. 


### Conclusion 
Imho, we should aim for timestamp as the default behavior. Ideally with a human display (rather than a large int). We should gather metrics together in one graph (like all the CPU cores ...). This should avoid creating weird "parallelization" metrics that are not well-defined.

tests/test_system_metrics.py
AlexandreKempf: I wanted to test the CPUMetrics' behavior and how it gathers metrics over time. How do you think we should do that?

AlexandreKempf:
@dberenbaum Revert all the changes about `timestamp` so that we use `step` instead. We might change it in the future, but for now, it produces this result when we compare two experiments: 
image
Which is not satisfying.
I tried to find a way to set the x-axis as relative to the first timestamp on vega-lite but I couldn't find the right config to make it work. 


Also I fixed the lightning issue and made sure it works with the vanilla one (`with Live() as live`).

dberenbaum:
> I tried to find a way to set the x-axis as relative to the first timestamp on vega-lite but I couldn't find the right config to make it work.

I think we could log the relative timestamp to the tsv file, but I'm fine to make this a follow-up issue and proceed with step for now.

AlexandreKempf:
> I think we could log the relative timestamp to the tsv file, but I'm fine to make this a follow-up issue and proceed with step for now.

It is sad that is the only way of dealing with it, and that Vega-lite doesn't seem to support that (even with all the transformations they provide). I'll give it a go today.

AlexandreKempf:
FYI I changed the name of the metrics so that they are a bit simpler and consistent with the other platforms (W&B, Neptune ...).

dberenbaum:
> I find this roadmap cleaner than trying to hack our way into the solution. What do you think?

Let's create a separate issue and discuss there.

Do we also need a follow-up issue for gpu metrics? Any other items that you consider out of scope but want to open an issue for?

Is it otherwise ready for review?

AlexandreKempf:
Yeah, I was waiting for this PR to be reviewed before I started working on GPU monitoring. No need to add the issue for the GPU, It'll be done by tomorrow on another PR. I'll create any relevant issues I can think of and that are missing here.

src/dvclive/studio.py
```diff
@@ -71,15 +60,18 @@ def get_studio_updates(live):
...
-    plots = {
-        _adapt_path(live, name): _adapt_plot_datapoints(live, plot)
-        for name, plot in plots.items()
-    }
-    plots = {k: {"data": v} for k, v in plots.items()}
+    plots_to_send = {}
+    for name, plot in plots.items():
+        path = _adapt_path(live, name)
+        nb_points_already_sent = live._nb_points_sent_to_studio.get(path, 0)
+        live._nb_points_sent_to_studio[path] = len(plot)
```
dberenbaum, line 67: Should we do this later after the call succeeds? Will this skip points if a call fails?

tests/test_system_metrics.py
dberenbaum: @iterative/dvc Any ideas on how to test this?

src/dvclive/studio.py
```diff
@@ -13,17 +13,11 @@
...
 
-def _get_unsent_datapoints(plot, latest_step):
-    return [x for x in plot if int(x["step"]) > latest_step]
-
-
 def _cast_to_numbers(datapoints):
     for datapoint in datapoints:
         for k, v in datapoint.items():
-            if k == "step":
+            if k in ["step", "timestamp"]:
```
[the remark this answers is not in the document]
AlexandreKempf: When I was playing with `timestamp`, I realized that it is decoded as a `str`, so I introduced this change to make it an `int` (which makes more sense, I think). What do you think?

src/dvclive/system_metrics.py
```diff
@@ -0,0 +1,91 @@
...
+        * 100
+        / nb_cpus,
+        "system/ram/usage (%)": ram_info.percent,
+        "system/ram/usage (GB)": (ram_info.percent / 100)
+        * (ram_info.total / GIGABYTES_DIVIDER),
+        "system/ram/total (GB)": ram_info.total / GIGABYTES_DIVIDER,
+        "system/io/read speed (MB)": io_info.read_bytes
+        / (io_info.read_time * MEGABYTES_DIVIDER),
+        "system/io/write speed (MB)": io_info.write_bytes
+        / (io_info.write_time * MEGABYTES_DIVIDER),
```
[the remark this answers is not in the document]
AlexandreKempf: I know that some people read data from the disk when they train a ML model (like every batch, they read data from the disk again and again). It is the recommended way it your dataset doesn't fit in memory. That is why I included it here so that it can be monitored. But actually, now that you mention it, I don't know how these values help you debug/improve anything. 
W&B has a similar concept (the total amount of GB read or write to disk since the beginning of the experiment). They also have (like LabML) the amount of disk taken / available. Neptune has nothing related to disk usage. 
My feeling is that is it best to have the amount of disk used / available (like labML and W&B).

I'll do the change if you agree.

tests/test_system_metrics.py
AlexandreKempf: The problem is that a step could be very long in real experiments (like one hour) and you probably want to have a finer resolution for hardware metrics :)

src/dvclive/studio.py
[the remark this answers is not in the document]
dberenbaum: It makes sense, but I'm not sure why it was explicitly skipped and left as a string before. Do you know if studio or dvc-studio-client expect a certain type?

src/dvclive/system_metrics.py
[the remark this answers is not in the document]
dberenbaum: I'm fine with whatever you decide here. Just curious to get your thoughts. It should be easy to change these later if needed.

src/dvclive/studio.py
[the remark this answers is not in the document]
AlexandreKempf: My guess was that it was always used raw. Both `numbers` and `str` works well if you're not transforming data in vega-lite. As soon as you want to add or subtract values, then `str` values behave weirdly in vega-lite but `numbers` behave correctly.

tests/test_system_metrics.py
skshetry: Should `Live` itself run a loop to re-collect from callbacks time to time?

src/dvclive/live.py
```diff
@@ -379,6 +397,17 @@ def step(self, value: int) -> None:
...
+    def monitor_cpu(self) -> Optional[MonitorCPU]:
+        return self._monitor_cpu or None
+
+    @monitor_cpu.setter
+    def monitor_cpu(self, monitor_cpu: MonitorCPU) -> None:
+        if self._monitor_cpu is not None:
+            self._monitor_cpu.end()
+        self._monitor_cpu = monitor_cpu
+        self._monitor_cpu(self)
+
```
dberenbaum, line 410: Is it necessary to have this and to have similar logic in init and end? Why is it here for monitor_cpu but there's nothing similar in the gpu pr?

tests/test_monitor_system.py
```diff
@@ -0,0 +1,126 @@
...
+        monitor.end()
+
+    assert not Path(missing_directories).exists()
+
+    assert "system/disk/usage (%)/1" not in metrics
+    assert "system/disk/usage (GB)/1" not in metrics
+    assert "system/disk/total (GB)/1" not in metrics
+
+
+def test_monitor_system(tmp_dir, mocker):
```
dberenbaum, line 82: Still not that comfortable with this test. WDYT about splitting into 2 tests?

1. Call `Live(monitor_system=True)` and only test that it sets the monitor, calls it, and ends it.
2. Call `live = Live()` followed by `self._monitor_cpu = MonitorCPU(interval=0, num_samples=1)` and test that it logs all the expected values.

cc @iterative/dvc

shcheklein:
folks, product question (I think I asked it before, but I'm can't find it and if it was answered before) - do we have to do sub-callbacks / modules? can it be all part of the DVCLive itself? Is it automatic and default behavior in other products?

src/dvclive/live.py
```diff
@@ -7,8 +7,19 @@
 import os
 import shutil
 import tempfile
+
 from pathlib import Path, PurePath
-from typing import Any, Dict, List, Optional, Set, Tuple, Union, TYPE_CHECKING, Literal
+from typing import (
```
shcheklein, line 12: does it change only formatting ? not critical, but can we do it please in a separate PR? (not a blocker, minor)

src/dvclive/live.py
```diff
@@ -379,6 +397,17 @@ def step(self, value: int) -> None:
         self._step = value
         logger.debug(f"Step: {self.step}")
 
+    @property
+    def monitor_cpu(self) -> Optional[MonitorCPU]:
+        return self._monitor_cpu or None
```
shcheklein, line 402: let's rename it to `_cpu_monitor` (to distinguish from monitor_system that is bool) ? (again, minor, but when I'm reading this code both blend a bit and make me think that there are multiple bool flags - e.g. monitor CPU, monitor GPU in the future that could be enabled / disabled)

shcheklein: or may be rename `monitor_system` to make is simpler ... again, not a blocker, just sharing my experience reading this after a break

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+MEGABYTES_DIVIDER = 1024.0**2
+GIGABYTES_DIVIDER = 1024.0**3
+
+MINIMUM_CPU_USAGE_TO_BE_ACTIVE = 20
+
+
+class _MonitorSystem:
+    """
+    Monitor system resources and log them to DVC Live.
+    Use a separate thread to call a `_get_metrics` function at  fix interval and
```
shcheklein, line 23: minor: extra space before `fix`

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+logger = logging.getLogger("dvclive")
+MEGABYTES_DIVIDER = 1024.0**2
+GIGABYTES_DIVIDER = 1024.0**3
+
+MINIMUM_CPU_USAGE_TO_BE_ACTIVE = 20
+
+
+class _MonitorSystem:
```
shcheklein, line 20: it it abstract class, consider using ABC (I'm not an expert in those, but it seems we should be using it here cc @skshetry )

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+    _plot_blacklist_prefix: Tuple = ()
+
+    def __init__(
+        self,
+        interval: float = 0.5,
+        num_samples: int = 10,
+        plot: bool = True,
+    ):
+        if not isinstance(interval, (int, float)):
+            raise InvalidDataTypeError("MonitorSystem.interval", type(interval))
```
shcheklein, line 36: let's just use ValueError here? InvalidDataTypeError is primarily about user data point, etc ... not arguments

shcheklein: the first argument here is usually the value that user passed, not name of the argument

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+    def _monitoring_loop(self):
+        while not self._shutdown_event.is_set():
+            self._metrics = {}
+            for _ in range(self._nb_samples):
+                last_metrics = {}
+                try:
+                    last_metrics = self._get_metrics()
+                except psutil.Error:
+                    if self._warn_user:
+                        logger.exception("Failed to monitor CPU metrics")
```
shcheklein, line 63: does it dump the whole stack trace? should we just include message if we are not in the DEBUG mode (I think we have a flag for this in DVCLive).

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+        plot: bool = True,
+    ):
+        """Monitor CPU resources and log them to DVC Live.
+
+        Args:
+            interval (float): interval in seconds between two measurements.
+                Defaults to 0.5.
+            num_samples (int): number of samples to average. Defaults to 10.
+            directories_to_monitor (Optional[List[str]]): paths of the directories that
+            needs to be monitor for disk storage metrics. Defaults to the current
```
shcheklein, line 109: typo: monitored

w/o reading the code it was quite hard to understand what does this option mean

shcheklein: this is about disk usage, right?

that's how psutil does this:

`Return disk usage statistics about the partition which contains the given path as a named tuple ...`

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+            "system/ram/usage (%)": ram_info.percent,
+            "system/ram/usage (GB)": (ram_info.percent / 100)
+            * (ram_info.total / GIGABYTES_DIVIDER),
+            "system/ram/total (GB)": ram_info.total / GIGABYTES_DIVIDER,
+        }
+        for idx, directory in enumerate(self._directories_to_monitor):
+            if not Path(directory).exists():
+                continue
+            disk_info = psutil.disk_usage(directory)
+            result[f"system/disk/usage (%)/{idx}"] = disk_info.percent
```
shcheklein, line 153: can we instead of idx get disk name / volume name? (not a blocker)

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+        pass
+
+    def end(self):
+        self._shutdown_event.set()
+
+
+class MonitorCPU(_MonitorSystem):
+    _plot_blacklist_prefix: Tuple = (
+        "system/cpu/count",
+        "system/ram/total (GB)",
```
shcheklein, line 91: we need to make all this names as const, or even introduce a scheme (voluptuous, etc) ...since we are using these in a few places

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,158 @@
...
+                        logger.exception("Failed to monitor CPU metrics")
+                        self._warn_user = False
+
+                self._metrics = merge_with(sum, self._metrics, last_metrics)
+                self._shutdown_event.wait(self._interval)
+                if self._shutdown_event.is_set():
+                    break
+            for name, values in self._metrics.items():
+                blacklisted = any(
+                    name.startswith(prefix) for prefix in self._plot_blacklist_prefix
```
shcheklein, line 72: I think it's better to have exact metrics names vs startswith. startswith can lead to unexpected false positives later. Was there a strong reason to use it?

tests/test_monitor_system.py
```diff
@@ -0,0 +1,126 @@
...
+    mock_psutil(mocker)
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        monitor_system=True,
+    ) as live:
+        time.sleep(5 + 1)  # allow the thread to finish
```
shcheklein, line 89: just some ideas how to make it more robust (we need more research on this)

- we can set a smaller interval
- then do while True with a small sleep inside and attempt to read the file, or check the Live class to see if there are new metrics already

the problem with the current approach primarily is that even after 6 seconds there is no guarantee, also it's a long time

tests/test_monitor_system.py
```diff
@@ -0,0 +1,126 @@
...
+    assert "system/ram/total (GB)" in metrics
+    assert "system/disk/usage (%)/0" in metrics
+    assert "system/disk/usage (%)/1" in metrics
+    assert "system/disk/usage (GB)/0" in metrics
+    assert "system/disk/usage (GB)/1" in metrics
+    assert "system/disk/total (GB)/0" in metrics
+    assert "system/disk/total (GB)/1" in metrics
+
+
+def test_ignore_missing_directories(tmp_dir, mocker):
```
shcheklein, line 62: missing == non-existent?

src/dvclive/live.py
AlexandreKempf: I updated the GPU PR, and I'll keep updating it based on this discussion.

What do you have in mind? If we don't call `end` the thread will continue. Do you want to keep the same instance?
Something like:
```
with Live(monitor_system=True) as live:
    live.monitor_cpu.interval = 0.1
```
?

src/dvclive/monitor_system.py
AlexandreKempf: I agree with the general conclusion on the false positives. But as soon as we allow the user to specify paths in the metric name, we are forced to test only the beginning of the name.
For instance, the total amount of memory for disk usage looks like this: `system/disks/total (GB)/<path from the user>`. I cannot check the full metric name in that case. 

What I could do is check the full path for metrics names that are not user dependent. Then add the `startswith` check for the metrics name that are user dependant (like disk usage)

tests/test_monitor_system.py
AlexandreKempf: Yeap it is clearer that way :) thx

dberenbaum:
Note there are errors on windows

dberenbaum:
@AlexandreKempf You can just ping here or request a review again once you feel it's ready. 🙏

src/dvclive/live.py
AlexandreKempf: I should definitely make the `end` method private. I don't want the user to use that for now. 

I'm not sure I get you on the first question: "We are starting and ending the thread from the internal self._monitor_cpu already, right?". 
Correct me if I'm wrong, but if the user creates a `Live` object it will assign an instance of CPUMonitor (call "A" for the example). Then it creates another `CPUMonitor` with custom parameters (instance called "B"), and give it to the `Live` object with `live.cpu_monitor = CPUMonitor(...)`. 
If I don't end the instance A, if will continue to run in the background no? This is why I called the `end` method here.

dberenbaum: > Then it creates another `CPUMonitor` with custom parameters (instance called "B"), and give it to the `Live` object with `live.cpu_monitor = CPUMonitor(...)`.

I didn't understand that you want to expose this option to users. It means we will also have to document the whole CPUMonitor class, which breaks the current docs structure of the whole API being inside live.py. It also makes me wonder whether it makes sense to have separate CPUMonitor and GPUMonitor since customizing will likely mean needing to override both. It sounds nice to give all that flexibility, but not sure if it's needed yet/might be better in a separate PR.

src/dvclive/monitor_system.py
AlexandreKempf: I used `voluptuous` to make the test more readable. But I don't see how you want to use it in this file. Do you want to check the schema every time we gather data from psutil? 

I'll replace the string with constant, because I see the direct benefits of it. But I would love to know how you wanted to use voluptuous in that usecase :)

tests/test_monitor_system.py
AlexandreKempf: I agree on splitting the tests to test different functionalities. But the `MonitorCPU(interval=0, num_samples=1)` or any with a small interval value doesn't work if there is no sleep afterward. So I choose a small value to limit the time asleep.
Let me know if the current implementation seems better for you?

src/dvclive/live.py
AlexandreKempf: > I didn't understand that you want to expose this option to users.

Initially, we could have used a hidden object (and the user could not change the interval, and it was ok). But since we added the disk usage statistics, we **need** user inputs. There is no way for us to guess which partition the user wants to monitor. We could guess "/", but there is a high chance the data/images are not located here.

If we want to hide how the CPU monitoring is done to the user, we need to remove the disk usage statistics. 

I agree that the doc organisation could be a problem, even tho we are not forced to do the doc in full details now. The main behavior is to control the system metrics with a boolean value (True => save my system metrics, False => don't). The option of changing the CPUMonitor object can be explained later on if we want to :)

AlexandreKempf: > It also makes me wonder whether it makes sense to have separate CPUMonitor and GPUMonitor since customizing will likely mean needing to override both. It sounds nice to give all that flexibility, but not sure if it's needed

Yeap really good point. I agree on the flexibility boost, but I also agree that it is annoying for the user. There will be lots of import and code to change just the `interval` value. 

I can group them together and make them use the same thread, so `interval`, `num_samples`, and `plot` would be shared. What do you think?

tests/test_monitor_system.py
dberenbaum: Sounds good! It still seems like it could be flaky if the thread gets delayed, but not a blocker for me if it's not happening in practice. No need to duplicate the other discussion with @shcheklein, so let's resolve this and discuss there.

src/dvclive/live.py
dberenbaum: Okay, all your points make sense, esp. if we agree we don't need to document all options immediately, and sorry for the confusion. I'll resolve and leave up to you what you think is best for the public property, including/excluding disk usage, and combining/separating cpu/gpu threads.

dberenbaum (approved):
Would be good to get another look from @shcheklein and/or @skshetry, but product-wise it LGTM. Thanks @AlexandreKempf!

tests/test_monitor_system.py
```diff
@@ -0,0 +1,205 @@
...
+    mock_psutil(mocker)
+    interval = 0.05
+    num_samples = 4
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        monitor_system=False,
+    ) as live:
+        live.cpu_monitor = CPUMonitor(interval=interval, num_samples=num_samples)
+        time.sleep(interval * num_samples + interval)  # log metrics once
```
shcheklein, line 174: this is still unreliable, I would do a while loop with a small sleep interval that reads from disk or checks some Live vat to detect an update

to test that it respects intervals, etc - mock sleep, run thread itself from the test and check the result

shcheklein: just some ideas ^^ that might be a better way, but we can't go with unreliable / flakey tests

tests/test_monitor_system.py
```diff
@@ -0,0 +1,205 @@
...
+    mock_psutil(mocker)
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        monitor_system=False,
+    ) as live:
+        assert live._cpu_monitor is None
+
+
+def test_monitor_system_is_true(tmp_dir, mocker):
```
shcheklein, line 91: name is confusing ...

also we don't (usually) test protected state / vars. What is the purpose of this test?

tests/test_monitor_system.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def test_monitor_system_is_false(tmp_dir, mocker):
+    mock_psutil(mocker)
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        monitor_system=False,
+    ) as live:
+        assert live._cpu_monitor is None
```
shcheklein, line 88: we don't usually test protected state, it's an implementation detail and it's error prone (in a sense that test will pass, but actual implementation is broken)

if the actual goal to to see that it doesn't run monitor system- mock and check that MonitorSystem is not called, and / or check that metrics at the end are not created

tests/test_monitor_system.py
shcheklein, line 91: same here - cleanup is needed, and let's not assert on `_flag` and internal event

tests/test_monitor_system.py
shcheklein: we can also make a public getter probably in this case ..

tests/test_monitor_system.py
```diff
@@ -0,0 +1,205 @@
...
+def test_get_cpus_metrics_mocker(tmp_dir, mocker):
+    mock_psutil(mocker)
+    with Live(
+        tmp_dir,
+        save_dvc_exp=False,
+        monitor_system=False,
+    ) as live:
+        monitor = CPUMonitor(disks_to_monitor={"main": "/", "home": "/"})
+        monitor(live)
+        metrics = monitor._get_metrics()
```
shcheklein, line 60: what is the purpose of this test? why a single one that runs Live at the bottom not enough?

pyproject.toml
```diff
@@ -34,10 +34,12 @@ dependencies = [
   "dvc>=3.33.3",
   "dvc-render>=1.0.0,<2",
   "dvc-studio-client>=0.17.1,<1",
+  "dpath",
```
shcheklein, line 37: is it used only in tests?

pyproject.toml
```diff
@@ -51,12 +53,14 @@ tests = [
...
-  "ipython"
+  "ipython",
+  "pytest_voluptuous"
 ]
 dev = [
   "dvclive[all,tests]",
   "mypy>=1.1.1",
-  "types-PyYAML"
+  "types-PyYAML",
+  "voluptuous"
```
shcheklein, line 63: move to tests?

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,195 @@
...
+
+    _plot_blacklist_prefix: Tuple = ()
+
+    def __init__(
+        self,
+        interval: float = 0.5,
+        num_samples: int = 10,
+        plot: bool = True,
+    ):
+        if not isinstance(interval, (int, float)):
```
shcheklein, line 47: hmm, I don't think we are input types like this

shcheklein: in this case the idea was to test probably that value are meaningful and raise ValueError? or ignore CPU monitoring

src/dvclive/monitor_system.py
shcheklein: kk, I think it should be fine in this case since we control all these names.

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,195 @@
...
+            / nb_cpus,
+            METRIC_RAM_USAGE_PERCENT: ram_info.percent,
+            METRIC_RAM_USAGE_GB: (ram_info.percent / 100)
+            * (ram_info.total / GIGABYTES_DIVIDER),
+            METRIC_RAM_TOTAL_GB: ram_info.total / GIGABYTES_DIVIDER,
+        }
+        for disk_name, disk_path in self._disks_to_monitor.items():
+            if not Path(disk_path).exists():
+                continue
+            disk_info = psutil.disk_usage(disk_path)
```
shcheklein, line 185: I think it's better to catch OSError in this case instead of the `not Path` check (more reliable way)

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,195 @@
...
+        disks_to_monitor: Optional[Dict[str, str]] = None,
+        plot: bool = True,
+    ):
+        """Monitor CPU resources and log them to DVC Live.
+
+        Args:
+            interval (float): interval in seconds between two measurements.
+                Defaults to 0.5.
+            num_samples (int): number of samples to average. Defaults to 10.
+            disks_to_monitor (Optional[Dict[str, str]]): paths to the disks or
```
shcheklein, line 129: I'm not sure I understand this "aths to the disks or partitions to monitor disk usage statistics.". what is the difference disk vs parittion?

tests/test_monitor_system.py
AlexandreKempf: I wanted to ensure that the mock objects covered all the fields used in the cpu_monitoring. But after your comment, I realized this test was useless since it only tests behavior that is defined in the unit tests. I removed it.

pyproject.toml
AlexandreKempf: yeap I changed it to move it into "tests"

src/dvclive/monitor_system.py
AlexandreKempf: Oh ok I see. 
I changed to keep the monitoring even if the values are out of range, but I added a warning message. 
I took the min and max values from here: https://github.com/wandb/wandb/blob/852781a852a5ae63ea009b40fe923e1a80603fd0/wandb/sdk/internal/system/assets/interfaces.py#L120

src/dvclive/monitor_system.py
AlexandreKempf: I follow the same pattern as the other parameters. If there is an OSError, the program continues, but we won't track the disk usage for the non-existing folder.

tests/test_monitor_system.py
AlexandreKempf: I tried to respect what you explained. If there is still some misunderstanding between us, could you please refer to a link or a concrete example to illustrate your point? This way, I might be able to understand better what you meant :)

src/dvclive/monitor_system.py
AlexandreKempf: I tried to change the description for this arg. Let me know what you think. I mentioned `disk` because I'm afraid our users don't know what a partition is. 

Changes involve:
 - renaming to `folders_to_monitor`. While not perfectly accurate (because two folders on the same partition, that have different content, will have identical statistics with this function), it is the most understandable name I could find. I'm scared that `partition_to_monitor` doesn't speak to our users. 
 - mentioning the perfectly accurate definition in the rest of the docstring so people looking for deeper information find what they need.

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,192 @@
...
+        METRIC_CPU_COUNT,
+        METRIC_RAM_TOTAL_GB,
+        "system/disk/total (GB)",
+    )
+
+    def __init__(
+        self,
+        interval: float = 0.1,
+        num_samples: int = 20,
+        folders_to_monitor: Optional[Dict[str, str]] = None,
```
shcheklein, line 121: minor: I think we are using `directory` usually (instead of the `folder`)

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        super().__init__(interval=interval, num_samples=num_samples, plot=plot)
+        folders_to_monitor = (
+            {"main": "/"} if folders_to_monitor is None else folders_to_monitor
+        )
+        self._disks_to_monitor = {}
+        for disk_name, disk_path in folders_to_monitor.items():
+            if disk_name != os.path.normpath(disk_name):
+                raise ValueError(  # noqa: TRY003
+                    "Keys for `partitions_to_monitor` should be a valid name"
```
shcheklein, line 150: hmm, what is the valid name in this case?

shcheklein: I would still move try catch to where we use ... e.g. someone drops the directory, it should not be causing other metrics to disappear

tests/test_monitor_system.py
```diff
@@ -0,0 +1,219 @@
...
+        # wait for the metrics to be logged.
+        # METRIC_DISK_TOTAL_GB is the last metric to be logged.
+        while len(dpath.search(live.summary, METRIC_DISK_TOTAL_GB)) == 0:
+            time.sleep(0.001)
+        live.next_step()
+
+        _, latest = parse_metrics(live)
+
+    schema = {}
+    for name in [
```
shcheklein, line 159: do we check for actual values in both tests?

AlexandreKempf: I change one test to check for values :)

AlexandreKempf:
@shcheklein @dberenbaum 
Please consider this PR also, because it will probably be merge in this branch before we merge to `main`
https://github.com/iterative/dvclive/pull/785

PR #775, opened 2024-02-07
add toml formatter for pyproject.toml

AlexandreKempf:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,3 +34,8 @@ repos:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
+  - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+    rev: v2.12.0
+    hooks:
+      - id: pretty-format-toml
+        args: [--autofix]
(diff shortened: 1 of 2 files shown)


AlexandreKempf:
Ah! I understand now why it was not using a formatter already.
We cannot use a pre-commit formatter on cookie-cutter projects because the TOML/YAML files are not valid TOML/YAML before the project is initiated. 
We can keep it for this repo, or we can remove it, as you prefer ;)

AlexandreKempf:
@dberenbaum I'm wondering if I get you right. Or maybe I wasn't clear enough in my comment. 
If I add the formatters in the py-template's pre-commit, it will try to pre-commit the py-template project. Or there are plenty of toml files that are not valid for the formatter (stuff like `{%- if cookiecutter.docs != "False" %}`), and at the end, the project refuses to commit my changes. 

Is there something I'm missing? What did you have in mind exactly?

dberenbaum:
It might be my misunderstanding since I haven't tried it. Are you putting it [inside](https://github.com/iterative/py-template/blob/main/%7B%7Bcookiecutter.project_name%7D%7D/.pre-commit-config.yaml) or [outside](https://github.com/iterative/py-template/blob/main/.pre-commit-config.yaml) the cookiecutter? Does it run the pre-commit inside the cookiecutter?

AlexandreKempf:
Oh my mistake! I putted it outside :( Thanks for the explanation. 
I'm doing the PR on `py-template` :)

Issue #766, opened 2024-01-23
`log_image`: log bounding boxes


dberenbaum:
Related: https://github.com/iterative/dvc/issues/10198, https://github.com/iterative/vscode-dvc/issues/4917

We need a way to log bounding boxes (and maybe later other annotations like segmentation masks) for images saved with dvclive.

# p1

The API can look like this:

```python
boxes = [
  {"label": "cat", "box": {"x_min": 100, "x_max": 110, "y_min": 5, "y_max": 20}},
  {"label": "cat", "box": {"x_min": 30, "x_max": 55, "y_min": 75, "y_max": 90}},
  {"label": "dog", "box": {"x_min": 80, "x_max": 100, "y_min": 25, "y_max": 50}}
]
live.log_image("myimg.png", myimg, boxes=boxes)
```

In addition to saving the image to `dvclive/plots/images/myimg.png`, this will also save annotations to `dvclive/plots/images/myimg.json` in the following format:

```json
{"boxes":
  [
    {"label": "cat", "box": {"x_min": 100, "x_max": 110, "y_min": 5, "y_max": 20}},
    {"label": "cat", "box": {"x_min": 30, "x_max": 55, "y_min": 75, "y_max": 90}},
    {"label": "dog", "box": {"x_min": 80, "x_max": 100, "y_min": 25, "y_max": 50}}
  ]
}
```

# p2:
- [ ] Other box formats (using width, height, and x/y for the center/corner) ({"x_center": 100, "y_center": 50, "width": 10, "height": 20})
- [ ] Normalized coordinates (between 0 and 1) instead of pixel coordinates (we could probably auto-detect this)
- [ ] Scores (`"scores": {"acc": 0.9, "loss": 0.05}`) so that users can filter boxes based on thresholds (only show boxes where `acc > 0.8`)
- [ ] Segmentations masks (tbd, requires a class per pixel)

dberenbaum:
May need to consider whether it's necessary to list the universe of labels somewhere or if it's fine to parse them as the set of all individual labels.

AlexandreKempf:
I'm not 100% sure this is the right place for my first discussion on a feature. 
But I'll jump on that one and remove my comment if it is not the correct way of dealing with feature discussion internally. 

[...]

PR #776, opened 2024-02-07
DRAFT: add log_bounding_boxes to Live

AlexandreKempf:
Add the possibility of logging bounding boxes in DVCLive!

Context: 
- https://github.com/iterative/dvc/issues/10198
- https://github.com/iterative/dvclive/issues/766

The current implementation uses the `log_bounding_boxes` function to save a JSON file that should be close to the image and with the same name. The format for the JSON file is:
```json
{
  "boxes": [
    {"label": "cat", "box": {"left": 100, "right": 110, "top": 5, "bottom": 20}, "score": 0.3},
    {"label": "dog", "box": {"left": 200, "right": 210, "top": 100, "bottom": 150}, "score": 0.4},
    {"label": "cat", "box": {"left": 300, "right": 310, "top": 200, "bottom": 300}, "score": 0.5}
    ]
}
```

The JSON format is still under discussion [here](https://github.com/iterative/dvc/issues/10198), so I consider the PR as a draft.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -8,7 +8,7 @@ import os
 import shutil
 import tempfile
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Set, Union, TYPE_CHECKING
+from typing import Any, Dict, List, Literal, Optional, Set, Union, TYPE_CHECKING
 
 if TYPE_CHECKING:
     import numpy as np
@@ -33,7 +33,15 @@ from .error import (
     InvalidPlotTypeError,
     InvalidReportModeError,
 )
-from .plots import PLOT_TYPES, SKLEARN_PLOTS, CustomPlot, Image, Metric, NumpyEncoder
+from .plots import (
+    PLOT_TYPES,
+    SKLEARN_PLOTS,
+    BoundingBoxes,
+    CustomPlot,
+    Image,
+    Metric,
+    NumpyEncoder,
+)
 from .report import BLANK_NOTEBOOK_REPORT, make_report
 from .serialize import dump_json, dump_yaml, load_yaml
 from .studio import get_dvc_studio_config, post_to_studio
@@ -84,6 +92,7 @@ class Live:
         self._step: Optional[int] = None
         self._metrics: Dict[str, Any] = {}
         self._images: Dict[str, Any] = {}
+        self._bounding_boxes: Dict[str, Any] = {}
         self._params: Dict[str, Any] = {}
         self._plots: Dict[str, Any] = {}
         self._artifacts: Dict[str, Dict] = {}
@@ -398,6 +407,26 @@ class Live:
         image.dump(val)
         logger.debug(f"Logged {name}: {val}")
 
+    def log_bounding_boxes(
+        self,
+        name: str,
+        bboxes: Union[List[List[int]], np.ndarray],
+        labels: Union[List[str], np.ndarray],
+        scores: Union[List[float], np.ndarray],
+        format: Literal["tlbr", "tlhw", "xywh"],  # noqa: A002
+    ):
+        BoundingBoxes.could_log(name, bboxes, labels, scores, format)
+
+        if name in self._bounding_boxes:
+            bounding_boxes = self._bounding_boxes[name]
+        else:
+            bounding_boxes = BoundingBoxes(name, self.plots_dir)
+            self._bounding_boxes[name] = bounding_boxes
+
+        bounding_boxes.step = self.step
+        bounding_boxes.dump(bboxes, labels, scores, format)
+        logger.debug(f"Logged {name}: {len(bboxes)} bounding boxes in {format} format")
+
     def log_plot(
         self,
         name: str,
@@ -585,6 +614,10 @@ class Live:
         if self._images and self._cache_images:
             images_path = Path(self.plots_dir) / Image.subfolder
             self.cache(images_path)
+
+        if self._bounding_boxes and self._cache_images:
+            bounding_boxes_path = Path(self.plots_dir) / BoundingBoxes.subfolder
+            self.cache(bounding_boxes_path)
 
         # If next_step called before end, don't want to update step number
         if "step" in self.summary:
--- a/src/dvclive/error.py
+++ b/src/dvclive/error.py
@@ -39,3 +39,11 @@ class InvalidReportModeError(DvcLiveError):
             f"`report` can only be `None`, `auto`, `html`, `notebook` or `md`. "
             f"Got {val} instead."
         )
+
+
+class InvalidSameSizeError(DvcLiveError):
+    def __init__(self, name, x, y):
+        self.name = name
+        self.x = x
+        self.y = y
+        super().__init__(f"Data '{name}': '{x}' and '{y}' must have the same length")
--- a/src/dvclive/plots/__init__.py
+++ b/src/dvclive/plots/__init__.py
@@ -1,5 +1,6 @@
 from .custom import CustomPlot
 from .image import Image
+from .bounding_boxes import BoundingBoxes  # noqa: F401
 from .metric import Metric
 from .sklearn import Calibration, ConfusionMatrix, Det, PrecisionRecall, Roc
 from .utils import NumpyEncoder  # noqa: F401
--- a/src/dvclive/plots/bounding_boxes.py
+++ b/src/dvclive/plots/bounding_boxes.py
@@ -1,0 +1,109 @@
+from typing import List, Literal, Union
+import math
+
+
+import numpy as np
+from pathlib import Path
+
+from dvclive.plots.utils import NumpyEncoder
+from dvclive.serialize import dump_json
+from dvclive.error import InvalidSameSizeError, InvalidDataTypeError
+
+from .base import Data
+
+BboxFormatKind = Literal["tlbr", "tlhw", "xywh"]
+
+
+class BoundingBoxes(Data):
+    suffixes = (".json",)
+    subfolder = "images"
+
+    @property
+    def output_path(self) -> Path:
+        _path = self.output_folder / self.name
+        _path.parent.mkdir(exist_ok=True, parents=True)
+        return _path
+
+    @staticmethod
+    def could_log(
+        name: str,
+        bounding_boxes: Union[List[List[int]], np.ndarray],
+        labels: List[str],
+        scores: List[float],
+        format: BboxFormatKind,  # noqa: A002
+    ) -> bool:
+        if not any(name.endswith(suffix) for suffix in BoundingBoxes.suffixes):
+            raise InvalidDataTypeError(name, type(name))
+
+        if len(bounding_boxes) != len(labels):
+            raise InvalidSameSizeError(name, "bounding_boxes", "labels")
+
+        if len(bounding_boxes) != len(scores):
+            raise InvalidSameSizeError(name, "bounding_boxes", "scores")
+
+        if format not in ["tlbr", "tlhw", "xywh"]:
+            raise InvalidDataTypeError(name, type(format))
+
+        if not all(isinstance(score, float) for score in scores):
+            raise InvalidDataTypeError(name, type(scores))
+
+        if not all(
+            isinstance(x, (int, np.int_)) for bbox in bounding_boxes for x in bbox
+        ):
+            raise InvalidDataTypeError(name, type(bounding_boxes))
+
+        if not all(isinstance(label, str) for label in labels):
+            raise InvalidDataTypeError(name, type(labels))
+        return
+
+    def dump(
+        self,
+        bboxes: Union[List[List[int]], np.ndarray],
+        labels: List[str],
+        scores: List[float],
+        format: BboxFormatKind,  # noqa: A002
+    ) -> None:
+        bboxes = self.convert_to_tlbr(bboxes, format)
+        json_content = {
+            "boxes": [
+                {
+                    "label": label,
+                    "box": {
+                        "top": tlbr[0],
+                        "left": tlbr[1],
+                        "bottom": tlbr[2],
+                        "right": tlbr[3],
+                    },
+                    "score": score,
+                }
+                for tlbr, label, score in zip(bboxes, labels, scores)
+            ]
+        }
+
+        dump_json(json_content, self.output_path, cls=NumpyEncoder)
+
+    @staticmethod
+    def convert_to_tlbr(
+        bboxes: Union[List[List[int]], "np.ndarray"],
+        format: BboxFormatKind,  # noqa: A002
+    ) -> List[List[int]]:
+        """
+        Converts bounding boxes from different formats to the top, left, bottom, right
+        format.
+        """
+        if format == "tlhw":
+            return [
+                [box[0], box[1], box[0] + box[2], box[1] + box[3]] for box in bboxes
+            ]
+
+        if format == "xywh":
+            return [
+                [
+                    box[0] - math.ceil(box[2] / 2),
+                    box[1] - math.ceil(box[3] / 2),
+                    box[0] + math.floor(box[2] / 2),
+                    box[1] + math.floor(box[3] / 2),
+                ]
+                for box in bboxes
+            ]
+        return bboxes
--- a/tests/plots/test_bounding_boxes.py
+++ b/tests/plots/test_bounding_boxes.py
@@ -1,0 +1,257 @@
+import numpy as np
+import pytest
+
+from dvclive import Live
+from dvclive.plots import BoundingBoxes
+from dvclive.error import InvalidDataTypeError, InvalidSameSizeError
+
+
+@pytest.mark.parametrize(
+    ("name", "bounding_boxes", "labels", "scores"),
+    [
+        ("image.json", [], [], []),
+        ("image.json", [[10, 20, 30, 40]], ["A"], [0.7]),
+        ("image.json", [[0, 0, 10, 10], [10, 150, 300, 500]], ["B", "C"], [0.5, 0.8]),
+        (
+            "image.json",
+            np.array([[0, 0, 10, 10], [10, 150, 300, 500]]),
+            ["B", "C"],
+            [0.5, 0.8],
+        ),
+        (
+            "image.json",
+            [[0, 0, 10, 10], [10, 150, 300, 500]],
+            np.array(["B", "C"]),
+            [0.5, 0.8],
+        ),
+        (
+            "image.json",
+            [[0, 0, 10, 10], [10, 150, 300, 500]],
+            ["B", "C"],
+            np.array([0.5, 0.8]),
+        ),
+    ],
+)
+def test_save_json_file(name, bounding_boxes, labels, scores, tmp_dir):
+    live = Live()
+    live.log_bounding_boxes(name, bounding_boxes, labels, scores, format="tlbr")
+    assert (tmp_dir / live.plots_dir / BoundingBoxes.subfolder / name).exists()
+
+
+@pytest.mark.parametrize(
+    ("name", "bounding_boxes", "labels", "scores"),
+    [
+        ("image.png", [], [], []),
+        ("image.jpg", [[10, 20, 30, 40]], ["A"], [0.7]),
+        ("image.yaml", [[0, 0, 10, 10], [10, 150, 300, 500]], ["B", "C"], [0.5, 0.8]),
+    ],
+)
+def test_invalid_extension(name, bounding_boxes, labels, scores, tmp_dir):
+    live = Live()
+    with pytest.raises(InvalidDataTypeError):
+        live.log_bounding_boxes(
+            name,
+            bounding_boxes,
+            labels,
+            scores,
+            format="tlbr",
+        )
+
+
+@pytest.mark.parametrize(
+    ("name", "bounding_boxes", "labels", "scores"),
+    [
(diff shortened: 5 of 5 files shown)


src/dvclive/live.py
```diff
@@ -398,6 +407,26 @@ def log_image(self, name: str, val):
...
         logger.debug(f"Logged {name}: {val}")
 
+    def log_bounding_boxes(
+        self,
+        name: str,
+        bboxes: Union[List[List[int]], np.ndarray],
+        labels: Union[List[str], np.ndarray],
+        scores: Union[List[float], np.ndarray],
+        format: Literal["tlbr", "tlhw", "xywh"],  # noqa: A002
+    ):
```
dberenbaum, line 417: TODO: docstring

dberenbaum:
Thanks @AlexandreKempf! Can you show an example of how to use it?

Why use a separate method rather than make it part of `log_image`? How do we ensure it's connected to the image?

shcheklein (commented):
A suggestion: To try the API I would make it part of the YOLO DVC logger. They have a few images with bounding boxes and there are some example in other loggers (W&B?) that log them. That will give us a good sense about the API.

dberenbaum:
@AlexandreKempf Without having looked deeply through the PR yet, I'm still fuzzy on how we associate the bounding boxes with the images. An example (whether yolo or something else) will go a long way here and help show the pros and cons and we can decide what works.

AlexandreKempf:
@dberenbaum Here is what I had in mind:

### using `log_bounding_boxes`

```python
from ultralytics import YOLO
from dvclive import Live

model = YOLO("yolov8n.pt") 
with live as Live():
    image_path = "https://ultralytics.com/images/bus.jpg"
    image_name = "image_bus"
    live.log_image(image_name, image)
    
    results = model(image)
    
    format = "tlbr"
    bboxes = results[0].boxes.xyxy.numpy()
    classes = results[0].boxes.cls.numpy()
    class_names = [results[0].names[class_index] for class_index in classes]
    scores = results[0].boxes.conf.numpy()
    live.log_bounding_boxes(image_name, boxes, class_names, scores, format=format)
    # or some dict processing then `live.log_bounding_boxes(image_name, image_path, boxes)`
```

### using `log_image`

```python
from ultralytics import YOLO
from dvclive import Live

model = YOLO("yolov8n.pt") 
with live as Live():
    image_path = "https://ultralytics.com/images/bus.jpg"
    image_name = "image_bus"
        
    results = model(image)
    
    format = "tlbr"
    bboxes = results[0].boxes.xyxy.numpy()
    classes = results[0].boxes.cls.numpy()
    class_names = [results[0].names[class_index] for class_index in classes]
    scores = results[0].boxes.conf.numpy()
    live.log_image(image_name, image_path, boxes, class_names, scores, format=format)
    # or some dict processing then `live.log_image(image_name, image_path, boxes)`
```


Image and bounding boxes can be matched by `image_name` in the `Live` object. Once the Python session is over, we can still match them by their path because they should have the same path but different suffixes (like you described [here](https://github.com/iterative/dvc/issues/10198#issuecomment-1889215366))

AlexandreKempf:
Sidenote to @shcheklein. 
From what I understood of the [Yolo W&B logger](https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/callbacks/wb.py), they never log the bboxes and the image. They are just saving images with already bounding boxes on top of them (in the pixels I mean), and ultralitycs construct these images. So technically, we could be using this technique already :)
Also, the ultralytics documentation on [how W&B can display/hide bounding boxes based on their labels](https://docs.ultralytics.com/integrations/weights-biases/#key-features-of-the-weights-biases-dashboard) is probably not accurate since they don't have the bounding boxes information with ultralytics.

shcheklein:
@AlexandreKempf sorry, I meant the Comet ML in this case https://github.com/ultralytics/ultralytics/blob/main/ultralytics/utils/callbacks/comet.py#L220 . It's the most complete logger for YOLO atm AFAIR.

> nd ultralitycs construct these images. So technically, we could be using this technique already :)

we do this already, yep

shcheklein:
They are using this call:

`experiment.log_image(image_path, name=image_path.stem, step=curr_step, annotations=annotation)`

And i like it tbh. It simple, it's clear what is happening. I like also that they are using `annotations` - there is a path to expand it beyond just bounding boxes.

dberenbaum:
> I like also that they are using `annotations` - there is a path to expand it beyond just bounding boxes.

Not a strong opinion, but discussed yesterday with @AlexandreKempf that this approach also has its downsides:
1. There is no way to know what format to include in `annotations` without going to their docs
2. You will have to write additional code to structure your data in that format

AlexandreKempf:
I'm working on that but I won't push until I have something satisfying on the VScode plots.
To keep you updated, I went for a solution that should satisfy all of us:
```
log_image(name, val, bboxes)
```
`bboxes` or `annotations`, we need to pick. It won't change the current PR but it will affect the following ones on segmentation masks. 

The format expected for `bboxes` is 
```
{
    "boxes": [[1,2,3,4], [5,6,7,8],[10,11,12,13]],
    "labels": ["cat", "dog", "boat"]
    "scores": [0.1, 0.3, 0.8]
    "format": "tlbr"
}
```

shcheklein:
> There is no way to know what format to include in annotations without going to their docs

true, but it is the same for bbox - I would have to go to docs to see what is expected

> You will have to write additional code to structure your data in that format

yep, here I have no idea how complicated it is compared to that approach

so, no opinion on my end, just a thing to consider ...

shcheklein:
@AlexandreKempf hey, is it ready to be reviewed? or is it still a draft ? (can we update the title and description please when you think it's ready to be reviewed)

AlexandreKempf:
@shcheklein It is still a draft because I want to ensure it is working E2E. The DVC part is done (no PR created, but it outputs what I want to display the bounding boxes on VScode, at least). Now, we are working with @julieg18 to have a working version from DVClive to VScode.

dberenbaum:
Some questions I'm still unclear on:
1. Is there a working version in VS Code? What about handling this in DVC? How can experiments can be compared in VS Code and Studio with this info?
2. What does a yolo or torchvision example look like with this method?
3. Should we require all fields or should some be optional?

AlexandreKempf:
> Is there a working version in VS Code? What about handling this in DVC? How can experiments can be compared in VS Code and Studio with this info?

We are currently debugging one with @julieg18 but are close to getting something working perfectly well. I'll update a video of the final result by the end of day to demo how it works. Also, I'm going to open the PR in DVC to add the feature code. 

> What does a yolo or torchvision example look like with this method?

torchvision & lightning integration looks like this for the user:
```python
class LightningModule(pl.LightningModule):
    
    # ... define `__init__` and `training_step`
    
    def validation_step(self, batch, batch_idxs):
        imgs, targets = batch
        
        # inference on validation images
        preds = self.forward(imgs)

        # log images with bounding boxes
        if batch_idxs == 0:
            live = self.logger.experiment
            for index, img in enumerate(imgs[:15]):
                prediction = preds[index]
                live.log_image(
                    f"val_images/{index}/{self.current_epoch}.png",
                    convert_image_to_np_array(img),
                    annotations={
                        "boxes": prediction["boxes"].cpu().numpy().astype(int),
                        "labels": [
                            self.class_names[i]
                            for i in prediction["labels"].cpu().numpy()
                        ],
                        "scores": np.around(prediction["scores"].cpu().numpy(), 3),
                        "format": "ltrb",
                    },
                )

        return
```
Note that this saves each image "A", "B", "C" into a structure that looks like this:
```
images/
    A/
        0.png
        0.json
        1.png
        1.json
        ...
    B/
    C/
```
Where 0, 1, ... are the validation number (=epoch number if we do validation at every epoch). It allows the `step` slider in VSCode and helps to see how the model learns.

For yolo integration, for the user it should looks like this:
```python
from ultralytics import YOLO
from dvclive import Live

model = YOLO("yolov8n.pt")
with live as Live():
    image_path = "https://ultralytics.com/images/bus.jpg"
    image_name = "image_bus"

    results = model(image)
 
    # log image with bounding boxes
    format = "tlbr"
    boxes = results[0].boxes.xyxy.numpy()
    labels_idx = results[0].boxes.cls.numpy()
    labels = [results[0].names[idx] for idx in labels_idx]
    scores = results[0].boxes.conf.numpy()
    live.log_image(
        image_name,
        image_path,
        annotations={
            "boxes": boxes,
            "labels": labels,
            "scores": scores,
            "format": format,
        },
    )
```

> Should we require all fields or should some be optional?

I wondered the same thing. I guess the first iteration should have all the fields needed, and we can always remove some hard constraints as we move on. It is way easier to do it this way than the other way around. If we start with optional fields and realize it is a mistake, it will be harder to revert and force the field. 
Options I could see, from useful to less useful (IMHO):
- "format" could have a default and be optional
- "score" could have a default to 1.0 for ground truth (target boxes in the dataset provided by human annotations)
- "labels" could be optional for use case where there is only one label and the user doesn't want to see the label on each boxes displayed.

src/dvclive/live.py
```diff
@@ -457,6 +465,7 @@ def log_image(
         self,
         name: str,
         val: Union[np.ndarray, matplotlib.figure.Figure, PIL.Image, StrPath],
+        annotations: Optional[Dict] = None,
```
dberenbaum, line 468: Should we include a more specific type here to generate better hints as we discussed @AlexandreKempf?

We also need to update the docstring to include `annotations`. Is it too much to fit the structure of the annotations into the docstring?

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+        boxes_and_labels_same_size = len(annotations[BOXES_NAME]) == len(
+            annotations[LABELS_NAME]
+        )
+        boxes_and_scores_same_size = len(annotations[BOXES_NAME]) == len(
+            annotations[SCORES_NAME]
+        )
+        if not boxes_and_labels_same_size or not boxes_and_scores_same_size:
+            logger.warning(
+                f"'{BOXES_NAME}', '{LABELS_NAME}', and '{SCORES_NAME}' should have the "
+                "same size."
```
dberenbaum, line 59: Let's make it more clear that the annotations will not be logged in this case.

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+            )
+            return False
+
+        # `format` should be one of the supported formats
+        if annotations[FORMAT_NAME] not in get_args(BboxFormatKind):
+            logger.warning(
+                f"Annotations format '{annotations['format']}' is not supported."
```
dberenbaum, line 66: Suggest correct formats or link to docs?

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+            return False
+
+        # `scores` should be a List[float]
+        if not all(
+            isinstance(score, (float, np.floating))
+            for score in annotations[SCORES_NAME]
+        ):
+            logger.warning(
+                "Annotations `'scores'` should be a `List[float]`, received "
+                f"'{annotations[SCORES_NAME]}'."
```
dberenbaum, line 77: Same as above - let's make clear annotations aren't being logged.

dberenbaum: And same for other warning below

src/dvclive/plots/annotations.py
dberenbaum: Never mind, I missed how this is used

src/dvclive/live.py
```diff
@@ -506,13 +515,23 @@ def log_image(
...
             image = Image(name, self.plots_dir)
-            self._images[name] = image
+            self._images[name] = {"image": image}
 
         image.step = self.step
         image.dump(val)
+
+        if annotations:
+            if not Annotations.could_log(annotations):
+                raise InvalidDataTypeError(name, type(annotations))
```
dberenbaum, line 528: Do you think we should actually raise an error? I know that's what we otherwise do if the image format is invalid, but we also tend to be cautious in raising errors in dvclive since it could potentially interrupt more important and long-running model training operations.

src/dvclive/live.py
```diff
@@ -506,13 +515,23 @@ def log_image(
...
+
+            annotation_object = Annotations(name, self.plots_dir)
+            self._images[name]["annotations"] = annotation_object
+            annotation_object.step = self.step
```
dberenbaum, line 532: Is this used somewhere?

dberenbaum:
I'm fine to move forward with this approach, but let's document the pros and cons once more so we can easily review the thought process in the future:

> ### pros of using `log_bounding_boxes`
> 
>     * I'm afraid we will flood the `load_image` with arguments. First, the bounding boxes, then the polygons/mask representation, then the segmentation... each takes a large `json` like structure and additional information on how to parse it (the `format="tlbr"` for instance).

This is mitigated by using a catchall `annotations` kwarg.
 
>     * In many cases, you want to see object detection on the validation set (a fixed set of images that are not augmented). This means that you want to see the same image, but you want to see how the bounding boxes evolved as the training goes. With the slider interface provided by our front end, I wanted us to be able to show the different epochs for the same image. That would be a nice feature for object detection users. If we use `log_image`  for that, we'll have to save the same image several times. Using a second method, we are free of this problem. It is not a strong argument, as we could save the image every time also and it will still work.

I don't see an easy way to do this regardless of which method we use, but maybe I'm missing something. We would need some way to capture bounding boxes per step.

>     * It will simplify documentation for the bounding box logging. It will be on another page, so it will be easier to read.

This is still a concern that we can revisit once @AlexandreKempf has drafted a docs PR.

> * There is no way to know what format to include in `annotations` without going to their docs

I think this would be easier in `log_bbox()` since the IDE could show the individual kwargs for `boxes`, `labels`, etc. We can partially mitigate this with types and docstrings (see [here](https://github.com/iterative/dvclive/pull/776#discussion_r1496331047)).

> * You will have to write additional code to structure your data in that format

Doesn't look like it makes much difference.

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+        _path.parent.mkdir(exist_ok=True, parents=True)
+        return _path
+
+    @staticmethod
+    def could_log(  # noqa: PLR0911
+        annotations: Dict[str, List],
+    ) -> bool:
+        # no missing fields
+        if any(
+            field not in annotations
```
shcheklein, line 40: can we use some schema parser like https://pypi.org/project/voluptuous/ ? do other project validate the schema?

shcheklein: @skshetry what are the options for validating a schema like this with some meaningful error and preferably with the least amount of dependenices?

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+                    "top": tlbr[0],
+                    "left": tlbr[1],
+                    "bottom": tlbr[2],
+                    "right": tlbr[3],
+                },
+                "score": score,
+            }
+            for tlbr, label, score in zip(boxes, labels, scores)
+        ]
+        # format for VScode and Studio
```
shcheklein, line 125: why is it about Studio and VS Code?

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,166 @@
...
+        dump_json(
+            {"annotations": boxes_json},
+            self.output_path.with_suffix(".json"),
+            cls=NumpyEncoder,
+        )
+
+    @staticmethod
+    def convert_to_tlbr(
+        bboxes: Union[List[List[int]], "np.ndarray"],
+        format: BboxFormatKind,  # noqa: A002
```
shcheklein, line 142: why A002 - can we cleanup it?

src/dvclive/plots/annotations.py
skshetry: `voluptuous` is not being maintained but has no dependencies.

I'd use `pydantic` as it works with type-hints and has dataclass-like interface and is more popular and has good DX.

Both `pydantic` and `voluptuous` are dependencies/transitive-dependencies of dvc, so I'd not worry about dependencies.

src/dvclive/live.py
AlexandreKempf: :+1:  I added the docstring, you were right, it was missing. Tell me if this is too much. I feel like the length of the docstring is fine. 

For better typing, I looked at several options and I picked the one that I preferred:
- Pydantic, data class, attrs ... Not a huge fan since it introduces a new object the user needs to import and use. ex: `log_image(path, image, annotations=Annotations(boxes= ..., labels=..., scores=..., format=....))`. If we are OK with that, it is probably the most elegant solution otherwise 
- TypedDict all the way. I tried to replace all the annotations dictionaries in `plots/annotations.py`. If we do that, mypy refuses to use constant names (`BOXES_NAMES,` `LABELS_NAME,`...) to access the typedict keys. It is not a problem, we could use "boxes" instead of a constant, and mypy would warn use if we try to access a field that is not valid. The problem is that if we change the name of a field, it is slightly annoying to replace the field everywhere (main code + log warning + tests). 
- **TypedDict only for the user (in `live.py`) -> the option I picked. A duplicated schema exists between what is written in `live.py`  and `annotations.py`. That is one downside, but I think it is the least annoying downside of all.**
- Dict as it was during your comment. The problem as you mention is that it is a bit vague for the user that uses it for the first time. 

What do you think? Do you see any other option?

src/dvclive/live.py
AlexandreKempf: Good point.

From my experience there is no good answer. I see two mood for datascientists:
 - the dev mood, where you're working on the training pipeline. You're launching small experiments with very fast iterations to fix problems on your data/training pipeline. In that case you want the training to crash and raise an error.
 - the production mood where you're working on a model and you don't want to touch the data/training pipeline. You're just changing parameters and running longs (hours/days) experiments. In this setup, I would be mad that the training crash at the end because my logger received a float instead of an int. This is quite easy to catch on the "dev mood" but there is some bugs that occurs very rarely like : the model didn't predict any object on the image, the bounding box have negative coordinates... In the production mode, I would like the logger to raise a warning. 
 
Based on this experience, and if we need to make a choice, I prefer DVClive to always behave in a production mood because this is where you have money at stake. 
There is only one downside about warning: datascientists don't read them. Their pytorch/tensorflow framework is already spamming with logs. Useful logs (for data scientists) like epoch number, the time it takes to complete an epoch, loss values...  But also a LOT of useless logs (for datascientists) about optimization using this env or this, deprecated function that one of your lib is using. Now if you had the logs cuda and co. are adding to optimize this kernel function... You get ~50 to 100 lines at the beginning of every training that nobody reads.
So warning is cool, but I doubt people will look at the logs to see our warning during training.

One cool improvement I could see is to use the `end` method to make a recap:
 - you could have done this better using this function
 - on this step, for this image, the coordinates were negative. Here is what we received, and here is what we expect...
 These types of improvements would be useful and looked at if they are a the end of the log file and well formatted.

src/dvclive/live.py
AlexandreKempf: No, it is inherited from the parent object so I figured it was nice to keep it updated with `live.step`. This way future developers will have the expected behavior.
Also I follow the pattern found for the image in `log_image`. The `step` value for an image is only used for `send_to_studio` I believe so that we don't send the image twice. So I believe the `annotation.step` is going to be used for the same reason. 

One other solution is to bring "step" at the same level as "image" and "annotations" in the `self._images` variable.
Today it looks like this:
```json
self._images = {"path": {"image": <ImageObject with a step property>, "annotations":  <AnnotationsObject with a step property>}, "path2": ...}
```

src/dvclive/live.py
AlexandreKempf: I'll be using pydantic

src/dvclive/live.py
dberenbaum: Yes, 100% agree with what you wrote about how data scientists work. Anyway, fine to keep as is and revisit if someone complains.

AlexandreKempf:
@dberenbaum @shcheklein @skshetry 

I used Pydantic at the end to the validation of user inputs. I would love to have this behavior:
```python
class BBox(BaseModel):
    boxes: ...
    labels: ...
    scores: ...
    box_format: ... 
   
def my_sexy_function(bbox: BBox):
    ...
```

and call it with a dict

```python
my_sexy_function({"boxes":..., "labels":..., "scores":...., "box_format": ...}) 
```

But I'm not sure it is possible with mypy.

So to stay consistent with what we said @dberenbaum, I created a typeddict so that users can see what fields are needed and their type. If they are still making a mistakes, pydantic errors should be enough to guide them through the perfect input.
I realize it is a bit ugly to have both the TypedDict and the pydantic model. I strongly believe that DVClive should improve the user feeling (more understandable errors and warning, better types and docstrings ...) even if it comes with a maintenance cost on our end.

shcheklein (approved):

AlexandreKempf:
Note: on the latest implementation `Annotations.could_log` is not called. I'm thinking of a better way to integrate it to the code.

src/dvclive/plots/annotations.py
```diff
@@ -0,0 +1,131 @@
+from typing import List, Literal, TYPE_CHECKING
+import math
+
+
+from pathlib import Path
+from pydantic import BaseModel, PydanticUserError, field_validator, model_validator
```
skshetry, line 6: Please add pydantic to the `dependencies`.

PR #781, opened 2024-02-12
check huggingface failure against local

mattseddon:
- [ ] ❗ I have followed the [Contributing to DVCLive](https://github.com/iterative/dvclive/blob/main/CONTRIBUTING.rst) guide.


src/dvclive/huggingface.py
```diff
@@ -45,7 +45,11 @@ def on_log(
         **kwargs,
     ):
         logs = kwargs["logs"]
+        logged: dict[str, str | float] = {}
         for key, value in logs.items():
+            if key in logged and logged[key] == value:
```
mattseddon, line 50: [F] This is papering over a crack in the integration where we are basically logging the same metric twice per step.

See https://github.com/iterative/dvclive/actions/runs/7866535786/job/21460950002 / https://github.com/iterative/dvclive/actions/runs/7866443198/job/21460718675 for examples of this.

There may be a more "correct" way to fix the integration but I do not know the transformers library enough to dig into this more.

src/dvclive/huggingface.py
```diff
@@ -45,7 +45,12 @@ def on_log(
         **kwargs,
     ):
         logs = kwargs["logs"]
+        logged: dict[str, Union[int, float, str]] = {}
         for key, value in logs.items():
+            is_duplicated_entry = key in logged and logged[key] == value
+            if is_duplicated_entry:
+                continue
```
mattseddon, line 52: [F] This is papering over a crack in the integration where we are basically logging the same metric twice per step.

See https://github.com/iterative/dvclive/actions/runs/7866535786/job/21460950002 / https://github.com/iterative/dvclive/actions/runs/7866443198/job/21460718675 for examples of this.

There may be a more "correct" way to fix the integration but I do not know the transformers library enough to dig into this more.

tests/frameworks/test_huggingface.py
```diff
@@ -100,33 +100,35 @@ def args():
         "foo",
         evaluation_strategy="epoch",
         num_train_epochs=2,
-        save_strategy="epoch",
+        save_strategy="no",
+        logging_strategy="no",
```
mattseddon, line 104: [F] [Defaults to steps](https://huggingface.co/docs/transformers/v4.37.2/en/main_classes/trainer#transformers.TrainingArguments.logging_strategy). Which I think means that we would be logging multiple times per epoch... which is unexpected.

tests/frameworks/test_huggingface.py
```diff
@@ -100,33 +100,35 @@ def args():
...
     spy.assert_called_once()
 
-    live = callback.live
-    assert os.path.exists(live.dir)
+    assert os.path.exists(callback.live.dir)
 
     logs, _ = parse_metrics(live)
 
-    assert len(logs) == 10
+    assert len(logs) == 11
```
mattseddon, line 128: [F] `total_flos` has been added to the metric list produced by `Trainer`: [total_flos](https://huggingface.co/docs/transformers/en/main_classes/callback#transformers.TrainerState.total_flos)

mattseddon:
My current theory is that a second instance of `DVCLiveCallback` is being attached to the `Trainer`.

tests/frameworks/test_huggingface.py
```diff
@@ -101,6 +101,7 @@ def args():
         evaluation_strategy="epoch",
         num_train_epochs=2,
         save_strategy="epoch",
+        report_to="none",
```
mattseddon, line 104: Docs are [here](https://huggingface.co/docs/transformers/main/en/main_classes/trainer#transformers.TrainingArguments.set_logging.report_to). Having this set to the default of `"all"` seems to create a second instance of the `DVCLiveCallback` which causes duplicate entries to be logged in all of the metric files.

@dberenbaum is this something we need to be aware of for #740?

mattseddon: [total_flos](https://huggingface.co/docs/transformers/en/main_classes/callback#transformers.TrainerState.total_flos) was also added as a metric before `report_to` is set to `"none"` (original failure)

dberenbaum (approved):

Issue #81, opened 2021-05-28
logger: track system metrics automatically


aguschin:
Right now dvclive doesn't track system metrics, to name a few, CPU, GPU, RAM utilisation. This is useful when comparing experiment results (e.g., I can get +1% accuracy, but what is the price in terms of time/resources?) and analysing how long it takes to run training on different GPUs (e.g., I can rent another GPU model and get 2x speedup?).

The usual practice is to log this somehow manually with https://github.com/giampaolo/psutil and analyse the results later, but because heavy experiments in ML require this quite often, IMO it makes sense to have this functionality out-of-the-box.

Also it would be great to have summary on these metrics in .json file produced in experiment to make quick decisions instead of diving too deep (e.g., the average CPU utilisation was 4 cores, my script doesn't utilise all 32 cores I have; the peak RAM utilisation was 8GB, that means I can rent a smaller server on aws to run this training; etc).

To name one example among ml tools, this is already tracked with W&B, see the bottom of dashboard: https://wandb.ai/stacey/estuary?workspace=user-lavanyashukla

This page states main metrics logged in W&B https://docs.wandb.ai/ref/app/features/system-metrics

If this will be useful, I could gather a list of metrics with notes describing the cases these metrics are helpful to a user in ml tasks.

pared:
Well, currently summary is just a gathering of lastly logged metrics. It seems that those system metrics would provide a lot of information that some users might find useful. 
Some questions that needs answering before proceeding implementation:
1. How psutil gathers the data? If its from current point of time, we might get untrustworthy results (since logging usually happens in the end of epoch.
2. How to store the summary? Make `summary.json` just log all values? Or update the structure, to contain, for example "metrics", "system_performace_metrics" subdicts?

dberenbaum:
>     1. How psutil gathers the data? If its from current point of time, we might get untrustworthy results (since logging usually happens in the end of epoch.

This is in the docs:

> wandb automatically logs system metrics every 2 seconds, averaged over a 30 second period.

pared:
We could probably achieve that by creating `dvclive` context that would start "system observer" thread.

daavoo:
This open source repo has a "hardware monitoring" functionality implemented with `psutil` (and `py3nvml` for GPUs):

https://github.com/labmlai/labml/blob/master/guides/hardware_monitoring.md

https://github.com/labmlai/labml/search?q=psutil

daavoo:
[...]

PR #785, opened 2024-02-14
monitor GPU ressources

AlexandreKempf:
# New feature: monitoring for harware metrics

For a first discussion on this PR content, you can look at [here](https://github.com/iterative/dvclive/pull/773). It was a outdated version of this PR that only contains CPU, ram and disk metrics.

### Monitoring hardware

In this PR we add the possibility for the user to monitor the GPU, CPU, RAM, and disk during one experiment.
The GPU metrics are only collected if at least one GPU is detected.

To use this feature you can use it with a simple argument:
```python
from dvclive import Live

with Live(monitor_system=True) as live:
    ...
```

If you want to use advance features you can specify each parameter this way:
```python
from dvclive import Live
from dvclive.monitor_system import SystemMonitor

with Live() as live:
    live.system_monitor = SystemMonitor(interval = 0.1, num_samples=15, directories_to_monitor={"data": "/path/to/data/directory", "home": "/home"}))
```

If you allow the monitoring of your system, if will track the following:
 - `system/cpu/count` -> number of CPU cores
 - `system/cpu/usage (%)` ->  the average usage of the CPUs. 
 - `system/cpu/parallelization (%)` -> How many CPU cores use more than 20% of their possibilities? It is useful when you're looking to parallelize your code to train your model or process your data faster. 
  - `system/ram/usage (%)` -> percentage of the RAM used. Useful to increase batch size or data processed at the same time in the RAM.
  - `system/ram/usage (GB)` -> RAM used. Useful to increase batch size or data processed at the same time. 
  - `system/ram/total (GB)` -> Total RAM in your system
  - `system/disk/usage (%)` -> Amount of disk used by the partition that contain the given path, in %. By default uses "/". You can specify the paths to the partition you want to monitor. For instance, the code example above monitors `/data` and `/home`. Data and code often live in very different paths/volumes, so it is useful for the user to be able to specify its own path. 
  - `system/disk/usage (GB)` ->  Amount of disk used at a given path.
  - `system/disk/total (GB)`  -> Total disk storage at a given path.
  - `system/gpu/count` -> Number of GPUs detected.
  - `system/gpu/usage (%)` -> Usage of each GPU in %.
  - `system/vram/usage (%)` -> Usage of each GPU virtual memory in %.
  - `system/vram/usage (GB)` -> Usage of each GPU virtual memory in GB.
  - `system/vram/total (GB)` -> total amount of GPU virtual memory in GB, for each GPU.

Note that as several paths can be specified, the full metric name is `system/disk/usage (%)/<user defined name>`. For instance it would be `system/disk/usage (%)/data` for the `/path/to/data/disk` and `system/disk/usage (%)/home` for `/home`.

Note that as several GPUs can be detected, the full metric name for GPU metrics (except `count`) is suffix with `/<idx>` that indicate the index of the GPU. Example: `system/gpu/usage (%)/0`

All the values that can change during an experiment can be saved as plots. Timestamps are automatically recorded with the metrics. Other metrics (that don't change) such as GPU count, GPU vram total, CPU count, RAM total and disk total are saved as metrics but cannot be saved as plots. 

I decided to split the usage in % and GB. First, because it is more consistent with the other loggers out there. Second, both are extremely relevant based on which cloud instance you run your experiment. If you only run your experiment on the same hardware, the distinction is not really interesting.

### Files generated
The metrics about the CPU are stored with the `log_metric` function. It means that the `.tsv` files are stored in the `dvclive/plots` folder. A specific folder, `system`, contains all the metrics about the CPU to distinguish them from the user-defined metrics. The metrics are also saved in the `dvclive/metrics.json` file. 

### Plot display
Here is what VScode extension looks like:
image

Here is what Studio looks like:
https://studio.iterative.ai/user/AlexandreKempf/projects/image_classification-imzssxc5ew

Note that studio live update is a little buggy, but it is fixed in [this PR](https://github.com/iterative/dvclive/pull/788)

Note that we're calling `nvmlShutdown` and `nvmlInit` at each fetch of the GPU metrics like in [labML code](https://github.com/labmlai/labml/blob/6dc2a2ad3bd3a46c5dab9be80f3030c5ac8cb133/client/labml/internal/computer/monitor/scanner.py#L141)
---

- [x] 📖 If this PR requires [documentation](https://dvc.org/doc) updates, I have created a separate PR (or issue, at least) in [dvc.org](https://github.com/iterative/dvc.org) and linked it here: https://github.com/iterative/dvc.org/pull/5138

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ lgbm = ["lightgbm"]
 huggingface = ["transformers", "datasets"]
 catalyst = ["catalyst>22"]
 fastai = ["fastai"]
-lightning = ["lightning>=2.0", "torch"]
+lightning = ["lightning>=2.0", "torch", "jsonargparse[signatures]>=4.26.1"]
 optuna = ["optuna"]
 all = [
    "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,lightning,optuna,plots,markdown]"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
   "ruamel.yaml",
   "scmrepo",
   "psutil",
-  "py3nvml"
+  "pynvml"
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,11 @@ tests = [
     "dvclive[image,plots,markdown]",
     "ipython",
 ]
-dev = ["dvclive[all,tests]", "mypy>=1.1.1", "types-PyYAML"]
+dev = [
+    "dvclive[all,tests]",
+    "mypy>=1.1.1",
+    "types-PyYAML",
+]
 mmcv = ["mmcv"]
 tf = ["tensorflow"]
 xgb = ["xgboost"]
@@ -65,7 +69,7 @@ fastai = ["fastai"]
 lightning = ["lightning>=2.0", "torch", "jsonargparse[signatures]>=4.26.1"]
 optuna = ["optuna"]
 all = [
-    "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,lightning,optuna,plots,markdown]",
+   "dvclive[image,mmcv,tf,xgb,lgbm,huggingface,catalyst,fastai,lightning,optuna,plots,markdown]"
 ]
 
 [project.urls]
@@ -133,54 +137,12 @@ ignore-words-list = "fpr"
 
 [tool.ruff]
 ignore = ["N818", "UP006", "UP007", "UP035", "UP038", "B905", "PGH003"]
-select = [
-    "F",
-    "E",
-    "W",
-    "C90",
-    "N",
-    "UP",
-    "YTT",
-    "S",
-    "BLE",
-    "B",
-    "A",
-    "C4",
-    "T10",
-    "EXE",
-    "ISC",
-    "INP",
-    "PIE",
-    "T20",
-    "PT",
-    "Q",
-    "RSE",
-    "RET",
-    "SLF",
-    "SIM",
-    "TID",
-    "TCH",
-    "INT",
-    "ARG",
-    "PGH",
-    "PL",
-    "TRY",
-    "NPY",
-    "RUF",
-]
+select = ["F", "E", "W", "C90", "N", "UP", "YTT", "S", "BLE", "B", "A", "C4", "T10", "EXE", "ISC", "INP", "PIE", "T20", "PT", "Q", "RSE", "RET", "SLF", "SIM", "TID", "TCH", "INT", "ARG", "PGH", "PL", "TRY", "NPY", "RUF"]
 
 [tool.ruff.per-file-ignores]
 "noxfile.py" = ["D", "PTH"]
-"tests/*" = [
-    "S101",
-    "INP001",
-    "SLF001",
-    "ARG001",
-    "ARG002",
-    "ARG005",
-    "PLR2004",
-    "NPY002",
-]
+"tests/*" = ["S101", "INP001", "SLF001", "ARG001", "ARG002", "ARG005", "PLR2004", "NPY002"]
 
 [tool.ruff.pylint]
 max-args = 10
+
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,38 +7,38 @@ description = "Experiments logger for ML projects."
 name = "dvclive"
 readme = "README.md"
 keywords = [
-    "ai",
-    "metrics",
-    "collaboration",
-    "data-science",
-    "data-version-control",
-    "developer-tools",
-    "git",
-    "machine-learning",
-    "reproducibility",
+  "ai",
+  "metrics",
+  "collaboration",
+  "data-science",
+  "data-version-control",
+  "developer-tools",
+  "git",
+  "machine-learning",
+  "reproducibility"
 ]
-license = { text = "Apache License 2.0" }
-maintainers = [{ name = "Iterative", email = "support@dvc.org" }]
-authors = [{ name = "Iterative", email = "support@dvc.org" }]
-requires-python = ">=3.8"
+license = {text = "Apache License 2.0"}
+maintainers = [{name = "Iterative", email = "support@dvc.org"}]
+authors = [{name = "Iterative", email = "support@dvc.org"}]
+requires-python = ">=3.9"
 classifiers = [
-    "Development Status :: 4 - Beta",
-    "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.8",
-    "Programming Language :: Python :: 3.9",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
+  "Development Status :: 4 - Beta",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12"
 ]
 dynamic = ["version"]
 dependencies = [
-    "dvc>=3.33.3",
-    "dvc-render>=1.0.0,<2",
-    "dvc-studio-client>=0.17.1,<1",
-    "funcy",
-    "gto",
-    "ruamel.yaml",
-    "scmrepo",
-    "psutil",
+  "dvc>=3.33.3",
+  "dvc-render>=1.0.0,<2",
+  "dvc-studio-client>=0.17.1,<1",
+  "funcy",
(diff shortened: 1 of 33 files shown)


pyproject.toml
```diff
@@ -39,6 +39,7 @@ dependencies = [
     "ruamel.yaml",
     "scmrepo",
     "psutil",
+    "py3nvml"
```
dberenbaum, line 42: It looks like this is quite old an unmaintained. Any reason not to use `pynvml` from https://pypi.org/project/nvidia-ml-py/? Also, we need to decide if we always install this or make it optional.

pyproject.toml
```diff
@@ -38,7 +38,8 @@ dependencies = [
   "gto",
   "ruamel.yaml",
   "scmrepo",
-  "psutil"
+  "psutil",
+  "py3nvml"
```
dberenbaum, line 42: Reminder to switch to pynvml

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,265 @@
...
+                    percent
+                    for percent in cpus_percent
+                    if percent >= MINIMUM_CPU_USAGE_TO_BE_ACTIVE
+                ]
+            )
+            * 100
+            / num_cpus,
+        }
+
+    def _get_disk_info(self) -> Dict[str, Union[float, int]]:
```
dberenbaum, line 208: If this is causing additional complications, I think we could drop it for now

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,265 @@
...
+                f"{METRIC_DISK_USAGE_GB}/{disk_name}": disk_info.used
+                / GIGABYTES_DIVIDER,
+                f"{METRIC_DISK_TOTAL_GB}/{disk_name}": disk_info.total
+                / GIGABYTES_DIVIDER,
+            }
+            disk_metrics = {k.rstrip("/"): v for k, v in disk_metrics.items()}
+            result.update(disk_metrics)
+        return result
+
+    def __get_gpu_info(self) -> Dict[str, Union[float, int]]:
```
dberenbaum, line 231: Why the double underscore in `__get_gpu_info`?

pyproject.toml
```diff
@@ -37,7 +37,9 @@ dependencies = [
   "funcy",
   "gto",
   "ruamel.yaml",
-  "scmrepo"
+  "scmrepo",
+  "psutil",
+  "pynvml"
```
dberenbaum, line 42: Should we include by default or make them optional?

src/dvclive/monitor_system.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def __call__(self, live):
+        self._live = live
+        self._shutdown_event = Event()
+        Thread(
+            target=self._monitoring_loop,
+        ).start()
+
+    def _monitoring_loop(self):
+        while not self._shutdown_event.is_set():
```
dberenbaum, line 144: Can we put the rest of this logic into its own method? That way, it should be easier to test the underlying logic without having to create a thread that has to be shut down.

shcheklein (approved):
Seems good to me. Thanks @AlexandreKempf !

src/dvclive/monitor_system.py
AlexandreKempf: What complication do you see? :)

dberenbaum: I know there was discussion in the other PR like how to handle partitions. No problem if it's no longer an issue.

dberenbaum:
Hey @AlexandreKempf, take a look at my remaining questions/suggestions, but if @shcheklein is satisfied from a technical perspective, feel free to merge when you feel it's ready.

pyproject.toml
shcheklein: Yep, my 2cs. I would prefer "system" to be enabled by default.  Agreed to check on how big / complicated those deps are.

src/dvclive/monitor_system.py
AlexandreKempf: I'm not quite sure I understand what you mean here :) 
There is lots of logic with the _shutdown_event inside the following lines, how should I handle them?

pyproject.toml
AlexandreKempf: Both, when installed together, are quite light: ~500KB. 

@shcheklein Then should I change the `monitor_system` default to `True` in the `Live` object?
Any idea on that topic @dberenbaum?

My 0.02$ is that it would be nice to have it by default, especially for new users. But for existing users, it can be annoying to have new plots all over the studio dashboard I spend days customizing. What should we prioritize?

shcheklein: > Then should I change the monitor_system default to True in the Live object?

yep, I would go for it. I think we should be optimizing for new users still. Not a strong opinion though.

src/dvclive/monitor_system.py
dberenbaum: Okay, sorry, didn't read/think through it carefully. If the tests are all stable now, no concerns from me.

pyproject.toml
dberenbaum: I would vote to keep it off and consider making it a default in 4.0 for the reasons @AlexandreKempf mentioned (annoyance to existing users)

shcheklein: > off and consider making it a default in 4.0

sounds good. let's make a ticket for this? :) and do it soon I hope. (I just feel it's a really good and appealing feature - people will stick better with the extension, Studio, etc).

shcheklein:
> but if @shcheklein is satisfied from a technical perspective, feel free to merge when you feel it's ready.

I've approved it already!

pyproject.toml
AlexandreKempf: Let's wait for 4.0. Do we have a roadmap to 4.0 by any chance, so I can add that ticket?
Also this ticket should mention that `psutil` and `pynvml` should not be optional anymore.

AlexandreKempf:
I added the doc [here](https://github.com/iterative/dvc.org/pull/5138) fyi.

pyproject.toml
dberenbaum: If we make the deps optional, do we need to change the PR a bit to make sure we fail gracefully if they are not installed? It looks like loading `dvclive.live` now will try to import those deps.

AlexandreKempf: Ok, I put them by default as they are quite light. Also, having them installed by default should simplify the documentation.

AlexandreKempf:
Added changes based on recommendations in the [documentation PR](https://github.com/iterative/dvc.org/pull/5138)

- install `psutil` and `pynvml` by default
- removed `plot` argument since we always want these metrics plotted
- add method `monitor_system` to `Live` instead of a property|

src/dvclive/live.py
```diff
@@ -370,6 +380,43 @@ def step(self, value: int) -> None:
...
+        directories_to_monitor: Optional[Dict[str, str]] = None,
+    ) -> None:
+        """Monitor GPU, CPU, ram, and disk resources and log them to DVC Live.
+
+        Args:
+            interval (float): the time interval between samples in seconds. To keep the
+                sampling interval small, the maximum value allowed is 0.1 seconds.
+                Default to 0.05.
+            num_samples (int): the number of samples to collect before the aggregation.
+                The value should be between 1 and 30 samples. Default to 20.
```
dberenbaum, line 396: Why between 1 and 30?

dberenbaum (approved):
I left a couple questions, but they aren't blockers, so feel free to merge when ready.

src/dvclive/live.py
```diff
@@ -370,6 +380,43 @@ def step(self, value: int) -> None:
         self._step = value
         logger.debug(f"Step: {self.step}")
 
+    def monitor_system(
+        self,
+        interval: float = 0.05,  # seconds
+        num_samples: int = 20,
+        directories_to_monitor: Optional[Dict[str, str]] = None,
```
dberenbaum, line 387: Why not default to `{"main": "/"}`?

src/dvclive/live.py
```diff
@@ -370,6 +380,43 @@ def step(self, value: int) -> None:
...
+    ) -> None:
+        """Monitor GPU, CPU, ram, and disk resources and log them to DVC Live.
+
+        Args:
+            interval (float): the time interval between samples in seconds. To keep the
+                sampling interval small, the maximum value allowed is 0.1 seconds.
```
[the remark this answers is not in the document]
AlexandreKempf: Following W&B behavior on this one.

src/dvclive/live.py
AlexandreKempf: Following W&B behavior on this one.

src/dvclive/live.py
AlexandreKempf: I think it is really bad practice to default a dict (or any mutable structure) :)

mattseddon:
Should this have closed #81?

PR #788, opened 2024-02-15
Add internal counters to send only the latest datapoints to studio

AlexandreKempf:
Following a bug detected during [this feature](https://github.com/iterative/dvclive/pull/773) development.

Using the `step` value to send the latest data to Studio can lead to weird behavior because the `step` is poorly defined in some loggers (eg. pytorch lightning logger). Because the step definition is poorly defined in lightning we used a hack to ensure the `log_metrics` calls by the lightning trainer were correct. But calling `log_metrics` from outside the lightning trainer (a separate thread for instance) leads to data not being sent to Studio or duplicates data. 

This PR introduces a counter for each metric that increments when Studio receives the data points. Instead of using the `step` property as a proxy for which data has been sent to studio, we literally count them now. This way, when we want to send data points to Studio, we can only send the points it hasn't received yet.

The test added in the PR fails in the `main` branch because `logs[test_metric]` is 
```python
[
    {'step': '0', 'test': '0.5'}, 
    {'step': '1', 'test': '0.5'}, 
    {'step': '2', 'test': '0.5'}, 
    {'step': '3', 'test': '0.5'}
]
```
which is expected. But `test_calls` is 
```python
[
    {'step': '0', 'test': '0.5'}, 
    {'step': '1', 'test': '0.5'},
    {'step': '1', 'test': '0.5'}, 
    {'step': '1', 'test': '0.5'},
    {'step': '2', 'test': '0.5'}, 
    {'step': '3', 'test': '0.5'},
    {'step': '3', 'test': '0.5'}
]
```

With this PR, both values `logs[test_metric]` and `test_calls` are the same and don't contain duplicated elements.

### DIFF
--- a/src/dvclive/studio.py
+++ b/src/dvclive/studio.py
@@ -13,10 +13,6 @@ from dvclive.utils import parse_metrics, rel_path
 logger = logging.getLogger("dvclive")
 
 
-def _get_unsent_datapoints(plot, latest_step):
-    return [x for x in plot if int(x["step"]) > latest_step]
-
-
 def _cast_to_numbers(datapoints):
     for datapoint in datapoints:
         for k, v in datapoint.items():
@@ -37,11 +33,6 @@ def _adapt_path(live, name):
     if live._dvc_repo is not None:
         name = rel_path(name, live._dvc_repo.root_dir)
     return name
-
-
-def _adapt_plot_datapoints(live, plot):
-    datapoints = _get_unsent_datapoints(plot, live._latest_studio_step)
-    return _cast_to_numbers(datapoints)
 
 
 def _adapt_image(image_path):
@@ -71,15 +62,18 @@ def get_studio_updates(live):
     metrics_file = _adapt_path(live, metrics_file)
     metrics = {metrics_file: {"data": metrics}}
 
-    plots = {
-        _adapt_path(live, name): _adapt_plot_datapoints(live, plot)
-        for name, plot in plots.items()
-    }
-    plots = {k: {"data": v} for k, v in plots.items()}
+    plots_to_send = {}
+    live._num_points_read_from_file = {}
+    for name, plot in plots.items():
+        path = _adapt_path(live, name)
+        nb_points_already_sent = live._num_points_sent_to_studio.get(path, 0)
+        plots_to_send[path] = _cast_to_numbers(plot[nb_points_already_sent:])
+        live._num_points_read_from_file.update({path: len(plot)})
 
-    plots.update(_adapt_images(live))
+    plots_to_send = {k: {"data": v} for k, v in plots_to_send.items()}
+    plots_to_send.update(_adapt_images(live))
 
-    return metrics, params, plots
+    return metrics, params, plots_to_send
 
 
 def get_dvc_studio_config(live):
@@ -120,6 +114,8 @@ def post_to_studio(live, event):
             live._studio_events_to_skip.add("data")
             live._studio_events_to_skip.add("done")
     elif event == "data":
+        for path, num_points in live._num_points_read_from_file.items():
+            live._num_points_sent_to_studio[path] = num_points
         live._latest_studio_step = live.step
 
     if event == "done":
--- a/src/dvclive/studio.py
+++ b/src/dvclive/studio.py
@@ -63,12 +63,10 @@ def get_studio_updates(live):
     metrics = {metrics_file: {"data": metrics}}
 
     plots_to_send = {}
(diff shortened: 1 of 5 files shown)


src/dvclive/studio.py
```diff
@@ -120,6 +114,8 @@ def post_to_studio(live, event):
             live._studio_events_to_skip.add("data")
             live._studio_events_to_skip.add("done")
     elif event == "data":
+        for path, num_points in live._num_points_read_from_file.items():
+            live._num_points_sent_to_studio[path] = num_points
```
dberenbaum, line 118: Is it enough to move the logic you had before down here (and drop `live._num_points_read_from_file`)?

```python
for name, plot in plots.items():
    live._nb_points_sent_to_studio[path] = len(plot)
```

AlexandreKempf: My bad I didn't see I could have access to `plots` here. I though I needed to reload the file, thus my change. I corrected that.

dberenbaum (approved):
LGTM! Do you think there are any additional tests that would be helpful to add?

AlexandreKempf:
> LGTM! Do you think there are any additional tests that would be helpful to add?

Sure! I took a bad habit the last few years not to write tests. It needs to become a habit once again. Sorry for that, I'll add them :)

tests/test_post_to_studio.py
```diff
@@ -464,3 +464,17 @@ def test_post_to_studio_skip_if_no_repo_url(
         live.next_step()
 
     assert mocked_post.call_count == 0
+
+
+def test_post_to_studio_increment_internal_state(
```
shcheklein, line 469: we should not be testing an internal state, rather the actual bug / scenario that you we are trying to solve / prevent

tbh, I don't quite understand it from the description / I don't have enough context.

could you describe it and / or write an actual test that would cover it? it will be easier to review it then + it'll force us to have a proper test for this

shcheklein (changes requested):
We need to have a proper test, from the brief description I don't quite understand in what situations we have a bug - it would be helpful to have a test that makes it obvious / clear

tests/frameworks/test_lightning.py
```diff
@@ -350,3 +350,53 @@ def test_dvclive_logger_init_args():
...
+    while True:
+        if live.step == step:
+            live.log_metric("test", 0.5)
+            step += 1
+        else:
+            time.sleep(0.001)
+
+
+@pytest.mark.timeout(3)
+def test_metrics_from_outside_lightning_updates_to_studio(
```
shcheklein, line 366: let's put a link as a comment to this PR to keep the context

tests/frameworks/test_lightning.py
```diff
@@ -350,3 +350,53 @@ def test_dvclive_logger_init_args():
...
+    dvclive_logger = DVCLiveLogger()
+    trainer = Trainer(
+        logger=dvclive_logger,
+        max_steps=4,
+        val_check_interval=2,
+        log_every_n_steps=1,
+        enable_checkpointing=False,
+    )
+
+    independent_logger = threading.Thread(
```
shcheklein, line 381: do we step the tread anywhere?

shcheklein (commented):
Thanks for updating the description, I think I understand it better now.

Just a few small questions to clarify re the test. Otherwise I think it should be fine (though I'm not an expert in the details re the data points management)

shcheklein:
@daavoo it would be great if you can take a look :)

tests/frameworks/test_lightning.py
AlexandreKempf: I'm guessing your meaning "stop". If that is the case, the thread runs as a daemon, which means that when Python ends execution, it stops the thread by itself :)

AlexandreKempf: It happens at the end of all the tests. In the meanwhile the thread is still active.
You're right it would have been a problem for the following tests.
I think we will change the implementation with the one @dberenbaum wrote in #790 as I find it cleaner.

shcheklein:
thanks for the update @AlexandreKempf ! can we add a test / change this test a bit to have more datapoints with a different cadence of updates. Correct me folks, if I'm wrong but we keep counter per metric path, right? some metric can be updated a few times before the next step (and even before sync), some once. And we need in all cases make sure that on `sync` we send "delta" properly.

shcheklein (approved):

dberenbaum:
@AlexandreKempf Are you ready to merge this one?

PR #790, opened 2024-02-20
Test studio counter

dberenbaum:
Adds a test for #788. I don't think we need to make a test specific to lightning or threads. The underlying problem is that if dvclive sent plots to studio multiple times with the same step value, only the first value would get sent to studio. Added a test here for that, which fails on main and passes in #788.

### DIFF
--- a/tests/frameworks/test_lightning.py
+++ b/tests/frameworks/test_lightning.py
@@ -268,7 +268,6 @@ class ValLitXOR(LitXOR):
 
 
 def test_lightning_val_updates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio_post):
-    """Test the `self.experiment._latest_studio_step -= 1` logic."""
     mocked_post, _ = mocked_studio_post
 
     model = ValLitXOR()
@@ -286,8 +285,6 @@ def test_lightning_val_updates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio
     logs, _ = parse_metrics(dvclive_logger.experiment)
     latest = logs[val_loss][-1]
     calls = mocked_post.call_args_list
-    # Without `self.experiment._latest_studio_step -= 1`
-    # Latest data would not be posted
     latest_called = False
     for call in calls:
         try:
--- a/tests/frameworks/test_lightning.py
+++ b/tests/frameworks/test_lightning.py
@@ -1,6 +1,4 @@
 import os
-import threading
-import time
 from contextlib import redirect_stdout
 from io import StringIO
 from unittest import mock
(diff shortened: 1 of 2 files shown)


dberenbaum:
@shcheklein Hopefully this helps clarify #788

dberenbaum:
The changes to the lightning tests just revert what's in #788 since I don't think we need to change any framework-specific tests here

tests/test_post_to_studio.py
```diff
@@ -464,3 +464,38 @@ def test_post_to_studio_skip_if_no_repo_url(
...
+def test_post_to_studio_repeat_step(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+    live = Live()
+
+    foo_path = (Path(live.plots_dir) / Metric.subfolder / "foo.tsv").as_posix()
+
+    mocked_post, _ = mocked_studio_post
+
+    live.step = 0
+    live.log_metric("foo", 1)
+    live.sync()
```
shcheklein, line 478: is `live.sync()` considered public / documented? or is it an internal implementation?

dberenbaum: We haven't had a good reason to document it, but it was written with the intent it could be made public if a use case for it arises. See the note in https://github.com/iterative/dvclive/pull/749#issue-2030974900.

dberenbaum: The test could also be written by using `next_step` followed by manually overriding the step value.

shcheklein: okay, I don't have a strong preference here tbh. If we feel `sync` won't be refactored we can keep it this way (probably it's better vs patching the step).

do we need to add tests for when there are a few values logged at a different pace?

shcheklein: anyways, folks, your call on all this ... I can review one more time from the technical point of view when you decide how to go

tests/frameworks/test_lightning.py
```diff
@@ -270,6 +268,7 @@ def validation_step(self, *args, **kwargs):
 
 
 def test_lightning_val_updates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio_post):
+    """Test the `self.experiment._latest_studio_step -= 1` logic."""
```
AlexandreKempf, line 271: you can remove this comment, we don't have that hacked line anymore ;)

tests/frameworks/test_lightning.py
```diff
@@ -287,7 +286,8 @@ def test_lightning_val_updates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio
     logs, _ = parse_metrics(dvclive_logger.experiment)
     latest = logs[val_loss][-1]
     calls = mocked_post.call_args_list
-
+    # Without `self.experiment._latest_studio_step -= 1`
+    # Latest data would not be posted
```
AlexandreKempf, line 290: you can remove this comment, we don't have that hacked line anymore ;)

AlexandreKempf (approved):
Ok I see what you meant now. 
Your test is more straightforward and easy to understand. Thanks for the PR.

:+1: Approved, with some minor comments

PR #808, opened 2024-04-05
Align JSON/YAML indentation settings better with common formatting styles

sisp:
Just a minor esthetics change: I've updated the indentation settings for JSON and YAML to align better with common formatting styles.

JSON is commonly indented with 2 spaces and not 4. 2 spaces also reduces storage and bandwidth waste a little.

YAML is often formatted with a 2 spaces offset for lists. For instance, Prettier formats it like this.

```diff
 k:
-- 1
+  - 1
-- 2
+  - 2
```

Also, `ruamel.yaml`'s documentation states:

> The above example with the often seen `yaml.indent(mapping=2, sequence=4, offset=2)` indentation:
>
> ```yaml
> x:
>   y:
>     - b: 1
>     - 2
> ```
>
> &mdash; https://yaml.readthedocs.io/en/latest/detail/#indentation-of-block-sequences

---

### DIFF
--- a/src/dvclive/serialize.py
+++ b/src/dvclive/serialize.py
@@ -45,7 +45,7 @@ def dump_yaml(content, output_file):
         yaml.dump(content, fd)
 
 
-def dump_json(content, output_file, indent=4, **kwargs):
+def dump_json(content, output_file, indent=2, **kwargs):
     make_dir(output_file)
     with open(output_file, "w", encoding="utf-8") as f:
         json.dump(content, f, indent=indent, **kwargs)
--- a/src/dvclive/serialize.py
+++ b/src/dvclive/serialize.py
@@ -31,6 +31,7 @@ def get_yaml():
 
     yaml = YAML()
     yaml.default_flow_style = False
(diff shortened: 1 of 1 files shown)


dberenbaum:
Related: https://github.com/iterative/dvc/pull/10379

shcheklein:
I think we should be aligned with how DVC formats its dvc.yaml. `.dvc`, etc. Otherwise it's hard to come up with a set of linter rules to cover both. @sisp could double check this please?

sisp:
@shcheklein DVCLive and DVC are currently (prior to this PR) aligned. For YAML(-like) files (`dvc.yaml`, `*.dvc`), this means there's a list offset of 0, which is `ruamel.yaml`'s default setting but IMO not so common (e.g., Prettier uses a list offset of 2 as I suggest in this PR). To keep formatting aligned between DVCLive and DVC, I created also https://github.com/iterative/dvc/pull/10379.

shcheklein:
Got it, thanks @sisp . Let's keep the discussion there then.

PR #813, opened 2024-04-18
Fixes #812

### LINKED ISSUE #812: log_sklearn_plot: exception when plot params are passed.
When plot params (title, x_label etc.) are passed, e.g.:

````python
from dvclive import Live

with Live() as live:
  y_true = [0, 0, 1, 1]
  y_pred = [1, 0, 1, 0]
  y_score = [0.1, 0.4, 0.35, 0.8]
  live.log_sklearn_plot("roc", y_true, y_score)
  live.log_sklearn_plot(
    "confusion_matrix", y_true, y_pred, name="cm.json", title="Test")
````

the exception "TypeError: got an unexpected keyword argument 'title'" is thrown.

It's because of this line in the log_sklearn_plot method:
````python
sklearn_kwargs = {
    k: v for k, v in kwargs.items() if k not in plot_config or k != "normalized"
}

````
The condition ````if k not in plot_config or k != "normalized"```` is always true for all params different that "normalized".

### PR DESCRIPTION
**Pull request by ocraft**
Fix for #812

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -621,6 +621,10 @@ class Live:
         labels: Union[List, np.ndarray],
         predictions: Union[List, Tuple, np.ndarray],
         name: Optional[str] = None,
+        title: Optional[str] = None,
+        x_label: Optional[str] = None,
+        y_label: Optional[str] = None,
+        normalized: Optional[bool] = None,
         **kwargs,
     ):
         """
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/dvclive/live.py [hunk 1]
```diff
@@ -664,7 +664,7 @@ def log_sklearn_plot(
             raise InvalidPlotTypeError(name)
 
         sklearn_kwargs = {
-            k: v for k, v in kwargs.items() if k not in plot_config or k != "normalized"
+            k: v for k, v in kwargs.items() if k not in plot_config or k == "normalized"
```
dberenbaum, line 667:
```suggestion
            k: v for k, v in kwargs.items() if k not in plot_config
```

We don't need any check for normalized here. It's already part of `plot_config`, which is all that's needed. We don't want to include it in `sklearn_kwargs` since it's purely to set the dvc plots template to `confusion_normalized`.

dberenbaum:
Thanks @ocraft! Ideally, I would like to make `title`, `x_label`, `y_label`, and `normalized` all named arguments to `log_sklearn_plot()` for a couple reasons:

1. It looks like all of these are undocumented and hidden now (I'm curious how you found out how to use them)
2. The docs state that `kwargs` are passed to the underlying sklearn methods, which is inaccurate since it is really a mix of these dvc plot parameters and sklearn args. It would be better to keep these separate rather than rely on this implicit parsing of the kwargs.

If you are interested, feel free to try this, but I'm fine to merge without that if it's beyond what you want to take on.

ocraft:
I think that I can do that. Personally, I just expected that log_plot and log_sklearn_plot would have the same arguments; it seemed logical to me.

src/dvclive/live.py
```diff
@@ -663,9 +676,7 @@ def log_sklearn_plot(
         else:
             raise InvalidPlotTypeError(name)
 
-        sklearn_kwargs = {
-            k: v for k, v in kwargs.items() if k not in plot_config or k != "normalized"
-        }
+        sklearn_kwargs = {k: v for k, v in kwargs.items() if k not in plot_config}
         plot.step = self.step
         plot.dump(val, **sklearn_kwargs)
```
dberenbaum, line 681:
```suggestion
        plot.step = self.step
        plot.dump(val, **kwargs)
```

Since the plot config options are now named args, I don't think it should be necessary to do anything here but pass `kwargs` directly.

dberenbaum (changes requested):
@ocraft Please see the one comment. If you can ping me once it's ready, it should be good to approve after that. Thanks again!

dberenbaum (approved):

PR #814, opened 2024-04-18
post to studio in thread to avoid blocking

dberenbaum:
Closes https://github.com/iterative/studio-support/issues/95

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -6,6 +6,7 @@ import logging
 import math
 import os
 import shutil
+import queue
 import tempfile
 import threading
 
@@ -172,6 +173,7 @@ class Live:
         self._studio_events_to_skip: Set[str] = set()
         self._dvc_studio_config: Dict[str, Any] = {}
         self._num_points_sent_to_studio: Dict[str, int] = {}
+        self._studio_queue = None
         self._init_studio()
 
         self._system_monitor: Optional[_SystemMonitor] = None  # Monitoring thread
@@ -297,7 +299,7 @@ class Live:
             self._studio_events_to_skip.add("start")
             self._studio_events_to_skip.add("done")
         else:
-            self.post_to_studio("start")
+            post_to_studio(self, "start")
 
     def _init_report(self):
         if self._report_mode not in {None, "html", "notebook", "md"}:
@@ -429,7 +431,7 @@ class Live:
 
         self.make_report()
 
-        self.post_to_studio("data")
+        self.post_data_to_studio()
 
     def next_step(self):
         """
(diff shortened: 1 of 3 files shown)


src/dvclive/live.py
```diff
@@ -882,7 +883,8 @@ def make_dvcyaml(self):
 
     @catch_and_warn(DvcException, logger)
     def post_to_studio(self, event: Literal["start", "data", "done"]):
-        post_to_studio(self, event)
+        thread = threading.Thread(target=post_to_studio, args=(self, event))
```
shcheklein, line 886: A few questions:

- do we still need `@catch_and_warn(DvcException, logger)` above?
- do we start a new thread per each request now? sound expensive and can for example exhaust OS if post_to_studio is running faster then we process them
- can we use a single thread (and a queue for example)?

shcheklein (commented):
Thanks @dberenbaum for a prompt fix! A few questions / comments.

src/dvclive/live.py
```diff
@@ -880,9 +883,19 @@ def make_dvcyaml(self):
...
+        if not self._studio_queue:
+            self._studio_queue = queue.Queue()
+
+            def worker():
+                while True:
+                    item = self._studio_queue.get()
+                    post_to_studio(item, "data")
+                    self._studio_queue.task_done()
+
+            threading.Thread(target=worker, daemon=True).start()
```
shcheklein, line 896: where is post_to_studio defined now? 

One thing to test - Ctrl-C handling for this. Just to make sure that our threads and queue wait doesn't mess up apps in any way. Thanks @dberenbaum for iterating on this!

shcheklein: should we wrap the call to post_to_studio in @catch_and_warn(DvcException, logger) ?

src/dvclive/live.py
shcheklein: why not run start and done through the same queue? it's better a bit since start will initialize the queue - makes it a bit more thread-safer

src/dvclive/live.py
dberenbaum: > where is post_to_studio defined now?

https://github.com/iterative/dvclive/blob/edb5ee3182f93d3f5e0d5f63416ec83ea3daf431/src/dvclive/studio.py#L99

> One thing to test - Ctrl-C handling for this. Just to make sure that our threads and queue wait doesn't mess up apps in any way. Thanks @dberenbaum for iterating on this!

No, thanks for keeping me honest here. It's a pretty simple change but could easily see it breaking some things. This is one reason I wanted to keep start and done separate per your question above. The queue is set up in daemon mode which can exit without making all expected calls (without it I noticed tests were hanging). By separating start and end, we ensure those are called and returned in most scenarios. If a particular live data call gets missed, I don't think it's as critical.

PR #816, opened 2024-04-29
Drop outdated flakey lightning test

shcheklein:
Caused by https://github.com/iterative/dvclive/actions/runs/8872661976/attempts/1

Decision to drop the test: https://github.com/iterative/dvclive/pull/816#issuecomment-2083668012

### DIFF
--- a/tests/frameworks/test_lightning.py
+++ b/tests/frameworks/test_lightning.py
@@ -267,39 +267,6 @@ class ValLitXOR(LitXOR):
         return loss
 
 
-def test_lightning_val_updates_to_studio(tmp_dir, mocked_dvc_repo, mocked_studio_post):
-    mocked_post, _ = mocked_studio_post
-
-    model = ValLitXOR()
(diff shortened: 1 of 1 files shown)


dberenbaum:
Were you seeing tests fail locally? Was it the same one failing in the CI here?

shcheklein:
Same CI - it's flakey, probably since floats comparison is not reliable. It takes a few runs to hit it.

dberenbaum:
It may be related to https://github.com/iterative/dvclive/pull/814. So far I can't reproduce locally though. As explained [here](https://github.com/iterative/dvclive/pull/814#discussion_r1572725259), there is no guarantee now that all expected data calls are sent before the main process dies. I think this is better for users (no reason to make the training process hang for something like a studio server error) but annoying for tests.

dberenbaum:
I think we should just drop this test. It covered special lightning logic that was removed in https://github.com/iterative/dvclive/pull/788. There's no reason we should need to specifically test lightning for this scenario now.

PR #827, opened 2024-06-08
fix(studio): wait for studio metrics publish to complete on end

shcheklein:
W/o this fix in the `example-get-started-experiments` we are not getting images published to Studio (it doesn't have time to complete, `evaluate` exits faster).

--------------

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -905,6 +905,11 @@ class Live:
 
         self._studio_queue.put(self)
 
+    def _wait_for_studio_updates_posted(self):
+        if self._studio_queue:
+            logger.debug("Waiting for studio updates to be posted")
+            self._studio_queue.join()
+
     def end(self):
         """
         Signals that the current experiment has ended.
@@ -945,6 +950,8 @@ class Live:
                 )
 
         self.save_dvc_exp()
+
+        self._wait_for_studio_updates_posted()
 
         # Mark experiment as done
         post_to_studio(self, "done")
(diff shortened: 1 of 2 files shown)


dberenbaum:
Looking at the [discussion](https://github.com/iterative/dvclive/pull/814/files#r1572688608) from when we added threads, I think this was intentional and is why the thread has `daemon=True`. Some questions:

1. Does dropping `daemon=True` accomplish the same thing?
2. What happens if any call to Studio hangs? I think that was why I added `daemon=True`.
3. Could we make a final `data` post to Studio in the main thread? That would ensure the final data is posted without needing every post to return.

shcheklein:
> Looking at the [discussion](https://github.com/iterative/dvclive/pull/814/files#r1572688608) from when we added threads, I think this was intentional and is why the thread has daemon=True. Some questions:

Right, but at the end of the process if becomes unpredictable / we are losing data ... not sure this is expected tbh ...

> Does dropping daemon=True accomplish the same thing?

We would need probably then a way to signal that thread to complete?

> Does dropping daemon=True accomplish the same thing?

In this case it be waiting _only_ at the end ... we can probably put some timeout? But as far as I understand we already do have some timeouts on post to studio call. So, it should not be infinite ...

> That would ensure the final data is posted without needing every post to return.

Could you clarify please? Not sure how this idea is easier / can help tbh ...

dberenbaum (approved):
If we already have some timeouts for Studio posts, I'm fine to merge this. I assume it should be no worse than what we had before threads.

shcheklein:
I assume that timeout is specified here:

```python
response = requests.post(
            url,
            json=body,
            headers={
                "Content-type": "application/json",
                "Authorization": f"token {token}",
            },
            timeout=(30, 5),
        )
```

PR #833, opened 2024-07-10
make hf tests work

dberenbaum:
@shcheklein Added a couple minor changes to your PR

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ tests = [
   "pytest_voluptuous",
   "dpath",
   "transformers[torch]",
-  "keras<3"
+  "tf-keras"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -57,7 +57,7 @@ tests = [
   "pytest_voluptuous",
   "dpath",
   "transformers[torch]",
-  "keras<3"
+  "tf-keras"
```
dberenbaum, line 60: See https://github.com/iterative/dvclive/pull/832#discussion_r1672621506. This should work in huggingface while still allowing us to run keras tests with keras 3.

shcheklein (approved):

Issue #769, opened 2024-02-05
Path not supported for some path arguments


AlexandreKempf:
Most of the path arguments handle both `str` and `pathlib.Path` object. 

It is not the case for `Live`'s `dvcyaml` while it should probably be.

AbdullahMakhdoom:
Hi @dberenbaum , I would like to contribute for this issue. If I understand correct `dvcyaml (str | None)`

AbdullahMakhdoom:
Hi @dberenbaum, I'm interested in contributing to dvclive, and I believe this issue is a great starting point. If I'm not mistaken, it appears that the `Live`'s `dvcyaml` argument should also accept `pathlib.Path` objects. Therefore, the argument should be defined as `dvcyaml: Optional[Union[str, pathlib.Path]] = "dvc.yaml"`
Could you confirm if my understanding is correct?

dberenbaum:
> If I'm not mistaken, it appears that the `Live`'s `dvcyaml` argument should also accept `pathlib.Path` objects.

Correct.

> Therefore, the argument should be defined as `dvcyaml: Optional[Union[str, pathlib.Path]] = "dvc.yaml"`

Partially correct. Fixing the type hint would be nice, but the underlying issue goes deeper. If you try an example with `dvcyaml` set to some `pathlib.Path` value, you should see where it breaks. This issue is to fix any underlying problems so that `pathlib.Path` values work.

AbdullahMakhdoom:
@dberenbaum I tried passing `dvcyaml` arguments a `pathlib.Path` value, it did not break but rather silently defaulted to "dvc.yaml" as mentioned by this [function](https://github.com/iterative/dvclive/blob/main/src/dvclive/live.py#L267).

One solution I could think of is type converting to string, like....

```
  def _init_dvc_file(self) -> str:
      if isinstance(self._dvcyaml, Path):
          self._dvcyaml = str(self._dvcyaml)
      if isinstance(self._dvcyaml, str):
          if os.path.basename(self._dvcyaml) == "dvc.yaml":
              return self._dvcyaml
          raise InvalidDvcyamlError
      return "dvc.yaml"

```

I've manually tested the changes, and they work as expected. What do you think of this solution?

shcheklein:
@AbdullahMakhdoom I think it's fine to fix it this way. I would also add something like:

if self._dvcyaml:
  raise InvalidDvcyamlError("DVC file path mush be of str or Path type")

before the `return dvc.yaml`

[...]

PR #842, opened 2024-09-09
Support `pathlib.Path` values for `dvcyaml` argument in `Live()`.

AbdullahMakhdoom:
…cyaml` argument.

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -82,7 +82,7 @@ class Live:
         resume: bool = False,
         report: Literal["md", "notebook", "html", None] = None,
         save_dvc_exp: bool = True,
-        dvcyaml:  Union[str, os.PathLike, bool, None] = "dvc.yaml",
+        dvcyaml: Union[str, os.PathLike, bool, None] = "dvc.yaml",
         cache_images: bool = False,
         exp_name: Optional[str] = None,
         exp_message: Optional[str] = None,
@@ -269,14 +269,14 @@ class Live:
             return "dvc.yaml"
         if isinstance(self._dvcyaml, bool):
             return "dvc.yaml"
-        
+
         self._dvcyaml = os.fspath(self._dvcyaml)
         if (
             isinstance(self._dvcyaml, str)
             and os.path.basename(self._dvcyaml) == "dvc.yaml"
         ):
             return self._dvcyaml
-        
+
         raise InvalidDvcyamlError
 
     def _init_dvc_pipeline(self):
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -82,7 +82,7 @@ class Live:
         resume: bool = False,
         report: Literal["md", "notebook", "html", None] = None,
         save_dvc_exp: bool = True,
-        dvcyaml: Optional[str] = "dvc.yaml",
+        dvcyaml:  Union[str, os.PathLike, bool, None] = "dvc.yaml",
         cache_images: bool = False,
         exp_name: Optional[str] = None,
         exp_message: Optional[str] = None,
@@ -104,11 +104,11 @@ class Live:
                 part of `Live.end()`. Defaults to `True`. If you are using DVCLive
                 inside a DVC Pipeline and running with `dvc exp run`, the option will be
                 ignored.
-            dvcyaml (str | None): where to write dvc.yaml file, which adds DVC
+            dvcyaml (str | Path | None): where to write dvc.yaml file, which adds DVC
                 configuration for metrics, plots, and parameters as part of
                 `Live.next_step()` and `Live.end()`. If `None`, no dvc.yaml file is
                 written. Defaults to `"dvc.yaml"`. See `Live.make_dvcyaml()`.
-                If a string like `"subdir/dvc.yaml"`, DVCLive will write the
+                If a string or Path like `"subdir/dvc.yaml"`, DVCLive will write the
                 configuration to that path (file must be named "dvc.yaml").
                 If `False`, DVCLive will not write to "dvc.yaml" (useful if you are
                 tracking DVCLive metrics, plots, and parameters independently and
@@ -265,11 +265,19 @@ class Live:
             self._include_untracked.append(self.dir)
 
     def _init_dvc_file(self) -> str:
-        if isinstance(self._dvcyaml, str):
-            if os.path.basename(self._dvcyaml) == "dvc.yaml":
-                return self._dvcyaml
-            raise InvalidDvcyamlError
-        return "dvc.yaml"
+        if self._dvcyaml is None:
+            return "dvc.yaml"
+        if isinstance(self._dvcyaml, bool):
+            return "dvc.yaml"
+        
+        self._dvcyaml = os.fspath(self._dvcyaml)
+        if (
+            isinstance(self._dvcyaml, str)
+            and os.path.basename(self._dvcyaml) == "dvc.yaml"
+        ):
+            return self._dvcyaml
+        
+        raise InvalidDvcyamlError
 
     def _init_dvc_pipeline(self):
         if os.getenv(env.DVC_EXP_BASELINE_REV, None):
@@ -334,6 +342,7 @@ class Live:
         """
         with tempfile.TemporaryDirectory() as dirpath:
             self._dir = os.path.join(dirpath, self._dir)
+            self._dvcyaml = os.fspath(self._dvcyaml)
             if isinstance(self._dvcyaml, str):
                 self._dvc_file = os.path.join(dirpath, self._dvcyaml)
             self._save_dvc_exp = False
--- a/tests/test_make_dvcyaml.py
+++ b/tests/test_make_dvcyaml.py
@@ -2,6 +2,7 @@ import os
 
 import pytest
 from PIL import Image
+from pathlib import Path
 
 from dvclive import Live
 from dvclive.dvc import make_dvcyaml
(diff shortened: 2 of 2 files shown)


src/dvclive/live.py
```diff
@@ -82,7 +82,7 @@ def __init__(
         resume: bool = False,
         report: Literal["md", "notebook", "html", None] = None,
         save_dvc_exp: bool = True,
-        dvcyaml: Optional[str] = "dvc.yaml",
+        dvcyaml: Optional[str, Path] = "dvc.yaml",
```
skshetry, line 85:
```suggestion
        dvcyaml: Optional[str, os.Pathlike[str]] = "dvc.yaml",
```

Use `os.PathLike[str]`, which is more generic than `Path`.

src/dvclive/live.py
```diff
@@ -265,10 +265,14 @@ def _init_dvc(self):  # noqa: C901
             self._include_untracked.append(self.dir)
 
     def _init_dvc_file(self) -> str:
+        if isinstance(self._dvcyaml, Path):
+            self._dvcyaml = str(self._dvcyaml)
```
skshetry, line 269:
```suggestion
            self._dvcyaml = os.fspath(self._dvcyaml)
```

AbdullahMakhdoom:
Fixes [Issue#769](https://github.com/iterative/dvclive/issues/769)

dberenbaum:
Thanks @AbdullahMakhdoom! Please note in the description that it fixes #769.

src/dvclive/live.py
```diff
@@ -265,10 +265,14 @@ def _init_dvc(self):  # noqa: C901
...
         if isinstance(self._dvcyaml, str):
             if os.path.basename(self._dvcyaml) == "dvc.yaml":
                 return self._dvcyaml
             raise InvalidDvcyamlError
```
dberenbaum, line 273: Can we drop this line if we are adding the exception below?

tests/test_dvc.py
```diff
@@ -220,12 +221,13 @@ def test_get_exp_name_duplicate(tmp_dir, mocked_dvc_repo, mocker, caplog):
     assert msg in caplog.text
 
 
-def test_test_mode(tmp_dir, monkeypatch, mocked_dvc_repo):
+@pytest.mark.parametrize("dvcyaml_path", ["dvc.yaml", Path("dvcyaml/dvc.yaml")])
```
dberenbaum, line 224: Sorry for not getting back to you on this earlier, but I would vote to put this in a separate test in `tests/test_make_dvcyaml.py`. It's probably enough to add a parameter [here](https://github.com/iterative/dvclive/blob/68884628e5845f95b89c950016d1b9a1a190cb60/tests/test_make_dvcyaml.py#L425).

src/dvclive/live.py
AbdullahMakhdoom: Hi @skshetry , thanks for reviewing this.
So this PR fixed this [issue](https://github.com/iterative/datachain/issues/386), for `dvcyaml` argument to support `path.Pathlib` value. Will the use of `os.PathLike[str]` be more appropriate here instead of pathlib.Path?

tests/test_dvc.py
AbdullahMakhdoom: Yeah, this makes sense.

src/dvclive/live.py
AbdullahMakhdoom: Yep, makes sense.

shcheklein (approved):

dberenbaum (approved):

shcheklein:
@AbdullahMakhdoom it seems mypy failed, could you please take a look. Thanks!

src/dvclive/live.py
```diff
@@ -265,11 +265,18 @@ def _init_dvc(self):  # noqa: C901
...
 
     def _init_dvc_file(self) -> str:
-        if isinstance(self._dvcyaml, str):
-            if os.path.basename(self._dvcyaml) == "dvc.yaml":
-                return self._dvcyaml
-            raise InvalidDvcyamlError
-        return "dvc.yaml"
+        if self._dvcyaml is None:
+            return "dvc.yaml"
+        if isinstance(self._dvcyaml, bool):
```
shcheklein, line 270: hmm, so is it something new? do we want to support bool as well?

we need to update docs then?

what happens if it's False (sounds like `dvc.yaml` should not be produced at all then?)

AbdullahMakhdoom: On True/False/None, it defaults to "dvc.yaml".
I had to make this change to pass an already written test case, which tests`True`, `False` values to  `dvcyaml` argument.

shcheklein: @AbdullahMakhdoom could you please point me to that test?

skshetry: @shcheklein, you can read the docstring for `dvcyaml`.

shcheklein: okay, so:

```
                If `False`, DVCLive will not write to "dvc.yaml" (useful if you are
                tracking DVCLive metrics, plots, and parameters independently and
                want to avoid duplication).
```

it means we can't default to `dvc.yaml` - we should probably not running this function at all then if it's False.

src/dvclive/live.py
```diff
@@ -265,11 +265,18 @@ def _init_dvc(self):  # noqa: C901
...
+            return "dvc.yaml"
+        if isinstance(self._dvcyaml, Path):
+            self._dvcyaml = str(self._dvcyaml)
```
skshetry, line 273: Hi, I'll suggest avoiding `isinstance(..., Path)` checks as much as possible.

You can use `os.fspath()` to convert `Path` or `PathLike` to `str` (`str` are  returned as-is).

src/dvclive/live.py
dberenbaum: @shcheklein The code is maybe a bit of a mess here, but this sets `self._dvcfile`, which shouldn't matter for whether the `dvc.yaml` file is written. The code still checks `self._dvcyaml` to decide whether to write `dvc.yaml` [here](https://github.com/iterative/dvclive/blob/68884628e5845f95b89c950016d1b9a1a190cb60/src/dvclive/live.py#L427-L428). This is tested here:

https://github.com/iterative/dvclive/blob/68884628e5845f95b89c950016d1b9a1a190cb60/tests/test_make_dvcyaml.py#L461-L463

It shouldn't matter much if `self._dvcfile` is set in this scenario, but the slight justification for still setting it to `dvc.yaml` is that if someone has set `Live(dvcyaml=False)` but then manually calls `make_dvcyaml()`, it will still write to `dvc.yaml`.

src/dvclive/live.py
AbdullahMakhdoom: @skshetry replaced these lines with `os.fspath()`

src/dvclive/live.py
AbdullahMakhdoom: It's unclear why using `False` is necessary here, as setting `dvcyaml` to `None` appears to produce the same outcome. Both tests ensure that the `dvc.yaml` file is not created when `dvcyaml` is either False or None, indicating that these options are functionally equivalent in this context. 

```

def test_make_dvcyaml_false(tmp_dir, mocker):
    dvclive = Live("logs", dvcyaml=False)
    dvclive.end()

    assert not os.path.exists("dvc.yaml")


def test_make_dvcyaml_none(tmp_dir, mocker):
    dvclive = Live("logs", dvcyaml=None)
    dvclive.end()

    assert not os.path.exists("dvc.yaml")

Comment (AbdullahMakhdoom) [reply src/dvclive/live.py:270]: Also, apologize for the delayed responses, I've been quite busy with my day job. Please let me know the best way to move forward, and I'll make sure to do my best.

Comment (dberenbaum) [reply src/dvclive/live.py:270]: No problem @AbdullahMakhdoom, thanks for keeping on top of this!

`dvcyaml` used to be a bool argument and support for bools was left to avoid making a breaking change. We should probably have dropped this during the last major release but missed it.

**Comment by shcheklein** [comment]
@skshetry please approve and merge if everything looks good to you.

**Review by skshetry — APPROVED** [review]

**Review by skshetry — APPROVED** [review]

**Review by skshetry — APPROVED** [review]

**Review by skshetry — APPROVED** [review]

**Comment by skshetry** [comment]
Apologies for multiple approval. GH bugged out. :(

File: src/dvclive/live.py [hunk 7]
```diff
@@ -265,11 +265,19 @@ def _init_dvc(self):  # noqa: C901
...
     def _init_dvc_file(self) -> str:
-        if isinstance(self._dvcyaml, str):
-            if os.path.basename(self._dvcyaml) == "dvc.yaml":
-                return self._dvcyaml
-            raise InvalidDvcyamlError
-        return "dvc.yaml"
+        if self._dvcyaml is None:
+            return "dvc.yaml"
+        if isinstance(self._dvcyaml, bool):
+            return "dvc.yaml"
```
skshetry, line 271: Are we okay with returning `dvc.yaml` in case of `dvcyaml=False`?

skshetry: But this is unrelated to this PR. We were already doing that.

AbdullahMakhdoom: @skshetry as pointed out by @dberenbaum, "the slight justification for still setting it to dvc.yaml is that if someone has set Live(dvcyaml=False) but then manually calls make_dvcyaml(), it will still write to dvc.yaml. "dvc.yaml" is being returned ".

PR #852, opened 2024-12-18
Fix local variable 'last_metrics' referenced before assignment

mluscon:
This happens on line :132 in case of handled exception in the first iteration of the loop.


shcheklein (approved):
thanks!

PR #860, opened 2025-01-25
fix(studio): package data to send in main thread

### LINKED ISSUE #858: exp hangs forever when logging images
Hey,

I started logging images more specially matplotlib figures using log_image. 
Everything seemed to be working but yesterday I started two experiments and they were both stuck in their last epoch. 
When I checked the logs both of them report following RuntimeError:
```
Exception in thread Thread-3 (worker):
Traceback (most recent call last):
  File "/home/fuchsfa/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/lib/python3.12/threading.py", line 1075, in _bootstrap_inner
    self.run()
  File "/home/fuchsfa/.local/share/uv/python/cpython-3.12.7-linux-x86_64-gnu/lib/python3.12/threading.py", line 1012, in run
    self._target(*self._args, **self._kwargs)
  File "/home/fuchsfa/wavelet-extraction-model/.venv/lib/python3.12/site-packages/dvclive/live.py", line 911, in worker
    post_to_studio(item, "data")
  File "/home/fuchsfa/wavelet-extraction-model/.venv/lib/python3.12/site-packages/dvclive/utils.py", line 182, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/fuchsfa/wavelet-extraction-model/.venv/lib/python3.12/site-packages/dvclive/studio.py", line 114, in post_to_studio
    metrics, params, plots = get_studio_updates(live)
                             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/fuchsfa/wavelet-extraction-model/.venv/lib/python3.12/site-packages/dvclive/studio.py", line 82, in get_studio_updates
    plots_to_send.update(_adapt_images(live))
                         ^^^^^^^^^^^^^^^^^^^
  File "/home/fuchsfa/wavelet-extraction-model/.venv/lib/python3.12/site-packages/dvclive/studio.py", line 56, in _adapt_images
                 ^^^^^^^^^^^^^^^^^^^^^
RuntimeError: dictionary changed size during iteration
```
So first off why did this not crash and just mark the experiment as failed but kept running indefinitely?
Secondly what is happening to cause the error?

The error does not occur using `dvc repro` , which makes sense because it occurs in the `post_to_studio` method.

Also is there a way for me to stop posting to studio for just this repository?

shcheklein:
Fixes root cause of the #858. We'll still need another fix / research on why it was hanging.

There was a race condition since both (main and Studio updates) threads were reading / updating the same `live` instance values simultaneously.

It seems it might be happening if a training loop is faster then Studio API call (can be when we have a lot of images).

## TODO

- [x] Add tests
- [ ] Do a followup on "Handle exceptions in the Studio thread"
- [x] Can we cut on memory footprint for plots (we read all the files and package them to send into the Studio post) - we can potentially conservatively "cut" them
- [x] Check what happens if we send images / plots data twice (e.g. images keep coming and we now filter only by `step > current_step- potentially means we could be sending same data twice)

----------------------------------

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -55,6 +55,7 @@ from .utils import (
     inside_notebook,
     matplotlib_installed,
     open_file_in_browser,
+    parse_metrics,
 )
 from .vscode import (
     cleanup_dvclive_step_completed,
@@ -135,7 +136,7 @@ class Live:
         self._save_dvc_exp: bool = save_dvc_exp
         self._step: Optional[int] = None
         self._metrics: Dict[str, Any] = {}
-        self._images: Dict[str, Any] = {}
+        self._images: Dict[str, Image] = {}
         self._params: Dict[str, Any] = {}
         self._plots: Dict[str, Any] = {}
         self._artifacts: Dict[str, Dict] = {}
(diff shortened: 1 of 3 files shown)


shcheklein:
@skshetry PTAL

src/dvclive/live.py
```diff
@@ -901,19 +902,41 @@ def make_dvcyaml(self):
...
+        params = load_yaml(self.params_file) if os.path.isfile(self.params_file) else {}
+        plots, metrics = parse_metrics(self)
+
+        # Plots can grow large, we don't want to keep in memory data
+        # that we 100% sent already
+        plots_to_send = {}
+        plots_start_idx = {}
+        for name, plot in plots.items():
+            num_points_sent = self._num_points_sent_to_studio.get(name, 0)
+            plots_to_send[name] = plot[num_points_sent:]
```
skshetry, line 915: I see below that it did `_cast_to_numbers` before? Do we need that here?

skshetry (approved):

PR #864, opened 2025-02-04
fix(studio): handle unexpected exceptions in updates thread

### LINKED ISSUE #858: exp hangs forever when logging images

shcheklein:
Continuation of https://github.com/iterative/dvclive/pull/860

Fixes #858 

We catch properly unexpected exceptions in Studio updates thread and continue consuming from the queue (to make it empty so `join` below doesn't hang). 

## TODO:

- [x] Add test

--------------------

### DIFF
--- a/src/dvclive/live.py
+++ b/src/dvclive/live.py
@@ -929,10 +929,17 @@ class Live:
             self._studio_queue = queue.Queue()
 
             def worker():
+                error_occurred = False
                 while True:
                     item, data = self._studio_queue.get()
-                    post_to_studio(item, "data", data)
-                    self._studio_queue.task_done()
+                    try:
+                        if not error_occurred:
+                            post_to_studio(item, "data", data)
+                    except Exception:
+                        logger.exception("Failed to post data to studio")
+                        error_occurred = True
+                    finally:
+                        self._studio_queue.task_done()
 
             threading.Thread(target=worker, daemon=True).start()
 
--- a/tests/test_post_to_studio.py
+++ b/tests/test_post_to_studio.py
@@ -274,6 +274,36 @@ def test_studio_updates_posted_on_end(tmp_path, mocked_dvc_repo, mocked_studio_p
     assert metrics_file.read_text() == metrics_content
 
 
+def test_studio_update_raises_exception(tmp_path, mocked_dvc_repo, mocked_studio_post):
+    # Test that if a studio update raises an exception, main process doesn't hang on
+    # queue join in the Live main thread.
+    # https://github.com/iterative/dvclive/pull/864
+    mocked_post, valid_response = mocked_studio_post
+
+    def post_raises_exception(*args, **kwargs):
+        if kwargs["json"]["type"] == "data":
+            # We'll hit this sleep only once, other calls are ignored
+            # after the exception is raised
+            time.sleep(1)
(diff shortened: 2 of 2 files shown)


src/dvclive/live.py
```diff
@@ -929,10 +929,17 @@ def post_data_to_studio(self):
...
-                    post_to_studio(item, "data", data)
-                    self._studio_queue.task_done()
+                    try:
+                        if not error_occurred:
+                            post_to_studio(item, "data", data)
+                    except Exception:
+                        logger.exception("Failed to post data to studio")
+                        error_occurred = True
+                    finally:
+                        self._studio_queue.task_done()
```
skshetry, line 942: Can we use `thread.is_alive()` to check and never push the data (or some other `Event`) instead of keeping a thread alive?

skshetry (approved):

src/dvclive/live.py
shcheklein: there can be still a race condition - thread is just still alive (but about to exit), we do a check, we push data and then  `queue.join()` hangs. I would rather to something dumb and simple (even in this case easily I missed some issues - multithreading is hard)

